PR #13, opened 2021-11-04
event-listener options & better handler event type

thetarnav:
Changed createEventListener types a bit: now when creating an event listener, the handler's event prop will have a correct type.

Before:
image

Now:
image

Also added `options` property to createEventListener function, for passing options to addEventListener, like `{passive: false}`.

### DIFF
--- a/packages/event-listener/src/index.ts
+++ b/packages/event-listener/src/index.ts
@@ -6,21 +6,25 @@ import { onMount, onCleanup } from "solid-js";
  * @param eventName - Event name to bind to
  * @param handler - Function handler to trigger
  * @param element - HTML element to bind the event to
+ * @param options - *useCapture* boolean or an object that specifies characteristics about the event listener.
  *
  * @example
  * ```ts
  * createEventListener("mouseDown", () => console.log("Click"), document.getElementById("mybutton"))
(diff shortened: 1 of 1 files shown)


davedbase:
Yup, this is a pretty reasonable expectation from the types. Thanks for cleaning that up! I'll merge and publish in the next 10 minutes.

davedbase:
@thetarnav sorry for the confusion! I ended up realizing that @atk had prepared a complete refactor of this primitive in a separate branch. I went ahead and published version 1.1.4 in case you're using. We however encourage you to jump on 1.2.3 with some radical improvements + tests. We expect another adjustment coming as well as soon as atk has a bit more free time. :-) just thought I'd document today's releases as to not confuse you.

thetarnav:
My bad, I should have checked other branches I guess. But I see that the rewrite has the same problem with the event type.

image

davedbase:
No, no worries at all. Honestly even I forgot that @atk had refactored. He did mention his own changes reflected the same problem. That's why I mentioned that we expect more adjustments, so factor in your change. If you have a moment we'd love another PR :-) if not I think Alex will take a stab at it.

PR #15, opened 2021-11-06
Intersection observer improvements

thetarnav:
some changes to the intersection-observer module to improve usability:

Non-breaking changes: 
- the `elements` argument now also accepts function.
This is so that createIntersectionObserver could be called without wrapping it with onMount when we want to pass in a ref, since onMount is used internally anyway.
- improved onChange callback type
- added a new helper function — createVisibilityObserver — to reduce the amount of code that has to be written just to observe if a single element is on a screen.

Breaking changes:
- combined root, rootMargin & threshold arguments into single `options` object. so that it matches the native IntersectionObserver api.
- changed the return values from being a 4-5 item tuple into an object. I think it's just easier to destructure and use it like that.

### DIFF
--- a/packages/intersection-observer/src/index.ts
+++ b/packages/intersection-observer/src/index.ts
@@ -1,41 +1,48 @@
-import { onMount, onCleanup } from "solid-js";
+import { onMount, onCleanup, createSignal, Accessor } from "solid-js";
+
+export interface IntersetionObserverOptions {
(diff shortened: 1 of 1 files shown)


davedbase:
This looks really great! I like the cleanup and general improvements. I'll update the README, example and publish a new version =)

PR #16, opened 2021-11-07
improve createViewportObserver

thetarnav:
An attempt to improve createViewportObserver.
If I understand the idea of this helper correctly — observing a dynamic number of elements, while easily handling their visibility changes via independent callbacks — then there were two odd things: the "setter" type, and passing initial elements as an argument when those elements wouldn't have their callbacks set.


davedbase:
Looking good. Also yes you are correct in how you've read the primitive. There's a demo of it on solid-site: https://github.com/solidjs/solid-site/blob/master/src/pages/Docs.tsx#L31 in case you'd like to see it in practical use.

I see you cleaned the setter types up, thanks for doing that. Much neater.

In regards to the initial elements I think it's nice to let the user provide an initial set of observers so that they don't have to use the add method. Having the flexibility for both is nice and not that much more code to support.

That's silly, seems like I forgot to add in the callback in the params :-p likely because I was scrambling to get it working to use right away. I'd prefer if we added support for it back in this PR but if you have a strong opinion otherwise, could forgo it.

thetarnav:
Yeah, I agree that initial elements would be helpful, I just took removing it as an easier option, since they weren't working anyway. But sure, I can add them back in, just what type should the param have? An array of tuples [Element, callback]?

davedbase:
Really good question and I'm pretty open to anything really. Array of tuples gives the user the ability to specify callbacks for each el. Honestly they can achieve this with the add method. I'm wondering if we keep the first param as a list of els, the second a common callback for all of them/

Basically the thought I have is use the params if you have a bunch of els with a single similar callback. Use the alternate add method when you want to bind specific callbacks.

The param interfaces could be the following for convenience:

```ts
createViewportObserver(els: Element[] | Element, callback: EntryCallback, options: IntersectionObserverOptions);
// or
createViewportObserver(options: IntersectionObserverOptions);
```

What do you think? Totally open here

thetarnav:
Well, this depends on the application. Something like that would cover the widest range of scenarios, I think:
```ts
const createViewportObserver = (
   initial: [MaybeAccessor<Element>, EntryCallback][] = [],
   options?: IntersectionObserverOptions,
) => {}
```
But then yeah if you want to use the same callback for all of them, then this would be a pain to do. It's hard to decide on this since my own applications rely only on the `add` method.

I think all of them can be covered by different overloads anyway

atk:
Whenever it makes sense, we should allow using our primitives as directives, too.

thetarnav:
I guess createIntersectionObserver could be used on a single element as a directive. But to use it with multiple elements in jsx, something like an IntersectionObserverProvider would be needed, I've made [this](https://github.com/thetarnav/life.data/blob/84e832298b2ad9ec489bb53f704656c9453af50e/src/utils/solid/ViewportObserver.tsx#L26) in my own project, but I guess it's outside the scope of "primitives".
@atk Do you have an example of a nice directive implementation that I can use?

davedbase:
I think this recent change that supports the three types of interfaces is still clean, concise and flexible. I'm feeling good about it. @atk any opinions?

In regards to the directives you can take a look at solid-site again. There's an example of createIntersectionObserver: https://github.com/solidjs/solid-site/blob/206138c5cbaf10d5f483f3dcc9c0390ec336a423/src/components/Nav.tsx#L100

We could modify the add method in  createViewportObserver to accept a similar interface so that it can be use as a directive as well.

@thetarnav no Provider necessary, Directives are like Svelte Actions. It's just a JSX convenience syntax Solid supports =)

thetarnav:
I see. Well the createIntersectionObserver works out of the box then, but with createViewportObserver I have an issue with the callback. Because add method accepts a function as the second param anyway, I have no way of checking if it's the callback or a callback accessor from the directive.
Other than something hacky like `cb.toString().startsWith("() => (") || cb.toString().startsWith("() => function")`.
So maybe createViewportObserver should return another function, that is exclusive for directive usage?
And is there even a good way to make types for these directives?

atk:
There's currently only one way to make the types for the directives:

```ts
declare module "solid-js" {
  namespace JSX {
    interface Directives {
      observer: (ref: HTMLElement, props: () => void) => void;
    }
  }
}
// this ensures the `JSX` import won't fall victim to tree shaking before typescript can use it
export type E = JSX.Element;
```

Also, we wanted to do named exports for those directive primitives.

davedbase:
Ok, thinking through these here are the options:

**Option 1**

A second exported method makes sense. I would have the `add` method replicate what `createIntersectionObserver` does so that they are equivalent in use. A user with a more advanced use case can use something like `addWithCallback` or something that's named nicer.

**Option 2**

Change the callback to an object in the second or third slot so you can detect its type properly (ie. if it contains callback).

```ts
// As directive
use:observer={{ callback: () => console.log('foo') }}
// As add method
add(el, { callback: () => console.log('foo') })
```

On a different note, I'm noticing that returning as an object can be a bit awkward for these primitives if we're encouraging use as a directive. `observer` is currently exporting the raw observer but if I want to do `use:observer` that could be so confusing.

Maybe we should consider:
```ts
return [ observer, { add, remove, start, stop, rawObserver } ];
// or
return { add, remove, start, stop, observer, observerDirective };
// or this lets you de-structure and specify a custom name
return { add, remove, start, stop, observer: [ directive, raw ] };
```
Sheesh, designing good/reasonable primitives is hard. :-p

thetarnav:
Actually, I've found a solution for the createViewportObserver directive use. No need to return another function, I just modified the logic of calling callbacks: basically, if the "callback" returns another callback it is also run with the same params.

Regarding the issue with return, I have to agree, after using it a little longer, renaming `add` every time is not the best experience. I had in mind the idea to use a tuple to split the common usage: `add` and "advanced" usage: *the rest*.

```ts
return [ add, { remove, start, stop, instance } ]
// or is we want to keep add & remove together:
return [ add, remove, { start, stop, instance } ]
```
This way `add` can be easily named as "observer" for directive usage or something else when using it as a hook.

Or we can go back to the idea of a separate function anyway if we want to maximalize clarity over easiness:
```ts
// observer for directive
// instance for intersection observer instance
return { add, remove, start, stop, observer, instance }
```

davedbase:
Oh great. I'm kind of just passively throwing weird ideas down at you on this topic, thanks for actually considering and making them practical and focusing on this. :-)

I personally think think this works very well:
```ts
return [ add, { remove, start, stop, instance } ]
```
I think in the documentation if we explain the logic for this output users will understand the benefit. That'll be my job once we lock this in. I do like this approach, very reasonable!

thetarnav:
no worries, weird ideas helps to find the best solution :)

davedbase:
I agree. So these changes look great. Do you have a moment to attempt testing this? @atk are implementing a new testing pattern using solid-register. Also this will be a bit of a monster to test properly.

atk:
Nothing too problematic, we just need to properly mock the API. You can leave the testing to me if you want.

thetarnav:
I have no experience with testing, but I can try to do some with some guidance and a framework in place.

davedbase:
The more help we can get with test coverage, the better off we will be. @atk your time is better spent improving TS since you're a powerhouse there. If you can setup these primitives with a basic test using solid-register then we can take it from there!

atk:
I added a basic test setup to main, so you only need to rebase.

thetarnav:
No way rebase just duplicates all the commits... Well anyway, I think everything should be tested now.

davedbase:
Wow @thetarnav this is pretty amazing effort. Thank you so much from the whole community. I love the work you've done on this with us. I hate to be presumptive but if you had an energy/desire to work on other primitives on the project, I'd love to consider you a maintainer. We need more hands to bring it up to a level of quality that it's not at right now.

If you're on Solid's Discord as well, we can connect there to coordinate as well.

Thank you again. Going to merge this in =)

PR #18, opened 2021-11-12
fix readme & createViewportObserver start

thetarnav:
I realized that the `start` function from createViewportObserver wasn't actually doing anything, since the initial array of elements passed to the createIntersectionObserver is empty. So now "restarting" should work similarly to createIntersectionObserver.

Also fixed some issues in the usage section in the readme.


davedbase:
Good catches here! Thanks.

PR #19, opened 2021-11-13
WIP [Stage-1]: Composite Effect primitive proposal

thetarnav:
Coming from vue I thought of bringing some reactive utilities from [vueuse](https://vueuse.org), specifically the utilities focused of watching changes of computations. Most of them are very simple *(maybe with exception of [`until`](https://vueuse.org/shared/until/))*, they're just helpers built on top of `createRoot`, `createEffect` and `on` that aim for reducing the amount of code for some repetitive use cases.

Their usages can found be [here](https://vueuse.org/functions.html#watch), and the API for solid would be pretty similar.

This is not finished, I just wanted to get feedback on some of the considerations:
- First is the question: does Solid even need these?
Solid aims to be simple, not easy, preferring to have things explicit and maybe even overly verbose just to keep everything under control. Maybe because of that some of these helpers don't really make sense, and I just want them because I'm used to having them in vue. Or perhaps some other versions would be more suitable.
I know I'll definitely use some of them, like a watch, debouncedWatch, throttledWatch, and until are pretty useful.

- The naming may be odd, since it doesn't follow the "create___" rule, but I think not everything has to.

- Another thing is that I used debounce and throttle functions from lodash. I know that solid-primitives has its own createDebounced & createThrottled counterparts, but I don't understand why not just use the library that everyone will probably have in their codebase anyway, even if they are not using them directly. It's the same idea as with primitives, using what's already built and tested is better than writing your own.


davedbase:
Thanks for bringing this to solid-prims! Lots of good points here.

My opinion is that while Solid strives to make things simple, I'd argue that solid-primitives attempts to make things easier. Making things easy is everything from creating reactive wrappers around browser APIs or exposing helpful utilities. So this effort aligns with the purpose of this project.

I haven't read through the source in detail but I'm wondering if these reactive helpers are in line with Solid's core philosophy which is a question @ryansolid is better suited to answer. I'd like to ensure this project acts as an extension as opposed to an alternative of that vision/direction.

I hear you regarding the debounce and throttle methods. We've debated it as well. I've decided we will try to tightly control as many dependencies as possible, so for no primitives have any. The implementations we have for throttle and debounce strive to be smaller, more performant wherever possible and more appropriate implementations for our needs. As an example lodash.debounce is packaged with a number of unnecessary options that we don't leverage at all in any prims ie. leading, trailing, maxing etc. This wont be a strict rule, if there are extremely good cases for utilizing an external package, we will.

We're trying to nurture something more tightly insulated and self-dependent. I hope others can respect that vision/desire.

davedbase:
Also there's a lot packed into this one index, can we separate the utilities into separate files and export them from index to make it a lot more readable?

thetarnav:
That's the plan. I was just experimenting so it was easier to have them I'm one file.

davedbase:
All good! I shared this thread with the Solid Ecosystem team btw. So pulling in some more heavy opinions =)

thetarnav:
Regarding these "unnecessary options" in lodash, while you don't necessarily use them in other primitives, I think they are quite helpful in some situations, I personally use maxWait all the time, but that's just me.
But it's maybe not a bad idea to just have a "default" simple version just to power all those primitives and typical use-cases, while people that want to have something more, can always build their own.

davedbase:
Yeah precisely. I'd personally prefer shipping light/tight primitives as opposed to all the extra options that a specific heavy use audience would require. Kind of following that "do more with less" line. This isn't just my project though, I really want your and the broader communities opinions. I'll end up taking the path that is most rational and logically supported by the broader community.

Jutanium:
Hey, could you provide some examples where these watch primitives are useful (and working with `createEffect` isn't, or is cumbersome in comparison)?

I get that `watch` can allow you to state your dependencies up-front, but if that's the main goal, I would argue that Solid doesn't need them (to answer your first question) / we should probably stick with the Solid API for managing effects.

Honestly, as a Vue user, I try to use `watch` (and the Vue 3 versions) sparingly. That would apply to `createEffect` too, of course, as it's just the principle of keeping your side effects low in an otherwise declarative system. While I'm personally not enthused by watch helpers, I'm totally interested in being convinced.

atk:
I'm not exactly sure if these should be a part of solid-primitives. Their use case is pretty narrow (developers used to a less common API of Vue). But I'd like to take more time exploring them before I pass any verdict. How about you put them in a separate package while we take our time to consider including them or not?

thetarnav:
It's hard to provide good usage examples when it's a fresh concept in Solid even for me. I've only got to [use debouncedWatch once in my app](https://github.com/thetarnav/life.data/blob/daee53da5d834976d41da83651cc33601f34b175/src/logic/api.ts#L99), but more as a proof of concept, not a real use-case.
The "watch" is just a nice shortcut for `creareEffect(on(source, () => {}, { defer: true }))` that additionally can be manually stoped. 
The "debounced & throttled watch" works nicely with reactive sensors like [scroll](https://github.com/davedbase/solid-primitives/tree/main/packages/scroll-observer) or [mouse](https://vueuse.org/core/usemouse/#usemouse) inputs.
"whenever" also seems like a nice utility to watch things conditionally, while "until" turns that into a Promise, which can open a door for some specific use-cases, probably mostly used with [createFetch](https://github.com/davedbase/solid-primitives/tree/main/packages/fetch).
I rarely or never used the rest, so I'm pretty indifferent about them.
I definitely want to make these primitives not only for people coming from vue that also happen to use them. Therefore I think it's a good idea to make a separate package with the purpose of experimentation, and figuring out how these helpers can be changed to better fit Solid.

thetarnav:
The package has been made: https://github.com/thetarnav/solid-watch-primitives
I especially insist on playing with the `until`, because even if the rest of these "watch" helpers don't bring much value on their own, the `until` is there to actually solve a specific problem.

davedbase:
Awesome. Yeah the readme really lays out the uses and helps explain their benefits. OK this is great. Let's see if we can spark a conversation on Discord. I'm personally seeing cases where the throttle, debounce pausable could solve some of the trickier reactive modelling challenges I've had recently.

davedbase:
Ok! So I think the ecosystem team has formulated an opinion. I even got some feedback from Ryan.

First, from my perspective when I first saw your PR I felt really at odds with it because `watch` totally through me off. If didn't feel Solidlike™ (lol). It really isn't super clear what the use is. After digging into the implementation and talking through the uses with the group it feels like these are a cross between reactive utilities but also mostly extensions to Effect.

After a bunch of banter I pulled a WWRD (What Would Ryan Do) and he responded with:

```
Yeah i like that createRoot is a explicit opt out of the flow. Having stuff live with different lifecycles inside is fine but don't want to push people towards this.  It is like if CSS defaulted to position: absolute 
...
I like the idea of throttle debounce etc.. but designed on the way I did since we have lots if different computations. I figured these behaviors could be mixed and matched without a dozen new create_______. I suggest looking at the now deprecated solid-rx  Obviously some things deserve create_ but just categorizing whats a transformation vs whats a new functionality.
```

You can find the README for solid-rx here: https://www.npmjs.com/package/solid-rx. I really want to focus on Ryan's comment related to this being similar to the CSS property being absolute by default. It's a really apt analogy for the concern we have with this inclusion.

After studying this a bit I think we definitely concluded that there's utility for these primitives. I think we need to rename and align them to Solidify them. Here's my own suggestion, although take it with a grain of salt. Since most of these are effect-related utilities perhaps:

* createPausableEffect
* createThrottledEffect
* createDebouncedEffect
* createEffectOnce
* createConditionalEffect

In regards to the other utilities those should remain similar to how `on` is treated. Without a create or use keyboard. Perhaps `watch`, `until` and `whenever` remain as named.

I would rename the whole package to `effects` as opposed to watch for that reason.

davedbase:
I'm almost wondering if solid-rx could/should find a home in solid-primitives as an `rx` package.

thetarnav:
Yeah, I know what Ryan means with the createRoot, It felt dirty to do it that way, but I found no other way to permanently dispose of an effect, if it isn't wrapped in one. Maybe that's one of the things that should be changed. I mean that the default `watch` wouldn't be wrapped in createRoot, and it couldn't be stopped. This way we wouldn't be pushing something that is a "bad practice" of sorts but could still make a `stappableWatch` if needed for some usecases (like watchOnce).

When it comes to naming them as "createDebouncedEffect" and alike, there is a problem that they still would need to be provided an explicit source, and so naming it that way would suggest that it is literally a debounced createEffect, when the usage is more like createEffect(on(...)).
This may not sound like much but I think it would cause some confusion. Like: *if it is a createEffect then why do I need to add source?*
Additionally, some "watchers" require the `{ defer: true }` to be set on the `on` method. For example, `watchOnce`, if it wouldn't have the defer set, it would fire callback and delete itself on the initial setup. So this is another: *if this is a createEffect then why doesn't it run on the setup?*

Therefore, I think we should stick with the "watch" keyword, just to emphasize that it isn't just a createEffect. But I'm down to rename them  to things like: `createWatch`, `createDebouncedWatch`, `createWatchOnce`, ... *(although it's weird to have two verbs together)*
Or: `createWatcher`, `createDebouncedWatcher`, `createWatcherOnce`, ...

Also it would be nice I guess to have a more functional option, like:
* `createWatch(debounced(source, fn, time))`
* `const [toggle] = createWatch(pausable(source, fn))`
* `const stop = createWatch(stoppable(source, fn))`

Although that may look cool, it is effectively the same as just having them individually named, and I don't know if it is technically possible, when some of the watchers need to be wrapped in createRoot, and some not. But idk, will experiment with it later.

And regarding `rx`, solid-primitives should definitely have some primitives for it ([vueuse also haves some :)](https://vueuse.org/rxjs/readme.html)), but that's out of my reach.

Jutanium:
I think this PR/package really has two features: `watch`, allowing you to easily state effect dependencies explicitly, and the set of utilities that build off of those.

Based off of our conversations with Ryan, it doesn't seem that `watch` is idiomatic in Solid. And I don't think that all of the effects utilities need to be based off of `watch`. That's what I see in @davedbase's suggestion: you can keep the effects utilities without having them all be `watch` variants. For the ones that do require `on` with `{defer: true}` to work, the value of using `watch` becomes a lot more clear. For the rest, I agree with the composable version you lay out (if it does indeed work) and it allows you to decouple the effect utilities (which seem to me to be useful whether or not you want to state your dependencies explicitly) from the `watch` rephrasing.

thetarnav:
Alright, I have made a lot of changes to the API, with the goal of making it feel more functional and composable. And in the end, it actually works pretty well. You can try the new API [here](https://codesandbox.io/s/solid-watch-primitives-urb2u) and it is explained in detail in the [readme](https://github.com/thetarnav/solid-watch-primitives#the-usage), so I'll only leave this as a teaser:

```ts
const [count, setcount] = createSignal(0)

const { stop, resume } = createWatch(
   stoppable(
      pausable(
         debounce(
            whenever(
               atMost(
                  () => count() % 2 === 0,
                  () => {
                     console.log('is even:', count())
                     if (count() % 2 !== 0) stop()()
                  },
                  { limit: 8 },
               ),
            ),
            { wait: 300 },
         ),
         { active: false },
      ),
   ),
)

resume()
```

davedbase:
Hey! Sorry for the delay. Been getting caught up in work this week. I think this new API is a LOT more interesting. I like how there's Watch and then Filters modify it. Very neat. I posted this on our Ecosystem chat for fresh feedback =)

davedbase:
Oh this is just me personally but I still do feel like since Watch is just `createEffect(on())`, at its core these are pretty much Advanced Effect with Filterable options. If one of your goals is to make this accessible to a broader audience you'll have to teach them what Watch is. Whereas as an Effect or Advanced Effect it's easier for them to very easily reason them.

thetarnav:
Making these compatible with both `createEffect` and `createEffect(on())` was my ideal, before I started working on them. But it is just not possible, so this is why I went with `watch`. Even though it's not idiomatic in the community, at least it's short and well represents the behavior. Of course, `createWatch` could be renamed to `createAdvancedEffect` or `createFilteredEffect`, it still would require to have dependencies explicitly set, so it would feel off. At least for me. I know that `useEffect` in React natively requires stated dependencies, whereas Solid and Vue don't, so maybe this is the cause of different perspectives.
But as long as this is only a dispute about naming, it's okay, names can be changed easily. We just need a decision :)

davedbase:
We're definitely on the same page. I did ask Ryan to have his eye on it. I'm hoping he finds a moment and gives us implementation feedback.

In regards to naming, yeah that's just my personal opinion right now. No one else has made the same remarks to be honest. I mean I see both naming options as being viable. Watch is an effect tracking a source dependency, because it has that property of explicit tracking, it makes sense to differentiate it from a regular effect.

I suppose my thought process is:

* Naming it createdFilteredEffect makes it much easier to grasp what the primitive does when first seeing it. createWatch means nothing to a Solid audience or beyond.
* Judging purely the language Watch may be similar to Observe. In solid-primitives I'm trying to focus the word Observable as being a distinct thing. ie. resize-observer, scroll-observer.

Then again:

* Calling it an effect overlaps with Primary Primitives closely which *might* be confusing, naming this differently creates a distinct separation in purpose/use/value
* Calling it Watch does keep it friendly to Vue users, whatever that's worth to us
* If our goal isn't to promote it's use a common pattern then making it a bit more obscure is beneficial.

Sorry for all the writing, whenever I think through tedious things like naming sometimes it helps me to express the logic.

Well let's see what Ryan says on implementation. This might be worth a Poll in the Discord! I've been looking for a reason to run one haha.

davedbase:
So I'm getting a lot of mixed reactions. I believe Ryan sees utility in certain elements here but in many ways we need to leave it up to the community to decide what "fits". Solid is still young and we need to leave a bit of room for interpretation and experimentation. I think my hesitance on this PR isn't that it's not "good" it's mostly that I'm not qualified to pass judgment on the effects this has to Solidlike paradigms or patterns. Then again who really is?

The official answer is core needs to iterate slowly but it can learn from and evolve from implementations and patterns in projects such as solid-primitives. If we leave NO room for experimentation then we're breaking the innovation process.

Since I have to have a verdict on this PR (I feel like this is a judge's summary of a case hahaha) I've decided to merge it. This whole topic is excellent and I'm so appreciative of it because it's forced me to consider how to handle similar situations in the future with the other maintainers. We'll implement a new acceptance process for incoming PRs. I'll be modelling it on TC39 (https://tc39.es/process-document/) or something similar. More details to come.

As for this PR @thetarnav thanks so much for your awesome work and effort. Did you want to give me the thumbs up when you feel this is ready for a Stage-1 release? :-)

thetarnav:
There is still a lot for me to do before the release can happen, especially testing it properly.
Also, I am getting more and more convinced that renaming it to createFilteredEffect might be a good idea since it evolved into a much more flexible primitive than it was initially. The whole "watch serving as a shortcut" doesn't apply here anymore, so might as well drop it.
Thank you for bringing me so much feedback from the community and yourself, it definitely helped me shape it into something better. A well-structured PR progress for new primitives definitely sounds cool, but for that, some templates would be necessary, at least for new people, like what, how, and why should be contained in a new premise, all the stuff about directives, and tests.
Also on a side note, I was thinking of removing the `until` from this module and making it separate, because as a primitive it solves a completely different problem. Then the module could be renamed to filtered-effect instead of watch.

davedbase:
I'm good with that as well. Ultimately this is a contribution that you're actively iterating on and understand far better than I do. I see utility in my own projects and am willing to utilize them but I don't want to be too involved in effecting your ability to flesh out the implementation. Basically you're closer to the problems than I am to render an opinion.

:-)

davedbase:
I was just looking at the README and this triggered a thought:
```
createFilteredEffect(
   once(counter, n => console.log(n)),
   { defer: true },
)
```
Calling it a Filter in this cause is a bit awkward maybe? I could be wrong. createCompositedEffect or createCompositeEffect maybe createDynamicEffect? Sorry not meaning to make this more complicated, just thinking aloud.

thetarnav:
Yeah I agree, the more you think about it the less word "filter" has sense in this context, these "filters" do more than just filtering xd.
The `createCompositeEffect` sounds actually nice. But what these "filters" then would be called? (in the docks)

davedbase:
Ooo I like that. Yes modifier is cool although filter still works. It's expressive enough. I'm indifferent.

I do like Composite Effect because technically you can continue chaining the behaviour. Technically createCustomEffect would work as well. Something about Composite Effects has a cool ring to it.

thetarnav:
I only couldn't test the modifiers, because of some weird bug:

test_export error

But they are built on `createCompositeEffect` and `createEffectModifier`, which are fairly tested.

davedbase:
I was wondering, to shorten the naming I think createModifier or useModifier might be alright?

As for the testing issue, it's tripping over debounce likely because debounce isn't export cjs?

thetarnav:
I can change its name if you insist, but I don't know if the name length matters that much for this primitive, as modifiers will probably be much more often used than created.
I'm not sure what is causing the issue, but lack of cjs is a good lead. Is there something I can do with it?

davedbase:
Ach, good point. Leave it as is, I wouldn't insist haha.

I assume we need to have a CJS build process in the debounce primitive. Actually TBH maybe we should export ESM, CJS, MJS etc. to cover all our bases. We only support ESM for most of the primitives. If you added it to the debounce and throttle build commands I can replicate amongst the rest of the primitives.

davedbase:
@thetarnav ok assuming you got the CJS output right, the new versions are up on npm and ready for you to reinstall as dependencies. Let me know how that goes!

thetarnav:
The debounce works! But the throttle hasn't been updated yet.

davedbase:
Haha I'm an idiot, I checked my bash log and I built but didn't run publish for throttle. It's there now, sorry.

thetarnav:
Okay, I've read the last conversation on discord and what Ryan said has caught my attention:
> I don't mean that. I mean what if I want a composite Memo, or Computed, or renderEffect.

It's definitely doable, I just need to make a `createCompositeEffect` equivalent for each of those. And I guess `createEffectModifier` will have to be `createModifier` after all :)
But what exactly is "renderEffect"? I've never seen it before.

davedbase:
Oh neat. I think we should change the folder name though to something more general now. ie. maybe simply `composites` since it covers most of the Primary Primitives.

He means createRenderEffect (https://www.solidjs.com/docs/latest/api#createrendereffect). The difference from createEffect seems to be it has a "render" but I'm also admittedly confused how that makes sense.

@Jutanium out of curiosity have you hit this section in the docs yet and have you be able to decipher it yet? lol.

davedbase:
How we feeling about the PR? :-) close to being able to merge?

thetarnav:
Yup, It's ready to be merged. I need some time to use it and get some grip on the new usecases to be able to make some meaningful changes. Right now I won't be coming up with anything more.

davedbase:
@thetarnav awesome. I'll merge this in as a Stage 2 (Experimental) release then publish version 1.0.0 to npm. Thanks! Yeah I'm really interested to see how this turns out for people.

Thanks again for the awesome work.

davedbase:
Published: https://www.npmjs.com/package/@solid-primitives/composites =) congrats!

PR #20, opened 2021-11-18
add commonjs build to debounce & throttle

thetarnav:
To solve the issue in #19 
This should work, but I don't know too much about building js, so I'm not totally sure.


davedbase:
This looks great. Basic test worked well, so gonna merge it.

PR #21, opened 2021-11-22
new date-difference primitives

thetarnav:
This brings two new primitives: `createDateNow` and `createDateDifference`.
The `createDateNow` provides an autoupdating `Date.now()`. And the `createDateDifference` is basically a [useTimeAgo](https://vueuse.org/core/usetimeago/) from vueuse.
Both don't use any dependencies, the basic formating is pretty much stolen from vueuse, and should work in 90% of the usecases. But I've provided an easy interface to use any custom formatter you want.

You can see the demo here: https://codesandbox.io/s/solid-date-difference-hjxui?file=/index.tsx

I'm a little bit concerned about the name, I've taken `createDateDifference` as it was named like that in the planned list. But something like `createTimeAgo` or `createRelativeDate` might sound a little bit better.

I've included the "development environment" in the `dev/` folder (it's basically just the demo from above), as I think it helps in the development process.

Also, what is the current story with including primitives as components to this library? Something like `createDateDifference` might work great as a component, kinda like [here](https://vueuse.org/core/usetimeago/#component) or [here](https://gitlab.com/catamphetamine/react-time-ago#srclastseenjs).

### DIFF
--- a/packages/date-difference/LICENSE
+++ b/packages/date-difference/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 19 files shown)


davedbase:
Awesome work! I think the naming is fine. I'm going to commit this though and grant you repo access directly since you're helping out so willingly!

PR #22, opened 2021-11-23
add a bootstrap script

trusktr:
This PR

- removes lock files.
  - Libraries don't need these, only apps do. In practice, if we use lock files, we have less chance of catching in-range breaking changes in dependencies, and end users will experience such breakages before we notice them. These lock files are ignored by npm install or yarn install. Lastly, adding an npm shrinkwrap lock file will eliminate the ability for semver dependency version resolution to work properly, leading to issues like dependency forks (i.e. multiple of the same package with differing patch or minor version numbers even though they are compatible, leading to issues like libraries not working due to duplicate singleton instances existing within an app).
- updates `@types/jest` to avoid a dependency fork (npm 7+ throws errors on these)
- add `lerna bootstrap` to package.json `install` so that after installing dependencies at the root, dependencies for all packages will be installed, and dependent packages will be automatically linked together for faster dev iteration.


atk:
I wanted to replace jest with uvu anyways.

davedbase:
@trusktr mind resolving the conflicts? Then I can merge! =)

packages/event-listener/test/index.test.ts
```diff
@@ -7,7 +7,7 @@ describe("createEventListener", () => {
       createRoot(dispose => {
         const target = new EventTarget();
         const testEvent = new Event("test");
-        createEventListener<{ test: Event }>(target, "test", ev => {
+        createEventListener<{ test: Event }, EventTarget>(target, "test", ev => {
```
atk, line 10: That shouldn't be necessary.

trusktr: Hi @atk :). I'm seeing this in VS Code:

Screenshot from 2021-11-29 14-15-50

In general we need a pass at ensuring Solid and primitives are fully aligned on TS configuration, so that we'll get the same errors in build as with VS Code (currently not the case). That'll help clear things up.

This will involve making Solid follow a conventional TypeScript project layout and unduplicating some of its type definitions.

PR #23, opened 2021-11-23
add test script to run all tests

trusktr:
This

- adds a script to run all tests in all packages from the root of the repo.
- fixes several type errors.

Tests are not yet passing, there is a runtime error.

To try it, run the following at the root of the repo:

```sh
npm install
npm test
```

You will see tests exit with a runtime error inside the `event-listener` tests.

PR #24, opened 2021-11-23
add automated testing

trusktr:
- adds a build script at the root to build all packages at once
- run build and test of everything with GitHub Actions


trusktr:
This currently fails with the same runtime error as in #23

packages/gestures/package.json
```diff
@@ -14,8 +14,8 @@
   "scripts": {
     "start": "vite serve dev --host",
     "dev": "yarn start",
-    "build": "tsup",
-    "test": "uvu -r solid-register"
+    "build": "echo noop build",
```
atk, line 17: Why noop build?

thetarnav: Because it was breaking, and the package is not finished, so there is no value in having the build script anyway. It would be set back to normal once the package is completed.

thetarnav:
Why are we testing on Node 15 anyway?

davedbase:
How are we looking on this PR folks?

thetarnav:
I don't mind it, but since out testing story is far from perfect, I can already imagine all the tests that fail for some weird reason, causing us to do more very useful test-debugging work... But maybe it could be a useful insight even now, it the end all the checks did turn green.

PR #31, opened 2021-11-29
unbuild utils

thetarnav:
As we search for a perfect building solution, this one is at least simple, works, and lets us focus on making primitives without worrying if they'll break or not.
I'm only applying it to the utils for now, but it can be easily ported to the rest.


thetarnav:
bye bye cruel world

PR #33, opened 2021-12-05
New package: active-element

thetarnav:
This package brings two simple, new primitives: `createActiveElement` and `createIsElementActive`.

[The DEMO](https://codesandbox.io/s/solid-primitives-active-element-q4kul?file=/index.tsx)

The `createActiveElement` is basically a reactive `document.activeElement`. It makes is easier to deal with focus & blur events on multiple elements in a reactive way. This is the API:
```ts
const [activeEl, { stop, start }] = createActiveElement()
```
`createIsElementActive` extends the `createActiveElement` by focusing on a single element. It returns a boolean accessor.
```ts
const [isFocused, { stop, start }] = createIsElementActive(element)
```

Right now the `createActiveElement` primitive is handling event listening itself, but I'm changing that to use the event-listener primitive in [fix-even-listener branch](https://github.com/davedbase/solid-primitives/tree/fix-event-listener) (it wasn't working with current).

Also this primitive uses the utils package which I think wasn't actually republished with the new exports.


atk:
createIsElementActive could be double useful as a directive if it optionally supports a second argument that is an `Accessor<(isActive: boolean) => void>`.

thetarnav:
Alright, I've added an `isElementActive` directive. I know that you meant to add an option for `createIsElementActive` to be used as a directive. But I think the logic is better as separated, plus it's a little bit nicer for the user to drop the "create" in jsx.
```tsx
import { isElementActive } from "@solid-primitives/active-element";

const [active, setActive] = createSignal(false)

<input use:isElementActive={setActive} />
```

packages/active-element/README.md
```diff
@@ -0,0 +1,86 @@
+# @solid-primitives/active-element
+
+[![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg?style=for-the-badge)](https://lerna.js.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/active-element?style=for-the-badge)](https://bundlephobia.com/package/@solid-primitives/active-element)
+[![size](https://img.shields.io/npm/v/@solid-primitives/active-element?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/active-element)
+
+- [`createActiveElement`](#createActiveElement) - A reactive `document.activeElement`. Check which element is currently focused.
```
davedbase, line 7: I'm trying to move the docs away from listing the primitives at the top and treating the how to use as a way to list them with a description. Basically createActiveElement and createIsElementActive can be an h3 header within how to use.

packages/active-element/README.md
```diff
@@ -0,0 +1,86 @@
...
+  getter: Accessor<null | Element>,
+  actions: { stop: Fn; start: Fn }
+];
+```
+
+## `createIsElementActive`
+
+Pass in an element, and see if it's focused.
+
+### Usage
```
davedbase, line 36: Let's change this to "How to use it"

packages/active-element/test/index.test.ts
```diff
@@ -0,0 +1,94 @@
+import { createRoot } from "solid-js";
+import { suite } from "uvu";
+import * as assert from "uvu/assert";
+import { createActiveElement, createIsElementActive } from "../src";
+
+let listiners = 0;
```
davedbase, line 6: Minor spelling error listiners should be listeners. :-)

davedbase:
Put in some minor comments but everything else looks great here. Let me know when I can merge. =)

packages/active-element/README.md
thetarnav: Yeah, I saw that, but I like the listing because it lets put links to specific sections, making it work as a table of contents of sorts. And it's easier to scan that small section alone, than having to look through the whole readme.

davedbase: Fair enough. I did remove it from the other packages but I'll think of a nicer way to manage this.

thetarnav:
Okay, I've fixed the issues.

PR #34, opened 2021-12-05
New Package: mouse

thetarnav:
This brings 4 new primitives. Most of them do similar things, but I decided to split them instead of making one/two bigger primitives. This makes them simpler and more composable. The primitives:
- `createMousePosition` - Listens to the global mouse events, providing a reactive up-to-date position of the cursor on the page.
- `createMouseToElement` - Provides an auto-updating position relative to a provided element. It can be used with existing position signals or left to get the current cursor position itself.
- `createMouseInElement` - An alternative to `createMouseToElement`, that listens to mouse (and touch) events only inside the element. Provides information of position and if the element is being hovered.
- `createMouseOnScreen` - Answers the question: _Is the cursor on screen?_

[DEMO](https://codesandbox.io/s/solid-primitives-mouse-p10s5?file=/index.tsx)

[README](https://github.com/davedbase/solid-primitives/tree/mouse/packages/mouse#solid-primitivesmouse)

Similar to #33 this package also has utils dependency and will be updated to use the new event-listener.


packages/mouse/dev/index.tsx
```diff
@@ -0,0 +1,87 @@
+import {
+  createMousePosition,
+  createMouseOnScreen,
+  createMouseToElement,
+  createMouseInElement
+} from "../src";
+import { Component, createSignal, Show } from "solid-js";
+import { render } from "solid-js/web";
+import createRaf from "@solid-primitives/raf";
```
davedbase, line 9: Not that this matters but should keep it as createRAF for consistency with the actual primitive. :-p

davedbase:
This looks so good! Let me know if we can merge it in =)

thetarnav:
Everything should be fine now.

PR #36, opened 2021-12-06
event-listener primitive improvements

thetarnav:
This:
- Changes but simplifies the event-listener logic. While keeping the same functionality.
- Improves the user experience with types, they are properly inferred, based on the event name. Also, the usage of type generics in case of custom events is optional.
So this isn't causing type errors:
image
- Changes the returned functions: from `[add, remove]` to `[stop, start]`. So now it's easier to clear the event listeners.
This doesn't remove the option to edit the targets, now a reactive accessor can be passed as a target property, so to change the targets, all it takes is to edit the input signal.
- Add a new *component primitive*: `GlobalEventListener` that can be used to add global (window) event listeners in the jsx section of the component, rather than inside logic. This helps to keep all the event-listeners in one place, which can sometimes improve readability.
```tsx
  const [listen, setListen] = createSignal(true);
  const [mouse, setMouse] = createSignal({ x: 0, y: 0 });

  <Show when={listen()}>
    <GlobalEventListener onmousemove={({ x, y }) => setMouse({ x, y })} />
  </Show>
```
- Applies the new event-listener to mouse and active-element packages, as I couldn't use it before. (so they should also be republished)

I've already changed the version of the package to 1.3.0, so that I could set that version in mouse and active-element.
@atk You should probably review this as you've worked on event-listener last.
Also here is a new demo: https://codesandbox.io/s/solid-primitives-event-listener-elti5


atk:
It's not exactly the same functionality anymore; instead of "add" and "remove", there are now "stop" and "start" (feels out of order), which greatly reduces the usefulness; stopping the Event just includes replacing the event handler with `null`, which is simple enough. I like the idea of having different interfaces like createGlobalEvent, though.

packages/event-listener/src/index.ts
```diff
@@ -1,128 +1,135 @@
...
+> = (event: EventMap[EventName]) => void;
+
+export type EventListenerReturn = [stop: Fn, start: Fn];
 
-export type CreateEventListenerReturn = [
-  add: (el: EventTarget) => void,
-  remove: (el: EventTarget) => void
-];
+export type EventListenerDirectiveProps =
+  | [string, (e: any) => void, boolean | AddEventListenerOptions]
```
atk, line 29: you can use a named tuple here

```ts
[name: string, handler: (e: any) => void, options?: AddEventListenerOptions]
```

This is simpler to understand and also allows the use of optionals.

packages/event-listener/src/index.ts
```diff
@@ -1,128 +1,135 @@
...
- * // or as a directive
- * <MyButton use:createEventListener={() => ['click', () => console.log("Click")]}>Click!</MyButton>
- * // you can provide your own event map type:
- * createEventListener<{ myCustomEvent: Event }, Window>(window, 'myCustomEvent', () => console.log("yup!"));
- * ```
+ * const [stop, start] = createEventListener(
+ *   document.querySelector('.myEl'),
+ *   'click',
+ *   e => { ... }
+ * )
```
atk, line 60: Please provide the other examples as well. The previous example for the directive had a superfluous function wrapper, as the JSX expression is already wrapping its content.

packages/event-listener/src/index.ts
```diff
@@ -1,128 +1,135 @@
...
-    targets.forEach(remove);
-  });
-  return [add, remove];
-};
 
-export default createEventListener;
+  createEffect(start);
+  onCleanup(toCleanup.execute);
+
+  return [toCleanup.execute, start];
```
atk, line 113: This is not the same functionality, it contains [stop, start] (why not the other way around?) instead of [add, remove]. We should at least discuss this. Since we allow rebinding, using `null` as a handler already allows stopping and you can add a handler back later. On the other hand, the target can be an accessor as well and thus, add and remove are superfluous as well.

So I would probably vote for another option: to make the last event available as reactive signal, thus basically making this doing the work of rxjs' `fromEvent`.

packages/event-listener/test/index.test.ts
```diff
@@ -1,76 +1,99 @@
...
+import { dispatchFakeEvent } from "./setup";
+import { test } from "uvu";
 import * as assert from "uvu/assert";
-import { createRoot, createSignal } from "solid-js";
-import { createEventListener, EventListenerProps } from "../src/index";
+import { createRoot, createSignal, onMount } from "solid-js";
+import { createEventListener, EventListenerDirectiveProps } from "../src";
 
-const test = suite("createEventListener");
+test("it will add an event", () => {
```
atk, line 7: Removing the suite means that you no longer see the name of the tested function.

atk (changes requested):

packages/event-listener/src/common.ts
```diff
@@ -0,0 +1,14 @@
+import { Fn } from "@solid-primitives/utils";
+
+export const createCallbackStack = () => {
```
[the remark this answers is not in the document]
thetarnav: the createCallbackStack? yeah, it's a utility that I'm using in event-listener, mouse, and active element. The goal is to later place it in the utility package. Right now it's just for extracting the repetitive logic from the primitive function.
I'm fine with copy-pasting it wherever I'm using it until I would be satisfied enough to put in the utilities (don't want to change its API once it's in there).

packages/event-listener/src/component.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @see https://github.com/davedbase/solid-primitives/tree/main/packages/event-listener#GlobalEventListener
+ *
+ * @example
+ * <GlobalEventListener onMouseMove={e => console.log(e.x, e.y)} />
+ */
+export const GlobalEventListener: Component<EventAttributes<null>> = props => {
+  if (isClient) {
+    Object.keys(props).forEach(attr => {
+      if (!attr.startsWith("on")) return;
```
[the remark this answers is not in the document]
thetarnav: This is interesting. I guess I wanted to limit it to only events by design because it is a "GlobalEventListener".
If it would accept other attributes as well that to what they should be attached to? The body tag, HTML tag? Because I don't think they can be attached to the window object.
It is an interesting idea, but maybe making a separate component would be better. And keep GlobalEventListener only for window/document events.

packages/event-listener/src/index.ts
thetarnav: My idea here was that maybe in jsdoc we should place only a simple example, but provide a clickable link to the primitive readme with a deeper explanation. But that's true if this is supposed to show the API, even if briefly, then I should add the directive usage back in there.

packages/event-listener/src/common.ts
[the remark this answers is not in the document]
atk: I think it's already generic enough that it doesn't trigger the AHA rule.

packages/event-listener/src/component.ts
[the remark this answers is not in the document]
atk: You could rename it to `WindowEventListener` and `DocumentEventListener`.

packages/event-listener/src/index.ts
thetarnav: I've put [stop, start] in this order, because 1. stop will be more often used than start, and 2. you first need to first stop the events to be able to start them since they start automatically.
I switched add and remove because I needed an easy way to remove the event listeners, not by just passing a nop as a handler, but removing event listeners from elements.
I said that it has the same functionality because modifying the targets is still possible since the target arg is now reactive:
```ts
  const [targets, setTargets] = createSignal([
    document.getElementById("123"),
    document.getElementById("abc")
  ]);

  createEventListener(targets, "mouseenter", () => {});

  const add = (el: HTMLElement) => setTargets(p => [...p, el]);
  const remove = (el: HTMLElement) =>
    setTargets(p => {
      const copy = p.slice();
      copy.splice(p.indexOf(el), 1);
      return copy;
    });
```
This may be a bit lengthy, but it's a common pattern to use in Solid. Since signals are immutable, this is just how you handle arrays.
I've chosen this approach because it seemed more aligned with reactivity and solid patterns.
But I guess the same could be said about stop and start. Although this is a little bit different because these are supposed to be permanent actions, If you want to disable handling events temporarily, then putting that logic into a handler is fine.
So maybe we should move away from this too, and instead, rely on Solid's patterns again to provide solutions for users:
```ts
  const stop = createRoot(stop => {
    createEventListener(targets, "mouseenter", () => {});
    return stop;
  });
```
This way the space would be left for the `lastEvent` as a returned value:
```ts
const lastEvent = createEventListener(targets, "mouseenter", () => {});
```

atk: I would implement `remove` as

```ts
const remove = (el: HTMLElement) => setTargets(t => t.filter(target => target !== el));
// and this allows us also to remove/stop the events easily:
const clear = () => setTargets([]);
```
since we need a new reference for the array anyways to propagate the change.

@davedbase what's your opinion? I copied the "add" and "remove" methods from your initial implementation. I think we should rather make sure that w handle the target array reactively (do not needlessly remove and add the same listener to the same target).

If we want to add more methods, we could still use a `[lastEvent, methods]` pattern, but I think that's overengineering.

PR #37, opened 2021-12-08
event-props: initial commit

atk:
Based on our last discussion, I started a small experiment to create event props/store in a primitive for ease of reactive handling. How do you like it?


packages/event-props/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   }
+ * });
+ * 
+ * <div {...eventProps}>Click and drag on me</div>
+ * ```
+ */
+export const createEventProps = <Names extends EventNames>(
+  ...names: Names
+): [EventStore<Names>, EventProps<Names>] => {
+  type EventStore = Partial<Pick<HTMLElementEventMap, typeof names[number]>>;
```
davedbase, line 34: I find it strange to declare types within, shouldn't this be outside the function and exported?

atk: Ah, that somehow snuck in. Sorry, fixed it.

PR #39, opened 2021-12-12
Update README.md

nksaraf:
It looks like the API has changed to return just the `refCallback` so fixing the docs to reflect that change


thetarnav:
That seems about right, thanks for the change!

PR #44, opened 2021-12-28
New Stage-2 Package: "event-bus"

thetarnav:
There probably will be some changes in the future (some ideas like `toPromise` to explore). But I think it is mergeable as stage-2.

There are some updates to the `utils` package as well, so I already changed its version to a new minor `0.1.0`. I added 2 new entries: `/fp` and `/setter`. /fp is for functions that help deal with immutable data, and /setter is the same but easier usage with `setState()`. 

For example:
```ts
import { filterOut } from "@solid-primitives/utils/setter";
const [stack, setStack] = createSignal<V[]>([]);
setStack(filterOut(value))
```


thetarnav:
Update: added `toPromise` util function, it can be used like this:
```ts
const emitter = createEmitter<string>();
const event = await toPromise(emitter.listen);

// can be used together with raceTimeout from @solid-primitives/utils
import { raceTimeout } from "@solid-primitives/utils";
try {
  const event = await raceTimeout(toPromise(emitter.listen), 2000, true, "event was too slow");
  // if event is quicker:
  event; // => string
} catch (err) {
  // if timeouts:
  console.log(err); // => "event was too slow"
}
```

PR #47, opened 2022-01-02
navigator.onLine signal

KrofDrakula:


davedbase:
Thanks so much for this contribution. I have a couple minor suggestions!

- Can you add a server.ts with noop? We have a pattern in Prims that if the server doesn't need to do anything then just feed it a basic noop with the same interface to ensure SSR doesn't fail.
- What do you think about renaming this to `connectivity`? I'm trying to organize the packages in a manner that the primitive the total focus. Within a single package a number of helpful primitives can be composed on top of each other.
- This is mostly my own preference but maybe rename whenOnline, whenOffline to just online, offline to shorten it :-)
- Can you run `yarn ls updateReadme` in the root to update the main readme?

=)

KrofDrakula:
@davedbase Done. Couldn't find an entry for this specifically in the root README, but I think this covers your comments.

thetarnav:
I've added a fixed config for double-entry package here: https://github.com/KrofDrakula/solid-primitives/pull/1
Outside of that, the package looks good. Although you can add a note to readme about this online/offline thing is rather deceiving, basically what you told in the discord that it's just a browser's guess.

KrofDrakula:
@davedbase I've merged your changes and renamed the dir to `connectivity`, and linked the `navigator.onLine` to the MDN. I think this should make it releasable now.

thetarnav:
Okay, I'm gonna merge that since all the issues were resolved and the package looks nice.
@davedbase You can publish whenever you got some time :)

btw I'm thinking that the packages list in the root readme shouldn't really be updated in the PRs because it may cause unnecessary conflicts if multiple packages were to be made at the same time.

PR #59, opened 2022-02-13
New Control Flow package: Keyed

thetarnav:
This brings some Control Flow primitives and components that require specifying explicit keys to identify or rerender elements.
The component `<Key>` is for rendering a list mapped by explicit key, rather than ny reference to whole item or it's index in the input list. This component is powered by `mapKey` primitive, and accepts `fallback` prop, similarly to Solid's For, and Index. The map function in this component have available two arguments: item and index, that are **both signals**. The index signal is not created, if the mapFn is not using it.
```tsx
<Key each={items()} by={item => item.id} fallback={<div>No items</div>}>
  {(item, index) => <div data-index={index()}>{item()}</div>}
</Key>
```
prop `by` can also be an object key

```tsx
<Key each={items()} by="id">
```

The `Rerun` component causes the children to rerender when the `on` key changes. Equivalent of `v-key` in vue, and `{#key}` in svelte. Can be used for transition-between-state animations.
```tsx
<Transition name="your-animation" mode="outin">
  <Rerun on={count()}>
    <button onClick={() => setCount(p => ++p)}>{count()}</button>
  </Rerun>
</Transition>
```
[Key DEMO](https://codesandbox.io/s/solid-primitives-keyed-key-demo-gh7gd)
[Rerun DEMO](https://codesandbox.io/s/solid-primitives-keyed-rerun-demo-14vjr)
[README](https://github.com/davedbase/solid-primitives/tree/new-pkg-keyed/packages/keyed#readme)


packages/keyed/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+ * Causes the children to rerender when the `on` changes.
+ * @see https://github.com/davedbase/solid-primitives/tree/main/packages/refs#Rerun
+ */
+export const Rerun: Component<{ on: any }> = props => {
+  const key = createMemo(() => props.on);
+  return createMemo(() => {
+    key();
+    return props.children;
+  });
+};
```
TrueMrMaverick, line 146: Hi!
Do we really need 2 memos to implement this feature?
Personally, I would rather go with something like this:
``` 
export const Rerun: Component<{ on: Parameters<typeof on>[0] }> = props => {
  return createMemo(on(props.on, () => props.children));
};
```
With this, we will follow typings of the first argument of the `on` function for overall consistency (and also it will be possible to pass an array of dependencies).
Otherwise, if you do not like this approach for the `on` prop and still want to pass it as a value I suggest the second approach to remove additional memo:
```
export const Rerun: Component<{ on: any }> = props => {
  const key = typeof props.on === "function" ? props.on : () => props.on;
  return createMemo(on(key, () => props.children));
};
```
With this approach, it will be possible to pass both values and functions as in your initial suggestion.

thetarnav: Interesting, I knew there are some optimisations to be made, so thank you for pointing that one out. Or rather two.
I think I went with a separate memo for the key to not rerun the components, if the value didn't actually change, but has caused computation to run for some reason. Similarly how the Show component memoizes the condition.
But I wonder if that may be actually what the user wants, like you could have some special signal with `equals: false` used to re-render the component, where the actual value doesn't matter.
I'm down to changing it though, the future will tell what's the better approach anyway.
As for enabling functions to be passed to `on` I'm totally down. I've completely forgotten about that possibility. Thanks.

thetarnav: Litte update:
```tsx
<Rerun on={[count, className]}>
  {([count, className]) => (
    <button class={className} onClick={() => setCount(p => ++p)}>
      {count}
    </button>
  )}
</Rerun>
```

thetarnav:
Package published: [@solid-primitives/keyed](https://www.npmjs.com/package/@solid-primitives/keyed)

PR #61, opened 2022-02-19
New package: Context

thetarnav:
Adds very basic (at the moment) package focused around Solid's Context API. Right now the only primitive here is `createContextProvider` which helps with inferring types of created context, so they don't have to be typed up-front. And generally simplifies the creation of contexts.
[The readme](https://github.com/davedbase/solid-primitives/tree/context/packages/context#solid-primitivescontext)


packages/context/README.md
```diff
@@ -0,0 +1,80 @@
+# @solid-primitives/context
+
+[![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg?style=for-the-badge)](https://lerna.js.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/context?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/context)
+[![version](https://img.shields.io/npm/v/@solid-primitives/context?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/context)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fdavedbase%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/davedbase/solid-primitives#contribution-process)
+
+Primitives simplifying or extending the SolidJS Context API.
```
edemaine, line 8: I would rather describe this as "simplifying the creation and use of SolidJS Context Providers" or something along those lines.

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **[`createContextProvider`](#createContextProvider)** - Create the context provider component & useContext function with types inferred from the factory function.
```
edemaine, line 10:
```suggestion
- **[`createContextProvider`](#createContextProvider)** - Create a Context Provider component and useContext function, with types inferred from the factory function.
```

edemaine: Also, this is repeated verbatim below. Perhaps remove one of the duplicates?

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+yarn add @solid-primitives/context
+```
+
+## `createContextProvider`
+
+Create the context provider component & useContext function with types inferred from the factory function.
+
+### How To use it
+
+#### Import
```
edemaine, line 26: These two headings, or at least the first one, seem like overkill to me. (Also `How To use it` is inconsistently capitalized.)

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+// get the context
+const ctx = useCounter();
+ctx?.count(); // => 1
+```
+
+Or provide an fallback to be able to destructure safely.
+
+```ts
+const { count } = useCounter() ?? { count: () => 0 };
+```
```
edemaine, line 69: `createContext` allows providing a default value as a second argument, instead of needing to do this every time you call `useContext`. Perhaps this default could make sense as an optional second argument to `createContextProvider`?

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+#### Import
+
+```ts
+import { createContextProvider } from "@solid-primitives/context";
+```
+
+#### Creating the context provider and use-primitive
+
+Types of the context are inferred from it's factory function. Factory function will run when the provider component in executed. It takes the provider component `props` as it's argument, and what it returns will be available in the contexts for all the underlying components.
```
edemaine, line 34:
````suggestion
```js
function createContextProvider<T, P extends Record<string, any>>(
  factoryFn: (props: P) => T
): [provider: Component<P>, useContext: () => T | undefined]
```

Given a factory function, `createContextProvider` creates a SolidJS Context and returns both a Provider component for setting the context, and a useContext helper for getting the context. The factory function gets called when the provider component gets executed; all `props` of the provider component get passed into the factory function, and what it returns will be available in the contexts for all the underlying components. The types of the provider props and context are inferred from the factory function.
````

File: packages/context/README.md [hunk 6]
```diff
@@ -0,0 +1,80 @@
...
+
+```ts
+const [CounterProvider, useCounter] = createContextProvider((props: { initial: number }) => {
+  const [count, setCount] = createSignal(props.initial);
+  const increment = () => setCount(count() + 1);
+  return { count, increment };
+});
+```
+
+#### Place the Provider
```
edemaine, line 44:
```suggestion
#### Set Context using the Provider
```

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+The context will by default be `T | undefined`.
+
+```ts
+// get the context
+const ctx = useCounter();
+ctx?.count(); // => 1
+```
+
+Or provide an fallback to be able to destructure safely.
```
edemaine, line 65:
```suggestion
Or provide a fallback to be able to destructure safely.
```

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+```ts
+const { count } = useCounter() ?? { count: () => 0 };
+```
+
+## Changelog
+
+<details>
+<summary><b>Expand Changelog</b></summary>
+
+0.0.100
```
edemaine, line 76: Why this initial version number?

packages/context/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Initial realease of the context package.
```
edemaine, line 78:
```suggestion
Initial release of the context package.
```

edemaine (commented):
The tool looks really useful here, though I wonder whether a default makes sense.

Docs have a lot of typos and could be better organized. I've left a few direct comments; happy to review again.

packages/context/README.md
thetarnav: It takes a second parameter? I totally knew about that... 😁

packages/context/README.md
thetarnav: tradition 😃
actually @davedbase should know, I'm just following the flow

packages/context/README.md
thetarnav: It's repeated on purpose, although it can look silly right now as there aren't any other primitives in this package. For example this is how it looks when there are more primitives: [event-listener#readme](https://github.com/davedbase/solid-primitives/tree/main/packages/event-listener#readme)
Maybe this outline or a primitive is not needed in listing...

packages/context/README.md
davedbase: Minor versions mean pre-release or drafts. :-)

packages/context/README.md
edemaine: Maybe omit the outline until there are multiple primitives?

packages/context/README.md
edemaine: To clarify, it's an optional first parameter for `createContext`, and I'm proposing it as an optional section parameter for `createContextProvider`.

I'm trying to imagine practical use. Plausibly it'd be convenient for the default to be set by calling the factory function with no `props` argument. Or maybe the second argument should be the props argument to pass in... Not sure what's best here.

packages/context/README.md
edemaine: Just odd to start with 100. But not a big deal :)

packages/context/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+ *    <App/>
+ * </CounterProvider>
+ * // get the context
+ * const ctx = useCounter()
+ * ctx?.count() // => 1
+ * ```
+ */
+export const createContextProvider = <T, P extends Record<string, any>>(
+  factoryFn: (props: P) => T
+): [provider: Component<P>, useContext: () => T | undefined] => {
```
SonyStone, line 25: `=> T | undefined` 
Maybe remove that undefined assignment?

thetarnav: Will be adding a second argument for the defaults, if you provide them the type will switch to `T`. If you want a definite type without defaults, you're breaking the law, but I guess you could do it like this: *(probably should add this to the readme...)*
```ts
const useDefiniteCounter = () => useCounter()!
```

edemaine: Maybe the default argument should be of generic type `D`, and the resulting type is `T | D`? This would work even when the default argument is undefined, and allows providing some of the interface but not all in the default behavior. (I've wanted this in the past, though I might be weird.)

packages/context/README.md
thetarnav: This is actually quite a problem as the defaults shouldn't actually be needed since you should, well, provide the context if you are using it. It is definitely something to improve in the future, once somebody has the right idea, because yeah, passing an object of noop functions as default seams weird.
For now I think the simplest approach (what you've proposed actually) would be the best, as it still could be extended if needed.

packages/context/src/index.ts
thetarnav: I sometimes do the same in similar fallback-needing apis, but the worst part of this solution is that you loose the type suggestions/warnings. So I would leave that to the user to implement, for example:
```ts
const useCounterOrSomethingElse = () => useCounter() ?? fallback_with_different_type
```

thetarnav: I sometimes do the same in similar fallback-needing apis, but the worst part of this solution is that you loose the type suggestions/warnings. So I would leave that to the user to implement, for example:
```ts
const useCounterOrSomethingElse = () => useCounter() ?? fallback_with_different_type
```

thetarnav:
I think it about time to merge and publish it. @edemaine if you have any thoughts after applying your suggestions, would love to hear them. If not, I'll just merge it today, so that the package is available to be used.

packages/context/README.md
atk: I think you should use the default and allow to overwrite it by using a value in the provider, e.g. for testing purposes.

thetarnav: Don't know if it makes sense, if you use the provider the the defaults won't be used anyway, or do you want to overwrite the context values? If so, I thinks it's better to leave it up to the user to implement, instead of reserving the `value` prop name only for that. This way we keep the base `createContextProvider` as simple as possible, but there are always other primitives we could put in this package later.
For example one issue that the current `createContextProvider` has, is that the "jsx output" cannot be changed, so something like this becomes harder:
```tsx
// in the normal Provider component
return (
   <div>
      {some_additional_jsx_alongside_the_provider}
   </div>
   <Context.Provider value={value}>
      {props.children}
   </Context.Provider>
)
```

thetarnav:
Published: `@solid-primitives/context@0.0.100`

PR #65, opened 2022-02-25
New utilities package: Set

thetarnav:
This brings two reactive data structures: `ReactiveSet` and `ReactiveWeakSet`. Both are implemented as classes, but also have primitive function equivalents.
The way they work is very similar to object/array structures in [solid-proxies](https://github.com/Exelord/solid-proxies), maybe with an exception of using a class instead of a proxy. But I needed this for different project, so I've build this instead of waiting for sets to be implemented there :P
An snippet of usage:
```ts
const set = createSet([1, 1, 2, 3]);

// listen for changes reactively
createEffect(() => {
  set(); // => [1,2,3] (reactive on any change)
  set.has(2); // => true (reactive on change to the result)
});

// modify like a normal Set
set.add(4);
set.delete(2);
set.clear();
```
[The readme](https://github.com/davedbase/solid-primitives/tree/set-pkg/packages/set#solid-primitivesset)


thetarnav:
Allllrrrrright, time to merge this thing :)

davedbase:
Nice!

Exelord:
I just noticed this implementation! Great job! Looks like we went with quite similar approach. Would you be up for merging it into 1 version (feature complete) so that I could remove them from solid-proxies? 😊

thetarnav:
@Exelord Do you mean adding reactive Object and Array right? Can you make an issue for it so it's more clear?

Exelord:
That would be great as well! Actually I meant tweaking a bit your current implementation. I think I found a few issues that I caught in solid-proxies. Once they are equal, I would be happy to deprecate my proxies.

Also I have some thoughts on the TriggersCache. Its optimization for removing unused signals does help with memory consumption but it trades off operations performance. I was thinking about that, but then I realized it is more common re-access the property rather than access it once in a life time. In that case not cleaning up the signals would perform better inline with how the implementation would look like while using normal signals (without Map). Is that something you would reconsider?

PR #66, opened 2022-03-02
New package: Destructure

thetarnav:
This package provides primitives for destructuring reactive objects _– like props or stores –_ and signals into a tuple/map of separate accessors. The implementation is an updated version of the same primitives in [Alexis's solid-use](https://github.com/LXSMNSYC/solid-use).
Example: 
Spreading props:
```tsx
const ListItem: Component<{ title: string; label: string; highlight: boolean }> = props => {
  const { title, label, highlight } = spread(props);
  return <h2>{title()}</h2>;
};
```
Destructuring signals:
```ts
const [user, setUser] = createSignal({
  name: "John",
  lastName: "Brown",
  age: 25
});
const { age } = destructure(user);
age() // => 25
```
[The readme](https://github.com/davedbase/solid-primitives/tree/destructure/packages/destructure#solid-primitivesdestructure)


atk:
For me, this hits the sweet spot between being powerful and not being a foot-gun. This is exactly what I had in mind when we talked about this.

thetarnav:
Refactored the package a bit:
Previously separate `spread` and `destructure` are now a single function, that can be controlled by options object.
Options:
- `cache` - wraps accessors in `createMemo`, making each property update independently. _(enabled by default for signal source)_
- `lazy` - property accessors are created on key read. enable if you want to only a subset of source properties, or use properties initially missing
- `deep` - destructure nested objects 👀

destructuring deep objects:
```ts
const [store, setStore] = createStore({
  user: {
    name: "Johnny",
    lastName: "Bravo",
    age: 19
  }
});

const {
  user: { name, lastName, age }
} = destructure(store, { deep: true });

name(); // => "Johnny"
```

Also spreading lists is back in lazy mode (limited to available keys)
```ts
const [a, b] = destructure([1, 2], { lazy: true });
```

Added `wrapGetters` for reading object signals like a store -via individually cached getters. (not sure about the name here, if you have some better ideas, please, I need them)

```ts
const [getPlayer, setPlayer] = createSignal({
  name: "Raven",
  points: 350,
  weaponOfChoice: "Machete"
});
const player = wrapGetters(getPlayer);
player.points; // => 350 (reactive on updates to that property)
```

tomByrer:
Neat!

Should "cache" be renamed "memo"?  For some reason I believe that `createMemo` is more than a simple cache.

thetarnav:
You might be right again. I was looking from a perspective that I'm *caching* the whole object, by applying *memo* to all the properties. But yeah, the user is looking from the perspective of using the property accessors, so `memo` would be a better name. + it's 4 letters as the rest of the options, so it wins aesthetically :)

thetarnav:
Published at: [`@solid-primitives/destructure`](https://www.npmjs.com/package/@solid-primitives/destructure)

PR #67, opened 2022-03-05
New package: Map

thetarnav:
If there is a [Set](#65) there has to be a Map. The same concepts, the same architecture.
Here is a snippet:
```ts
const userPoints = new ReactiveMap<User, number>();
// reads can be listened to reactively
createEffect(() => {
  userPoints.get(user1); // => T: number | undefined (reactive)
  userPoints.has(user1); // => T: boolean (reactive)
  // ReactiveWeakMap won't have .size or any methods that loop over the values
  userPoints.size; // => T: number (reactive)
});
// apply changes
userPoints.set(user1, 100);
userPoints.delete(user2);
// setter works similarly as signal's
userPoints.set(user1, n => n * 10);

// mutations are NOT tracked:
map.get("John").age = 35;
```
[The readme](https://github.com/davedbase/solid-primitives/tree/map/packages/map#solid-primitivesmap)

### DIFF
--- a/packages/map/LICENSE
+++ b/packages/map/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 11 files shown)


thetarnav:
Published @ `@solid-primitives/map@0.0.100`

PR #69, opened 2022-03-09
debounce improvements

atk:
- types, fixes issue#68
- working uvu test suite
- type tests

### DIFF
--- a/packages/debounce/README.md
+++ b/packages/debounce/README.md
@@ -18,8 +18,9 @@ yarn add @solid-primitives/debounce
 ## How to use it
 
 ```ts
-const [fn, clear] = createDebounce(() => console.log('hi'), 250));
+const fn = createDebounce(() => console.log('hi'), 250));
 fn('my-new-value');
+fn.clear()
 ```
 
 ## Demo
@@ -58,5 +59,9 @@ Updated to Solid 1.3
 1.1.2
 
 Added missing automated clean-up.
+
+1.2.0
+
+Improved types, changed output format from `[fn, clear]` to `fn & { clear }`
 
 </details>
--- a/packages/debounce/src/index.ts
+++ b/packages/debounce/src/index.ts
@@ -1,5 +1,10 @@
 import { onCleanup } from "solid-js";
 
+export interface DebouncedFunction<Args extends any[], F extends (...args: Args) => any> {
+  (this: ThisParameterType<F>, ...args: Args & Parameters<F>): Promise<ReturnType<F>>;
+  cancel: (reason?: any) => void;
+}
+
 /**
  * Creates a method that is debounced and cancellable.
  *
@@ -13,20 +18,18 @@ import { onCleanup } from "solid-js";
  * fn('my-new-value');
  * ```
  */
-const createDebounce = <T extends (...args: unknown[]) => void, A = Parameters<T>>(
-  func: T,
+export function createDebounce<Args extends any[], F extends (...args: Args) => void>(
+  func: F,
   wait?: number
-): [fn: (...args: A extends unknown[] ? A : never) => void, clear: () => void] => {
+): F & { clear: () => void } {
   let timeoutId: ReturnType<typeof setTimeout>;
   const clear = () => clearTimeout(timeoutId);
   onCleanup(clear);
-  return [
-    (...args: A extends unknown[] ? A : never) => {
-      if (timeoutId !== undefined) clear();
-      timeoutId = setTimeout(() => func(...args), wait);
-    },
-    clear
-  ];
-};
+  const debounced = function (this: ThisParameterType<F>, ...args: Parameters<F>) {
+    if (timeoutId !== undefined) clear();
+    timeoutId = setTimeout(() => func.apply(this, args), wait);
+  };
+  return Object.assign(debounced, { clear }) as F & { clear: () => void };
+}
 
 export default createDebounce;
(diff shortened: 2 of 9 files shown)


packages/debounce/src/index.ts
```diff
@@ -13,20 +18,18 @@ import { onCleanup } from "solid-js";
  * fn('my-new-value');
  * ```
  */
-const createDebounce = <T extends (...args: unknown[]) => void, A = Parameters<T>>(
-  func: T,
+export function createDebounce<Args extends any[], F extends (...args: Args) => any>(
```
thetarnav, line 21: I don't think the debounced function should allow for return type, we are not handling the returned type in any way, it will always return undefined.
https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzDigxABEQAjHNMEARngB4AVeEADxNWAGd4AKAHTDYAcx4AueGgDWqHAHdUAbQC6ASngBeAHzwAbjizAANPACC2+AAVYUALYgSMHix07+AKHhIaU5ibe8ApQWBgA-FKoyPYUIDCe6lLKiKhSQiIw4lKWHFy80qhyiiqq8OEW8FEg+vGaugZGpgQQILDp9XqGxqqenqCQsAgo6Nh4BEQk5FQ0IABMTOZZfHkg3HxQqACeamYAYmycawUZgmKSFsud8JtbHkEjYFJ7gT4hYZHw0bHxic-wADJ4ABvFptGAdbRdJrwAC+fQG0DgvlGuHwhDaU0o1HQIAAzItlod8httmp7j5HulhGdljkrlDGsZXsFQhEojE4gkkgJTud6eJrt1gJogaDIODIQ1hXC+gB6OXwLA8HjIEBSAAOdkccBcBE28gw8Di8AwWw1IGAnjAeB4RuUoAodDMEtgZS0E0xZGxszo-H4aW+XOudDmePU1ttRrgPAYHsdftD4fliuVqvVNwgEEUfEQOBg8DgGGQMHwZotZgoyCNGAAFsMaGN8AosFnMyEtnwiyX8GhQIgsKhLZHUHb4I6Fh6McRvTNcXN-YHOXVGUmIzbR9GQDxJ+PKAu1yn4AArZBjmBYUS1o0SddRvcUAlTyaznG0PGLiRBlcNQ8b8-bk+D4fmuQA

thetarnav (changes requested):
Also, the readme should reflect the change to values returned by the primitive.

packages/debounce/src/index.ts
```diff
@@ -13,20 +18,18 @@ import { onCleanup } from "solid-js";
  * fn('my-new-value');
  * ```
  */
-const createDebounce = <T extends (...args: unknown[]) => void, A = Parameters<T>>(
-  func: T,
+export function createDebounce<Args extends any[], F extends (...args: Args) => void>(
+  func: F,
   wait?: number
-): [fn: (...args: A extends unknown[] ? A : never) => void, clear: () => void] => {
+): F & { clear: () => void } {
```
thetarnav, line 24:
```suggestion
export function createDebounce<Args extends any[]>(
  func: (...args: Args) => void,
  wait?: number
): ((...args: Args) => void) & { clear: () => void }
```
-> https://github.com/davedbase/solid-primitives/pull/69#discussion_r822564838

packages/debounce/README.md
```diff
@@ -18,8 +18,9 @@ yarn add @solid-primitives/debounce
 ## How to use it
 
 ```ts
-const [fn, clear] = createDebounce(() => console.log('hi'), 250));
+const fn = createDebounce(() => console.log('hi'), 250));
 fn('my-new-value');
```
thetarnav, line 22: I know that you didn't add the 'my-new-value', but I'm confused by it.

packages/debounce/src/index.ts
```diff
@@ -1,5 +1,10 @@
 import { onCleanup } from "solid-js";
 
+export interface DebouncedFunction<Args extends any[], F extends (...args: Args) => any> {
+  (this: ThisParameterType<F>, ...args: Args & Parameters<F>): Promise<ReturnType<F>>;
+  cancel: (reason?: any) => void;
+}
```
thetarnav, line 6:
```suggestion
export type Debounced<Args extends any[]> = ((...args: Args) => void) & { clear: () => void }
```

thetarnav (commented):
Not too sure about current types, I've given some suggestions, but do with them as you like.
The other thing besides it, would be the failing tests. Those are breaking changes so any dependent packages would need to be updated to the new changes.
Actually the only dependent package is my memo, so I'll take care of it.

packages/debounce/README.md
atk: If you don't define anything in your callback, you'll get `...args: any` - so that's what fn will accept.

thetarnav: not so sure about that one
image
`any` by default in not a good option anyway

atk: This is the case where there is no argument at all, which should be fine, because the callback receives no argument. If you have an untyped argument, eg.
```ts
createDebounce((untyped) => { ... })
```
then untyped will implicitly be of type `any`, as there is nothing for typescript to infer.

On the other hand, if you type your arguments, they should work as typed.

thetarnav:
Your force pushing has removed my updates to the memo package :(

atk:
Sorry.

thetarnav:
@atk What do you think of limiting the function's return type mentioned here: https://github.com/davedbase/solid-primitives/pull/69#discussion_r822564838?

thetarnav:
Alright, I'm gonna merge this. It deserves to be published finally :)

PR #73, opened 2022-03-14
[RAF] fix: it's running the loops twice 

titoBouzout:
It's running the loops twice when calling manually to `start()` without setting runImmediately to `false`

runImmediately is to true by default. This means that the loop would be running. If you then call `start()` like in the example provided

```js
const [running, start, stop] = createRAF(() => console.log('hi')));
start();
```

Then you will have two loops running, as `start()` doesn't check if the RAF is currently running. The animation will run twice as fast.

disclaimer: I only tested this on the playground provided by manually calling `start()` right after createRAF

### DIFF
--- a/packages/raf/src/index.ts
+++ b/packages/raf/src/index.ts
@@ -50,6 +50,7 @@ const createRAF = (
     requestAnimationFrame(loop);
   };
   const start = () => {
+    if (running()) return;
     setRunning(true);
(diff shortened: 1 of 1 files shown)


davedbase:
Good catch, thanks! I'll publish now as well.

davedbase:
Version 1.0.9 includes the patches and is now published.

PR #75, opened 2022-03-18
chore(media): add createBreakpoints primitive

itaditya:
To-Do

- [x] Change to createStore
- [x] Just expose `matches`
- [x] Return fallbackState on server and browsers without `matchMedia`.
- [x] Breakpoint generics
- [x] Option for desktop-first responsiveness
- [x] Update Tests
- [x] Readme
- [x] Add to contributors in readme and package.json


thetarnav:
Hello again, thanks for bringing that primitive here as you said :)
I have a couple of questions/suggestions for the implementation. Not that you need to resolve all of them, just a couple of things I'm wondering.

1. our primitives aim to be usable with SSR setup, so using Browser's APIs – such as `window` object – should be wrapped in a `isClient()` condition. should be an easy fix

2. Also on the topic of SSR, I'm wondering if perhaps one of the `matches` could be configured to equal `true` on the server. You said previously, that if all of the matches are false, then the smallest one is used, which should be fine.
But maybe this could be configurable, or just explained in the readme and left of the user to overwrite.

3. If there is `minMatch`, what about `maxMatch` and `exactMatch`? Would that be possible?

4. The `minMatch` function doesn't seem to be memoized/cashed, meaning that whatever computation is listening to it, will re-execute even if it's value didn't actually changed. You can see it [here](https://playground.solidjs.com/?hash=181044715&version=1.3.7) (added a console.log to your demo on line 90).

5. Allowing number values to be passed to the breakpoints object, which would default to pixels. seams convenient, right?

6. Better type support with generics. It would be nice to infer the keys from the breakpoints objects, and use them in the `minMatch` function for autocompletion/preventing typos of breakpoint names.

7. Types should probably be exported.

8. Readme.

Alright, can't think of more. Let me know what you think about some of them :)

itaditya:
> our primitives aim to be usable with SSR setup, so using Browser's APIs 

> I'm wondering if perhaps one of the matches could be configured to equal true

I am not able to think which token would be the right value to send in SSR. How about we let user pass a `fallbackState` like `createMediaQuery` does and we'll use that in SSR as well as other browser where `window.matchMedia` is not supported.

And because it's all user controlled they'll use minMatch accordingly. Like if they want SSR to be treated as `lg` then they'll do this:

```js
fallbackState = {
  sm: true,
  lg: true,
  xl: false,
}

minMatch('sm') // true if SSR or matchMedia >= sm
minMatch('lg') // true if SSR or matchMedia >= lg
minMatch('xl') // false if SSR or matchMedia >= xl
```

> If there is `minMatch`, what about `maxMatch` and `exactMatch`? Would that be possible?

The reason I only added `minMatch` was because it encourages mobile-first design. The smallest bp is true by default and then you can customise or conditionally render more DOM for larger breakpoints. Like tailwind does with Utility classes

```html
<img class="w-16 text-md md:w-32 md:text-xl lg:w-48" src="...">
```

here the important thing is that on `lg` text-xl & w-48 both apply even though we didn't add `lg:text-xl`

Also I think to support others we then would have to add `max-width` media queries and also ask user for order of breakpoints that would say `2xl > xl > lg`



> 4\. The `minMatch` function doesn't seem to be memoized/cashed

I was treating that function as a checker. Like suppose if there was a function which gave you latest width and in your conditions you would be doing `width > 200`, `width > 400`. Since we can't do `bp > 'lg'`, the API became `minMatch('lg')`

Didn't even know we can memoize when computation is based on caller arguments. [I tried that out now](https://playground.solidjs.com/?version=1.3.7#NobwRAdghgtgpmAXGGUCWEwBowBcCeADgsrgM4Ae2YZA9gK4BOAxiWGjIbY7gAQi9GcCABM4jXgF9eAM0a0YvADo1aAGzQiAtACsyAegDucAEYqA3EogcuPfld69mQqLjgBlNAHNoarA6cXNwBRGRk4Zlx-CEdnOFc4AFk4GFpox1oIAGE1eIh6QnTedwALWkMi90M0XGYSosTXOujpOQVlVQ1tPQsrKww3RhkoVl4AQUZGKHwAGTQAazgAHgAVAD57GMF4kUy1fF5ciC9cEsRefJgTcUstl12Ifd5gCHPL68YAXXOV28k+iAyegQSJoTK8WgmHQRXDBCC4RhoOBkVZrAAUtHOAmAZHOZARGC8314KykvAAPuNJtM5otUQBKc44gnHLAkz7AT6bRxCXBMGIAeShMIAdMICciMfS-gCCMReIkAI5qACSEHxUBByN4AF5eAAlCLcERLfGI1kKuAiNBQACK9HEszQ+LWtysQJBuDBMTiCUatRKACEXPMuAMyGiTCGw-CyPTuU5MvjeDBlWqNVqyLrAvE3MlUmi0fGdRsQAFYkm+KnVercJrWLiFWna-XtXqQJI3VsMsLInCJRGo-FQ7Rw-SRTJuMERiVC8BcLRFhA2dURKdPsXS+XHInaymrTbFQ7GAc9QADNEwDBaVenc4AEhAt5KknpZ9uO53zErvAwGdGerVKI5QiqgAbJNaUCXgeUBHo60oAp+jjVumdaZvOi7CFyep-mhrAfjur5dp+vL8imzb-siBFEYhsjAqC4JeHAuBqjUNpqP6dSSvGZbdruyZgVxWbtp2tE9tCfbwoikooS2mZFuOk6MNOdRzguS5stWG66lufHIU0JTIhhS7YeRaigQZVHbjREDbqRjAxIJhlkNRtHfnuwBOcibJkMxnHOaZvpuJ4PhQGoaJMSx1hemF-ncQhtlbB6DGORg-loupwg8XZzFkUFSQpLQhabgmn7ucmzr+YS2ZeRG9LGVhBEkblDm8IGtDqHkaKVQZhIJZ+NmOP8iWxEEcChOEkTFTppUQr2sJSUiEayZRdWKVOM5qZhy5mdpJazRWe4aPiwjiNmaJwAAbuKJW8UhO6+bgcURmihBCJdt3bvd9kxHd91ISKgNvVdRT-TuDUQMSV3ihZAbeV9SGiXphH9YjTU7tWIpQCIIjBNd8JzCdEDiGiKh1JqTEqGyx1uMTjAJQjmQ5HkBTTftf33ZjQipNdePioTtMk2TJQU3AVOHM6gv0+jQ2o7L1EM3cLW-duV4QP5oMppZZBFEjw1WOVfBDlAI7htmf1kDA5wqAAbAALAADIQFBUwEMAiNbYAAOw2wAHM7rtbGoXiewAjA7ABMdsB9gAQUGoYcR77Tsu7HEDSFAWaG12yVeuCADqmiRUWCaG-wKZpQZZJ6vl-nBsOMbkJG0ajrGiujbm41hDCbO6WVSadSKai0F4pNgMHiDi2r6UqMHKj0gpCvEdsfKtaTfFLNal1rAjSwlAAzGsKwntsZBoAAXtVQG7IYSz6AfO-I0sVQ1HUshhWoJgjPMOogJvaDb3cKgNQuR8R3y3msSQj8wZLH8rwQwhkIC-2ngZMe8d55FigcECgCIoC8BmFARgTE77+Wgf9WBVcEHCGQZXAMY855gAXvSKBBCiHLH0KQhGjgKEBngYgmh6tUEqHdhg5hawIJoHoDAEhBkyH3R4W-KhSCQAoLoSoS2oioFAI-jIgMcidx3xfno7c4CAFkISvrQE9E84xDGIQQgJc-plwwl4LwuQfLMRWCPNxcBApjRCr4DKjAHTtxXmRden5-7b13iYeguAFwxDlHAHUKhYnxMyCoCE2QNDMB-iAEu+1HpeNcbkQs6gRAlQAITlLEVwkk3jci730GkhJ+juGlHKHw6hIAFwlLgJgtpBjC4iEirwfQgzDFlEMPo0x0StgWIBEIUQJMCkbCWHYwgYy1hsl2MwKRMNIrBFyPAeEgZ8AqhEGPKA9j54JTAJIT4QA) but because it becomes wrapped in accessor having multiple conditions looks clunky

Is there a better way to do this?

> Allowing number values to be passed to the breakpoints object

I personally wouldn't prefer to default numbers to `px`. Some CSS properties support unit-less like lineHeight so in future if `solid-primitives` have any other thing related to CSS units it might cause confusion. Also this creator would probably be wrapped into a consumer's dedicated creator where they'd be pulling out the breakpoints from some static config or theme context so they won't have to write breakpoints object by hand. If they don't have such mapping then they'll need to do it once. 

Because of this I think letting user pass the entire css value should be good enough.


> Better type support with generics. It would be nice to infer the keys from the breakpoints objects, and use them in the minMatch function for autocompletion/preventing typos of breakpoint names.

I tried doing it but gave up because I'm new to TS and have no clue how to debug generics 🙈. Could you help me with this? You already have the Solid Playground link.

> Types should probably be exported.

Done

> Readme

Let's do it once the API is finalized 👍🏽 


Also I added Vitest based tests but the second last test which depends on event trigger to re-run `createEffect` it was working initially then suddenly stopped working. I'm new to Solid so I'm not sure what I'm doing wrong. cc @davedbase 

In React, I would have made a custom test component, used hook in it and rendered to DOM so that I could just assert the DOM without trying to capture output arrays. However `solid-testing-library` is not working with Vitest. I tried @atk setup [from this blog](https://dev.to/lexlohr/testing-your-solidjs-code-in-vitest-6dh)

davedbase:
Great job here! Lots of really valid points @thetarnav and @itaditya thanks for bringing this in. Can you please remember to add your name to a contributors list on the README and also in the "contributors" key in package.json? :-)

thetarnav:
@davedbase What do you think about moving the `createMediaQuery` primitive to a named export?
It could be still exported as both named and default for backward compatibility.

davedbase:
Fantastic work everyone.

PR #77, opened 2022-03-23
New package: URL

thetarnav:
This package focuses on three interfaces: Browser's `window.location`, `URL` and `URLSearchParams`. Providing a couple reactive primitives, reactive classes and additional utilities.

A detailed overview can be found in the [README](https://github.com/davedbase/solid-primitives/tree/url/packages/url#solid-primitivesurl)

The package has many possible improvements, [(some noted in the readme)](https://github.com/davedbase/solid-primitives/tree/url/packages/url#improvement-ideas) but I've decided to leave them for later. So that I'm able to use these primitives first and get some feedback, before extending them further.

A question:
Should this be kept as one package or split into separate `location`, `url` and `url-search-params` smaller packages?
I started it as a single package, and as all the primitives are related to each-other more or less, I've kept it as is. But it is possible to split them. *(already are kept in separate files anyway)* They would stay as dependencies when needed.


atk:
It would help if a fallback could also be given as string (we can easily detect this case and wrap in URL()).

davedbase:
I think having these under a URL package is fair. They are all generally related to url and packaging them as some keeps them contained. I think this is safe to merge unless you have further changes.

davedbase:
How do we feel about merging this in? :-)

packages/destructure/dev/index.tsx
```diff
@@ -43,7 +43,7 @@ const App: Component = () => {
       <div class="wrapper-v" onclick={() => setCount(p => ++p)}>
         <button class="btn">{count()}</button>
       </div>
-      <Inner seconds={seconds()} count={count()} list={list()} />
+      {/* <Inner seconds={seconds()} count={count()} list={list()} /> */}
```
atk, line 46: Why that change?

thetarnav: I honesty don't remember, for the destructure I was treating `dev/` as a playground than a demo of how things work.
Not that it should have been committed either way...

thetarnav:
@davedbase I was deferring the merge because I wasn't sure about the current API, it's "finished" but still requires a lot of further work on refactors and more features. Work that I lost interest in, for now ;)
I will try to merge and publish in soon. Maybe some minor adjustments. URL is a rabbit hole...

PR #78, opened 2022-03-24
Fixes #74 ([RAF] allow limiting fps when > 60)

### LINKED ISSUE #74: [RAF] allow limiting fps when > 60
https://github.com/davedbase/solid-primitives/tree/43d9abed17cd61cda4238957db82eabafe0ea5e9/packages/raf

Currently, the RAF package provides a way to run animations slower than 60 fps. When animations are faster than 60fps then it just uses the screen refresh rate and doesn't limit the execution. See https://github.com/davedbase/solid-primitives/blob/43d9abed17cd61cda4238957db82eabafe0ea5e9/packages/raf/src/index.ts#L38-L42

It would be nice to allow limit the fps also when faster than 60fps because allows saving resources like cpu, battery, etc. For some animation/computations, a lower fps could be considered enough by the developers. In an example, your screen refresh rate is 300, but you consider that 60 or 120 is more than enough for the computations at hand.

I would like to add this some day, so is more like a note to self than a request, but feel free to take over if you feel like.

titoBouzout:
Related to #74

The RAF package https://github.com/davedbase/solid-primitives/tree/main/packages/raf 

Provides a way to run animations slower than 60 fps. When animations are faster or equal to 60fps then it just uses the screen refresh rate and doesn't limit the execution (which could be high above 60  [300 in my case]). See

https://github.com/davedbase/solid-primitives/blob/43d9abed17cd61cda4238957db82eabafe0ea5e9/packages/raf/src/index.ts#L38-L42

After studying this code, I came to the realization there's a reason for the above. It's hard to target a frame rate, and it's harder without consuming resources(this is very important, we will know below).

The problem is the following: 

When you target a frame rate, you expect something to run after some `ms`. If your `targetFrameRate` is `60`, then you expect the next run to be in around or above `16ms`.  This is a problem because if you get `15ms` (values below 16sm are only possible because of a high refresh rate screen), then that won't satisfy you, and you would wait for the next frame, which will be: the delay already applied :`15ms` + your screen refresh rate. In other words, you didn't accept `15ms` as the time to run something, so you should wait for the next frame, which means that you should wait as long as your screen refresh time is. 

To illustrate it, 
chrome_b1hHQWkBHM 
Most of the time this value is near zero(if the callback does little to not work at all), We should always keep in mind that the worse case of this value would be your target frame ms -1 + your frame rate. Which explains why do you see on my screenshot values near 300 / 1000 and around `19ms`. Check it yourself looking at the console after pressing start/stop https://codesandbox.io/s/solid-create-raf-forked-h2bk62?file=/src/index.tsx

This is not a problem you can solve, as you can't change the past(if you were late you late!), but you can change the future, for the sake of "targeting a frame rate" See 
```js
    elapsed = timeStamp - lastRun;
    if (elapsed + missedBy >= interval) {
      lastRun = timeStamp;
      missedBy = Math.max(elapsed - interval, 0)
      callback(timeStamp);
    }
```

The `missedBy` accounts for how much time was wasted on the last render, and consider it on the next frame to recover from that to give the best of the experiences.

Here is the tricky part, there are two kinds of users for a `targetFps`:
- you can target a frame rate consistently as desired by consuming all or most of my CPU because that's your requirement (by counting how many frames you painted, etc)
- you can target a frame rate to do exactly the opposite, to clam down the CPU (theres no reason to run something so many times if you consider that less is better)

I am satisfying the latter here, and the former maybe can be satisfied by another package(which Im not even sure its worthy because they will most likely do it themselves)

Un/Related considerations of this pull request:

- The signature of requestAnimationFrame doesn't match what's provided by this, so I made them match. this is no longer providing a delta, it provides what requestAnimationFrame  gives, and if users want the delta we can do it on a future update as a secondary argument
- It changes the `targetFps` default value from `60` to `Infinity`, that way we can target values above 60
- it does some rename from `fps` to `targetFps`, as `fps` doesn't really mirror the current `fps`, it's just a target. I figure we can tell "at how many `fps` we are running and pass it down" as we are already running a RAF non-stop.

Disclaimer, I have no idea how to run this TypeScript, so I only tested by copying pasting on codesandbox, so if you can takes a test 🙏 would be welcome. I would like to try to setup myself to deal with TS but well.

### DIFF
--- a/packages/raf/src/index.ts
+++ b/packages/raf/src/index.ts
@@ -1,53 +1,61 @@
-import { createSignal, onMount, onCleanup, createMemo } from "solid-js";
+import { createSignal, onCleanup, createMemo, createEffect } from "solid-js";
 
 export type FPS = number | Function;
-const getFps = (fps: FPS) => (typeof fps === "function" ? fps() : fps);
-const calcFpsInterval = (fps: number) => 1000 / fps;
+const getFps = (targetFps: FPS) => (typeof targetFps === "function" ? targetFps() : targetFps);
+const calcFpsInterval = (targetFps: number) => 1000 / targetFps;
 
 /**
  * Creates a method to support RAF.
  * Based on useRafFN (https://github.com/microcipcip/vue-use-kit/blob/master/src/functions/useRafFn/useRafFn.ts).
  *
  * @param callback The callback to run on a frame
- * @param fps Target frame rate, defaults to 60
- * @param runImmediately Determines if the function could run immediately
- * @returns Returns a signal if currently running as well as star tand stop methods
+ * @param targetFps Target frame rate, defaults to Infinity
+ * @returns Returns a signal if currently running as well as start and stop methods
  *
  * @example
  */
 const createRAF = (
-  callback: (timeElapse: number) => void,
-  fps: FPS = 60,
-  runImmediately = true
+  callback: (timeStamp: number) => void,
+  targetFps: FPS = Infinity
 ): [running: () => boolean, start: () => void, stop: () => void] => {
+
   const [running, setRunning] = createSignal(false);
-  const fpsInterval = createMemo(() => calcFpsInterval(getFps(fps)));
-  let wasIdle = false;
-  let startTime = 0;
-  let timeElapsed = 0;
-  let timePaused = 0;
-  let timeDelta = 0;
+  const fpsInterval = createMemo(() => calcFpsInterval(getFps(targetFps)));
+
+  let elapsed = 0;
+  let lastRun = 0;
+  let missedBy = 0;
+
+  let interval = fpsInterval();
+  let fps = getFps(targetFps)
+  createEffect(() => {
+    interval = fpsInterval();
+    fps = getFps(targetFps)
+  });
+
+  let isRunning = running();
+  createEffect(() => {
+    isRunning = running();
+    lastRun = 0;
+    missedBy = 0;
+  });
+
   const loop = (timeStamp: number) => {
-    if (!startTime) startTime = timeStamp;
-    if (!running()) return;
-    if (wasIdle) {
-      timePaused = timeStamp - startTime - timeElapsed;
-      wasIdle = false;
-    }
-    timeElapsed = timeStamp - startTime - timePaused;
-    const shouldRun = getFps(fps) >= 60;
-    if (shouldRun) {
-      timeDelta = timeElapsed;
-      callback(timeElapsed);
-    }
-    if (!shouldRun) {
-      const elapsedTimeFromPrevLoop = Math.ceil(timeElapsed - timeDelta);
-      if (elapsedTimeFromPrevLoop > fpsInterval()) {
-        timeDelta = timeElapsed;
-        callback(timeElapsed);
-      }
-    }
+    if (!isRunning) return
+
     requestAnimationFrame(loop);
+
+    if(fps === Infinity) {
+      callback(timeStamp);
+      return
+    }
+
+    elapsed = timeStamp - lastRun;
+    if (elapsed + missedBy >= interval) {
+      lastRun = timeStamp;
+      missedBy = Math.max(elapsed - interval, 0)
+      callback(timeStamp);
+    }
   };
   const start = () => {
     if (running()) return;
@@ -56,9 +64,8 @@ const createRAF = (
   };
   const stop = () => {
     setRunning(false);
-    wasIdle = true;
   };
-  onMount(() => runImmediately && start());
+
   onCleanup(stop);
   return [running, start, stop];
 };
--- a/packages/raf/README.md
+++ b/packages/raf/README.md
@@ -19,14 +19,14 @@ yarn add @solid-primitives/raf
 
 ```ts
 const [running, start, stop] = createRAF(
-  (timeElapsed) => console.log('Time elapsed is', timeElapsed))
+  (timeStamp) => console.log('Time stamp is', timeStamp))
 );
 start();
 ```
 
 ## Warning
 
-To respect clients refresh rate, timeElapsed should be used to calculate how much the animation should progress in a frame, otherwise the animation will run faster on high refresh rate screens. As an example: A screen refreshing at 300fps will run the animations 5x faster than a screen with 60fps if you use other forms of time keeping. See https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame
+To respect clients refresh rate, timeStamp should be used to calculate how much the animation should progress in a given frame, otherwise the animation will run faster on high refresh rate screens. As an example: A screen refreshing at 300fps will run the animations 5x faster than a screen with 60fps if you use other forms of time keeping that don't consider this. Please see https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame
 
 ## Demo
 
@@ -52,5 +52,14 @@ Updated to Solid 1.3, switched to peerDependencies
 1.0.9
 
 Patched double running and added refresh rate warning (patch by [titoBouzout](https://www.github.com/titoBouzout)).
+
+2.0.9
+
+- allow to limit fps above 60fps
+- default `targetFps` to `Infinity` instead of 60fps
+- remove `runImmediately` as trying to change the default of this value will require you to provide a `targetFps` which may not be clear that should be `Infinity` in case you want to respect clients fps. Screens above 60fps are becoming increasing popular.
+- respect `requestAnimationFrame` signature and give `timeStamp` back to the callback instead of a `deltaTime` (deltaTime could be added back on a future version)
+
+(patch by [titoBouzout](https://www.github.com/titoBouzout)).
 
 </details>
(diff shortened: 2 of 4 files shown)


packages/raf/src/index.ts
```diff
@@ -1,53 +1,61 @@
...
-  let timeDelta = 0;
+  const fpsInterval = createMemo(() => calcFpsInterval(getFps(targetFps)));
+
+  let elapsed = 0;
+  let lastRun = 0;
+  let missedBy = 0;
+
+  let interval = fpsInterval();
+  let fps = getFps(targetFps)
+  createEffect(() => {
```
thetarnav, line 31:
```suggestion
  createComputed(() => {
```
If this effect (and the one below) are for setting variables, shouldn't this be a pure computation? Just so the changes happen as soon as possible.
Probably doesn't matter much. just nitpicking while reeding through the code

thetarnav:
Don't know if I'm doing the math right, (it's late) but it seams that the new implementation is more likely to skip frames, making it ultimately run in lower FPS than the declared value.
I've made a demo here: https://codesandbox.io/s/tito-raf-vs-old-raf-rd9hwp?file=/src/index.tsx

I like the idea of correcting for the "time lost" in the next run, but if that means that the animation will run at different speed overall, then maybe it's not worth it. Or just the implementation needs some correcting. Or it's just late and I don't know what I'm talking about...
But maybe it is fine as an alternative primitive, or some option to switch between "steady animation mode" and "don't overuse the CPU mode". With "steady animation" being the default I think, idk.

I like the idea of removing initial run. Makes it more explicit when you want to start the animation I think.

packages/raf/src/index.ts
titoBouzout: Thanks for your comments, appreciate the suggestion, very welcome and I would love to hear about the differences between the two, because I don't know this is just confusing. When would you use one over the other? Sorry do not rush, thanks

thetarnav: From my understanding, right now there are two differences: an effect is batching and runs after mount, while computed is not. There is probably more going on internally, but from the usage standpoint that's it.
You should utilise effects as much as possible for most cases, because batching helps to reduce unnecessary side-effects (like rendering). And will not run on the server for the same reason.
Computed is used for creating own primitives, for setting variables, where you don't have side effects, only outputs as signals.
In this example you're using createEffect – an impure computation – to have a derived value of sorts, so something you'd use createMemo normally – a pure computation – but if you can't, you can use createComputed – also a pure computation.

trusktr: I'd vote to just use `createEffect` if it works, as that will be more inline with potential ways the API could change later. Ryan did mention in Discord that he thought `createComputed` might be a mistake. Having both introduces room for timing differences, that can lead to unexpected results.

I described in https://github.com/solidjs/solid/issues/879 what I think would be the ideal behavior. In that ideal, 

- deferred/batched effects would be a default (not to confuse the traditional JavaScript notion of batching with Solid's concept of `batch`, which is something different),
- `createComputed` and `batch` would both be removed because the new effect behavior would eliminate the need for them,
- all signal reads would _always_ return their current value and never ever a value in the past (confusing),
- and memos would be deferred/batched too (due to being built on effects) except they'd also be lazy in case a current value is needed before a future effect runs.

In this way, we'd eliminate unnecessary excessive effect runs in a consistent way, and we'd also eliminate glitches.

---

The [Qt C++ framework](https://qt.io) (well known, used in many commercial settings, Linux desktops, and cross-platform apps), made the change I described to their reactive system: in a breaking change, all reactions to reactive states were moved to deferred/batched to improve performance and simplify code execution. I am looking for the article they wrote on it.

trusktr:
In the 3D animated products I've made for work projects, I haven't had to specify a target FPS. [LUME](http://lume.io/) doesn't have that option yet, and a vast amount of Three.js examples also don't have this concept (I haven't seen if any do).

I think that rAF should run at whatever speed the browser deems it can. The browser already balances workload between tasks and frames.

I do think that a higher-level animation API specialized in timing can be made, and _that_ can have similar options, but for the base `rAF` API, I don't think it is needed.

If we remove the FPS stuff, the implementation will get simpler, and is what most use cases need anyway.

trusktr:
@titoBouzout Do you have a particular use case for target FPS?

packages/raf/src/index.ts
thetarnav: So the only difference between computed/memo and effect will be that the later don't run on the server?
In this case it doesn't really matter, but in most cases, computed/memo should still be the winner for derived values.

titoBouzout:
> what kind of problem do you have with "TypeScript", do you want to have a dev environment with vite locally,

I didn't set up my dev environment to run this TS code yet (Im not a TS user). I do not like the idea of sending pull requests of code I'm testing on the playground. I will try to figure it out today.

> I've made a demo here: https://codesandbox.io/s/tito-raf-vs-old-raf-rd9hwp?file=/src/index.tsx
The target is 55.  I get 50 with the old version and 55 with my version. 

> Wonder if there is a way to keep it as close to desired average frame-rate as possible..

I wonder the same, if I am missing something, because it's always possible to target a higher frame. In an example, if I target 70 with your codesandbox, I get 67. If I target 90 I get 86, so in theory 70 should have been possible. 

My only explanation is that your worse frame drop will be a bit below your current screen refresh rate, which happens every some frames, the higher your target, the more possible to fall into this situation is. I have seen code around targeting a frame rate doing something similar and having a `tolerance` value, which also means your frame rate will fluctuate, so I avoided this. 

Feels like there's a need to adjust the value of the calculated interval with the screen refresh rate. Something like if my refresh rate is 3ms and I'm targeting 16ms then I would use 15ms(5 frames x 3ms), as it's closer to 16ms than 18ms(6 frames x 3s). Could be that? You will still be off every once in a while but should be closer. I will have to think harder, I'm bad at math. Thougths?

> I do think that a higher-level animation API specialized in timing can be made, and that can have similar options, but for the base rAF API, I don't think it is needed.

The thing is that you can use this code for other stuff than animating. In any case, is an optional argument to cap how many frames you want. I only got interested because of the frame limitation, else I don't think I would have a use for it. Lots of games provide the ability of setting a target fps. Its something nice to have.

> Do you have a particular use case for target FPS?

Besides keeping the fans silent xD I use it yes. Like collecting user input (mouse moves) and then send it via a socket every N frames, which allows making the message shorter, as you don't need in the future to give ids for these inputs.

titoBouzout:
Just updated flooring the interval and ceiling the elapsed to favour giving more rather than less.  Results seem to be more consistent.  Also updated the demo https://codesandbox.io/s/solid-create-raf-forked-h2bk62?file=/src/index.tsx

thetarnav:
I've added a /dev playground that you can use, and updated the types to use utils.
You'd need to resolve the conflicts if you want to work on it more, but I haven't touch the logic, so it should be quick.

titoBouzout:
@thetarnav Thanks much appreciated. I got vite running and improved the tests. Made some changes:

- changed the secondary argument to be an `options` object. That way, if this package receives optional features, then developers won't need to set a `targetFPS` to be able to use the other arguments. As most likely and demonstrated before, they will likely inadvertently set this to `60`
- it uses `cancelAnimationFrame` now instead of constantly checking  for `!isRunning` as it used to do.
- I had to add back the `getFPS` because as there's an options object now, we need to default to `Infinity`, not sure if this could be handled in TypeScript some other way to default to a value
- can you please check if this is correct, it reads weird and I don't use TypeScript: `callback: Get<number>,`

Possibly future updates:
- via the option `delta` pass down a delta value
- via the option `FPS` pass down at how many fps we're running

I'm somewhat happy with the state of it. I got what I wanted without complicating it, I think. 

Using `floor` and `ceil` is helping to target lower frame rates, but it makes difficult to target higher frame rates, as these two are dropping significant bits that should be considered. I was thinking a possible solution would be to find a sweet spot and do something like, `let ceil = fps < 100 ? Math.ceil : (n) => n `, but then again, I don't want to complicate this thing.

thetarnav:
Okay wow, now it actually works and it's so satisfying to see the average FPS value slowly getting closer and closer to the desired limit. Good Job!
There is still optimisation potential of removing unneeded computations when the `targetFPS` is not defined or is a constant value. But I can take care of that later.
The PR should be ready to me merged I think.

titoBouzout:
@thetarnav IHold on, just noticed this package is used around. I will update and also rename it to `createRequestAnimationFrame` as is goes more in hand with the naming conventions around.

thetarnav:
I think that changing the name of the primitive would be outside of the scope of this PR.
It is a default export anyway so renaming it can be done later, at a different PR/Issue and with a fresh discussion.

titoBouzout:
OK Thanks, changed an unrelated dev test case to start after calling createRAF given that we don't auto start anymore

thetarnav:
Alright, time to merge and work on publishing this.
Thank you @titoBouzout for the improvement and bringing it here.

We've also decided to take this opportunity of a major release to rewrite the API a bit. The `targetFPS` "feature" will be turned into a separate function for wrapping the callback. This will keep the base primitive as small as possible.
```ts
const [running, start, stop] = createRAF(
  targetFPS(() => {...}, 60)
);
```

titoBouzout:
@thetarnav sounds good, thanks for your help!

PR #79, opened 2022-03-25
input-mask primitive

atk:
Disclaimer: this is actually not even indirectly tied to Solid.js (it works happily with composed and native events and has no dependencies whatsoever). Still, I think it's worth keeping, as this is a regular requirement.


packages/input-mask/README.md
```diff
@@ -0,0 +1,82 @@
...
+- **Why `oninput`/`onchange` instead of e.g. `onkeyup`?**<br> There are a few things happening after keydown, -press and -up, which would result in flickering. Generally, you could use `onblur`, but then you'd attempt to apply the mask even if nothing changed, which just seems unnecessarily wasteful.
+- **Will it work with actual events?**<br> Yes, it will work with composed as well as native events. Solid's composed events are more performant, though.
+- **Is there a server version?**<br> No, since it only creates an event handler that will solely run on the client; it makes no sense to create a server version.
+- **Does this provide any error handling?**<br> There is no error handling, but it should work well together with any form handling library.
+- **Can I turn this off and on again?**<br> You can still wrap the output handler in your own handler to turn it off and on again:
+```jsx
+const [useMask, setUseMask] = createSignal(false);
+const mask = createInputMaskHandler('9999-99-99');
+const onInput = (ev) => useMask() && mask(ev);
+return <input oninput={onInput} />;
```
thetarnav, line 66: Is it necessary use signal here? It might be easier to understand if it were just a `let`.
```suggestion
let useMask = false
const mask = createInputMaskHandler('9999-99-99');
return <input onInput={e => useMask && mask(e)} />;
```

packages/input-mask/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+return <>
+  <label for="changeMask">
+    The mask will only be applied after you leave the field
+  </label>
+  <input type="text" id="changeMask" onchange={createInputMask('99/99/9999')} />
+  <label for="inputMask">
+    The mask will be applied on every single input
+  </label>
+  <input type="text" id="inputMask" oninput={createInputMask('99/99/9999')} />
```
thetarnav, line 48: How would you actually get a value out of the input? I think it's worth to include an example here, where you save the masked input value to some signal.

packages/input-mask/README.md
```diff
@@ -0,0 +1,82 @@
...
+npm install @solid-primitives/input-mask
+# or
+yarn add @solid-primitives/input-mask
+```
+## Usage
+
+The masks come in 3 different formats: function, array and string. There are tools to convert string masks to array masks and array masks to function masks.
+
+```ts
+// string mask:
```
thetarnav, line 22: Including imports is always nice to know what's coming from the package, and how to import it.
```suggestion
import { createInputMask } from "@solid-primitives/input-mask";
import { anyMaskToFn} from "@solid-primitives/input-mask";

// string mask:
```

packages/input-mask/README.md
```diff
@@ -0,0 +1,82 @@
...
+- **Why `oninput`/`onchange` instead of e.g. `onkeyup`?**<br> There are a few things happening after keydown, -press and -up, which would result in flickering. Generally, you could use `onblur`, but then you'd attempt to apply the mask even if nothing changed, which just seems unnecessarily wasteful.
+- **Will it work with actual events?**<br> Yes, it will work with composed as well as native events. Solid's composed events are more performant, though.
+- **Is there a server version?**<br> No, since it only creates an event handler that will solely run on the client; it makes no sense to create a server version.
+- **Does this provide any error handling?**<br> There is no error handling, but it should work well together with any form handling library.
+- **Can I turn this off and on again?**<br> You can still wrap the output handler in your own handler to turn it off and on again:
+```jsx
+const [useMask, setUseMask] = createSignal(false);
+const mask = createInputMaskHandler('9999-99-99');
```
thetarnav, line 64: It doesn't seam that the `createInputMaskHandler` is in the package, is it still a todo?

thetarnav:
Shouldn't the `createInputMask` primitive be renamed to something like `inputMask` or `maskInput`. It is not declarative, but imperative, nothing gets created, just "handled" :)

packages/input-mask/README.md
atk: Good catch. I considered a longer name to convey that this returns a handler, but then I thought, documentation is better than superfluously long names.

packages/input-mask/README.md
atk: Then I should probably include all of them.

packages/input-mask/README.md
atk: Good idea. I added two ways. I'm still considering if I should make the handler return the value for convenience reasons. It wouldn't exactly fit the signature of an event handler anymore if typed strictly to emit void, but this would allow to use the return value.

packages/input-mask/README.md
atk: I first thought of a prop. A signal at least conveys that there coudl be change. But you're right, we don't need that.

packages/input-mask/README.md
thetarnav: So the approach with "handler returns the value" would look like this?
```tsx
const [date, setDate] = createSignal('01/01/2020');
const mask = createInputMask('99/99/9999');
return <input onInput={e => setDate(mask(e))} />
```
This seams pretty clean, but how would the other approach look like?

thetarnav: Oh, and I guess there is a problem with the initial value, you cannot use signal for that...
```tsx
const initial = "01/01/2020";
const [date, setDate] = createSignal(initial);
const mask = createInputMask("99/99/9999");
return <input value={initial} onInput={e => setDate(mask(e))} />
```

thetarnav: Also how setting the value from outside would look like?
For example if you have a date picker with masked text input and calendar component connected to each other.
How would you apply value to the input, when user changes a day using the calendar?

atk: Just set the value of the input. Overwriting the value from JS is not going to trigger the handler.

atk: You also could still use it to write the let variable:

```ts
let date = "01/01/2020";
const mask = createInputMask("99/99/9999");
return <input value={date} onInput={e => { date = mask(e) }} />
```

atk: Yes, you've convinced me to add this.

thetarnav: So something like this? If this works, then that's great.
```ts
const [date, setDate] = createSignal("01/01/2020");
const handler = createInputMask("99/99/9999");
return <input value={date()} onInput={e => setDate(handler(e))} />

// later, a change from outside
setDate("12/09/2021")
```

thetarnav (approved):

PR #82, opened 2022-03-28
New Control Flow package: Range – closes #80

### LINKED ISSUE #80: package proposition: Range
If we decided to do a API preview/discussion before the implementation, then here we go :)

I'm proposing adding a new Control Flow package, that would be adding a nicer and more optimised way of repeating a JSX element n-times. For example when wanting to display:
- some kind of count with individual elements – _number of stars by rendering each one as a star icon? 😅_
- numbers from x to y
- ?

#### Optimisation opportunities: (compared to using `<Index>` for it)
- no array of empty items need to be created
- there is no array, so there are no array items, so no signals for those items need to be created and kept in memory
- cleaner interface

#### The `mapRange` primitive – powering `<Range>` component:
Usage:
```ts
const [length, setLength] = createSignal(10)
const mapped = mapRange(length, index => {
   const [value, setValue] = createSignal(index);
   createEffect(() => {...})
   return value
})
```
Definition:
```ts
function mapRange<T>(
  length: Accessor<number>,
  mapFn: (i: number) => T,
  options?: { fallback?: Accessor<T> }
): Accessor<T[]>;
```

#### The `<Range>` component:
Usage:
```tsx
<Range of={10}>
   <div></div>
</Range>

with a render prop:
<Range of={10}>
   {n => <div>{n}</div>}
</Range>

with fallback:
<Range of={0} fallback={<p>no items...</p>}>
   <div></div>
</Range>
```
Comparing to how it currently can be achieved:
```tsx
<Index each={new Array(10)}>
   {(_, n) => <div>{n}</div>}
</Index>
```
Definition:
```ts
function Range<T>(props: {
  of: number;
  fallback?: T;
  children: (index: number) => T;
}): Accessor<T[]>;
```

#### More props?

I think that a single `of` prop is enough for most usage cases, and any range can still be achieved:
```tsx
{/* from 5 to 15 */}
<Range of={10}>
   {n => <div>{n + 5}</div>}
</Range>
```
But a few additional props could possibly be helpful:
- `from` - start of the range (0 by default)
- `to` - end of the range *(an alternative to `of` as it also dictates the range size)*
- `by` - increment numbers by n (1 by default)

The downside of adding those could be:
- increased complexity of the component
- wasteful re-rendering of all of the elements in some cases *(e.g. changing the increment_by value)*

Thoughts? :)

thetarnav:
This brings new Control Flow package discussed here: #80 

[Package README](https://github.com/solidjs-community/solid-primitives/tree/range/packages/range#solid-primitivesrange)

Primitives TODO:
- [x] `repeat`
- [x] `<Repeat>`
- [x] `mapRange`
- [x] `<Range>`
- [x] `indexRange`
- [x] `<IndexRange>`

I consider the initial implementation to be finished, although it still have some rough edges to be improved with time *(need a break from numbersss)* listed [here](https://github.com/solidjs-community/solid-primitives/tree/range/packages/range#possible-improvements).


thetarnav:
Published at [`@solid-primitives/range`](https://www.npmjs.com/package/@solid-primitives/range)

PR #84, opened 2022-03-30
handle/document onPaste event

atk:
+ more examples


atk:
Well, it's an actual use case that I built this for, so why not include it?

thetarnav:
Of course you should. Just found that funny :)

PR #86, opened 2022-04-01
WIP new primitive: createFileUploader

frixaco:
Thank you for an amazing repo!
It's my first ever open source PR, so any kind of feedback is highly appreciated.

Snippet:
```ts
// createFileUploader
<button onClick={() => {
    selectFiles(([{ source, name, size, file }]) => console.log({ source, name, size, file })});
}}>

// multiple files
const {files, selectFiles} = createFileUploader({ multiple: true, accept: "image/*" });
selectFiles(files => files.forEach(file => console.log(file)));

// single file
const {file, selectFile} = createFileUploader();
selectFiles(([{ source, name, size, file }]) => console.log({ source, name, size, file }));

// createDropzone
<div ref={dropzoneRef} style={{ width: "100px", height: "100px", background: "red" }}>Dropzone</div>

const { setRef: dropzoneRef1, files: droppedFiles1 } = createDropzone({
  onDrop: async files => {
    await doStuff(2);
    files.forEach(f => console.log(f));
  },
  onDragStart: files => console.log("drag start")
  onDragStart: files => files.forEach(f => console.log(f)),
  onDragOver: files => console.log("drag over")
});

// fileUploader directive
const [files, setFiles] = createSignal<UploadFile[]>([]);

<input type="file" multiple use:fileUploader={{
  userCallback: fs => fs.forEach(f => console.log(f)),
  setFiles
}} />
```

Demo: https://codesandbox.io/s/solid-primitives-upload-ry042x?file=/src/index.tsx
Reference: https://github.com/Marvinified/use-file-upload

TODO:
- [x] Port the React hook
- [x] Add/expose additional helpers
- [x] Add async callback support
- [x] Improve types
- [x] Add createDropzone
- [x] Add fileUploader directive


davedbase:
Awesome this was much needed. The next step is a createDropzone utility =)

Two quick comments, can you include a noop server entry? A lot of our primitive have that part of our standard pattern. It just mitigates the server doing anything it doesn't need to.

Also I'm not totally sold on the input being created from the component. I rather let the user create the component and use a directive to bind the uploader to it.

thetarnav:
Welcome to the amazing world of open source, and thank you for bringing this here! :)
So far, you're nailing it for me, I don't have any suggestions for the work you've put in so far.

But I'm wondering a couple of things about the API: *(I know they you're just poring the use-file-upload, but we can always improve if there is a reason to)*

1. Wouldn't it be better to declare upload config on the `createUpload` primitive itself instead of on the `selectFiles` callback?

   This would allow to narrow the types of the callback and `files` accessor. I can see a lot of type casting in your demo.
The type of file you want to accept and it's count is usually known up-front. If you have a pfp component you only accept a single image. If you have an attachments component in your email, it accepts any kind of files and any number.
This would even make the the api closer to the `<input>` component where you declare those sort of restrictions up front.
```ts
const [file, selectFile] = createUpload({ accept: "image/*" });
file() // T: UploadFile
selectFile({ source, name, size, file }) => {...});

const [files, selectFiles] = createUpload({ multiple: true });
files() // T: UploadFile[]
selectFiles(files) => files.forEach(file => {...}));
```

2. What about native `<input type="file">` elements?

   The `createUpload` is for triggering uploads by other elements *(although the selectFiles could be used as a event handler of a input element, but it seams wasteful)*, but maybe some helpers could be exported separately so that using input elements could be done with similar api. This would make it easily swappable between these two when needed.
I imagine this could be done by providing a directive or an input event handler for input elements.

3. Shouldn't the primitive be called `createUploader` or `createFileUploader`? It just odd to have two verbs together.

@davedbase 
Creating an input component is a necessary trick to be able to programatically trigger file uploads, for you would want to upload files by clicking on a normal button. I would say that this is a must have. The component isn't being connected to the dom anyway.

frixaco:
@thetarnav
Thank you very much for suggestions! 
I'm trying to get away from the "porting" mindset now)

1. Ah, yes! Now I know why I was feeling uncomfortable with how I typed it :)
2. Additional helpers will definitely be useful I think. (maybe some from the snippet in my last comment?)
3. ✓
4. ✓
5. I think yes. I was also thinking about adding async user callback support. I was wondering whether it's a good idea/feasible  to make it possible to pass async user callback and depending whether it's a promise or not, handle it accordingly.

packages/upload/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+  const handleFilesInput = async (
+    event: (Event | DragEvent) & {
+      currentTarget: HTMLInputElement;
+      target: Element;
+    }
+  ) => {
+    const filesArr: File[] = Array.from(
+      event?.currentTarget?.files || (event as DragEvent)?.dataTransfer?.files || []
```
frixaco, line 129: Seems like I can use `change` event handler when I drag'n'drop files, so I decided to handle drop event in one handler fn. Would it be better if made a separate `createDropzone` utility as @davedbase suggested?

davedbase: Sorry for the delay! It's been a busy start to the week! I think a separate dropzone would be nice compliment or addition to this package.

thetarnav:
Published stage-0 at [`@solid-primitives/upload@0.0.100`](https://www.npmjs.com/package/@solid-primitives/upload).
I think we may have been a little bit too quick with merging :P but I've cleared some of the missing peaces, just so we have a live package available for first usage.
If you have any ideas for improvements @frixaco, PRs will always be welcome :)

davedbase:
Ah shit I should have caught this. Thanks for handling it.

frixaco:
Thanks @thetarnav @davedbase! And sorry for taking so long:) Past two months have been quite busy.

Sure, I'll continue to improve the package and then probably start working on another primitive.

PR #89, opened 2022-04-04
make node-fetch optional

atk:
To handle issue#88, I changed the server entry so that node-fetch was only required once fetch was called for the first time, with a soft failure function emitting just a nice console message and rejecting every request.

I've used the same pattern in solid-register for optional dependencies and found it very effective.

### DIFF
--- a/packages/fetch/README.md
+++ b/packages/fetch/README.md
@@ -19,6 +19,18 @@ npm install @solid-primitives/fetch
 yarn add @solid-primitives/fetch
 ```
 
+### Additional requirements
+
+Since nodejs 17.5.0, the fetch API is available in node via the `--experimental-fetch` command line option. From version 18.0.0 upwards, it is supposed to become available out of the box. If you want to use `createFetch` on your server, but your nodejs version does not support the fetch API, you need to install node-fetch alongside this primitive:
+
+```bash
+npm install node-fetch
+# or
+yarn add node-fetch
+```
+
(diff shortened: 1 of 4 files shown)


thetarnav (approved):
The fetch fallback looks very slick!
Btw the `graphql` package is also using fetch, and is "polyfilling" it on the server. Would it make sense to create a common `fetch-fallback` package for that it? Or should the [createGraphQLClient](https://github.com/solidjs-community/solid-primitives/tree/main/packages/graphql#createGraphQLClient) be using the `createFetch` internally, or just having a dependency to the `fetch` package?
Also, should the `fetch` package be marked in package.json with `sideEffects: true` as it is doing this kind of polyfilling?

atk:
I think we should either have a similar pattern in graphql or use createFetch directly. In both cases, we should document the behaviour. Yes, forgot to add sideEffects: true, did that just now. Nice catch, thanks!

I'm the last person to rely on external dependencies without a good reason. However, the fetch API is complex enough that we should consider using node-fetch, which seems a pretty decent and maintained package - especially if we only need it in some cases and may deprecate the use once node-17 or later are no longer supported.

I think we should let graphql either use the same pattern or createFetch, but that's one for a different PR.

thetarnav:
Nice! Yeah I agree with a different PR, this one is good as is.
I mentioned the "fetch-fallback" package not as our alternative to node-fetch, but just as a "fallback loader" thing, that would attach this node-fetch fallback to the globalThis as you did. This way we *(and others)* would be able to use this pattern without depending on our `fetch` package if the createFetch primitive is not used there.

PR #90, opened 2022-04-05
make category names headers

thetarnav:
Makes category names in to a `<h4>` in the primitives list, so that you could use them as anchors. For sending people links leading to specific category.


davedbase:
This looks good to me.

PR #94, opened 2022-04-11
lower throttle/debounce to stage 1

trusktr:


thetarnav:
The primitives list in the root readme is just an output generated from the "primitive" filled values in the package.json of all the packages. So to lower the stage, that filled would need to be edited. Additionally there is a stage badge in the packages readmes, that would also need to be changed manually.
As for the stage lowering idea itself, I like it. It would force us to go over those packages again, reevaluating them together.

thetarnav:
Moved to `scheduled` package in #134

PR #97, opened 2022-04-12
storage: allow disabling sync events

atk:
See https://discord.com/channels/722131463138705510/958428791124951050/963507468359585803


thetarnav (approved):

thetarnav:
LGTM :)
The added comments to types are nice.
But wow this file is a beast, probably should be refactored a bit. The three places where you've added the `props.sync` check are almost identical.

PR #98, opened 2022-04-13
Document event-listener w.r.t. refs

edemaine:
I had to read the code to check whether this was the case: `createEventListener` can take standard `let` refs. The current examples use `document.getElementById` and signal refs, but not normal refs.

### DIFF
--- a/packages/event-listener/README.md
+++ b/packages/event-listener/README.md
@@ -46,6 +46,12 @@ const clear = createEventListener(
 // to clear all of the event listeners
 clear();
 
+// when element is not wrapped in an accessor,
+// runs in createEffect so refs are already bound:
+let ref;
+createEventListener(ref, "mousemove", e => {});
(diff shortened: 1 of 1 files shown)


thetarnav:
This isn't right, it can take normal let refs, but they have to be passed as a function, so that when they are accessed in the effect, the ref is already assigned.
```ts
let ref
createEventListener(() => ref, "click", () => {});
```
If you want to pass the value, the whole primitive has to be wrapped in onMount
```ts
let ref
onMount(() => {
   createEventListener(ref, "click", () => {});
})
```

edemaine:
Oops, duh. I misread `if (isFunction(targets))` as `if (!isFunction(targets))` and got confused. Pushed a revision that seems more correct and useful. 🙂

thetarnav:
Looks good now, thanks for that :)

PR #99, opened 2022-04-14
timer: add createPolled, simplify createTimer code, allow accessor delay

otonashixav:
- Add `createPolled`, which polls a function. Returns an accessor containing the last returned value of the function.
- Removes extra logic in `createTimer` leftover from the original react code
- Allow `createTimer` and `createPolled` to accept accessors for their delays. If set to `false`, disables them.
- Simplify `schedule` parameter - now accepts `setTimeout` or `setInterval` instead of `Schedule.Timeout` or `Schedule.Interval`.
- `Schedule` is now an object containing keys `Timeout` and `Interval` for backwards compatibility.
- `createTimer` is now also exported as a named export in addition to the default export.
- Adds tests using uvu and solid-register

This is a more complex version of #100 that allows passing a delay as an accessor:
```ts
createTimer(callback, () => delay());
// use as polled()
const polled = createPolled(() => new Date(), () => delay());
```
The method in #100 is also usable here.


thetarnav:
What about a version where the `createTimer` doesn't use `createEffect` internally so it has to be wrapped inside one to be reactive, while the `createPolled` – depending on the type of `timeout` arg – is wrapping the `createTimer` with `createEffect` or not.
This is because the usage of `createPolled` that has to be wrapped with memo is not good, and recreating signals every run is wasteful. Memos and signals break this composable (simple version) pattern when used internally.
I know that the API may seam inconsistent, but we can be consistent about it.
For example a `createEventListener` may match the `createTimer` api, where it needs to be wrapped in a `createEffect` if you want to change the targets of listeners, while the `createEventSignal` would do this internally, like `createPolled`.
The usage:
```ts
const [delay, setDelay] = createSignal(1000)

createTimer(callback, 1000)
createEffect(() => createTimer(callback, delay()));

const polled = createPolled(() => new Date(), 1000)
const polled = createPolled(() => new Date(), delay)
```

Or... we could introduce another function variant, that wouldn't create any computations, only attach onCleanup (like the simpler version of createTimer), that could be called `runTimer` or `addTimer`, that would need to be used inside own effects. And leave the `createTimer` and `createPolled` as they are in this PR.

Also, what about returning an `update()` function from `createPolled` for manual updating of the signal value. This it how it works in the `createDateNow`.
```ts
// updates every second:
const [now] = createDateNow(1000);

// won't autoupdate:
const [now, update] = createDateNow(false);

// update manually:
update();
```

otonashixav:
Honestly working on this made me realise that I would prefer to have fewer primitives that implement unopinionated behaviour for interacting with non-solid APIs... so now I'm not sure if `createPolled` which is a trivial implementation of `createTimer` makes sense to be implemented side-by-side with `createTimer`. 

I can see a low-level `createTimer` which only accepts the same parameters as `setTimeout` or `setInterval`, and a higher-level `createTimer` and `createPolled` which also accept variable delays and other opinionated options, but I don't think they should be in the same file, or possibly even library. 

Currently solid-primitives doesn't treat the low-level primitives differently from the higher-level primitives, but IMO it should consider doing so. Is this an overly subjective take?

davedbase:
I'll step in here to try readdressing the goals of Solid Primitives in the hopes that it guides the conversation.

We want Solid to be perceived as a native-oriented library. Less abstraction and less magic means better DX. Our Primitives should feel like the reactive wrappers for browser APIs. In cases where that's difficult or warranted, making it feel ergonomic or as close to a natural native feeling is the goal. In some cases bringing "sane" approaches to simplifying the browser is certainly a goal as well.

With that being said, I'll suggest that we introduce `createSchedule(() => {}, 500, Schedule.Interval)` as the base primitive you mentioned. Then `createTimer()` can eliminate the schedule value and `createPolled` can be introduced as a new opinionated option. Perhaps even `createInterval()` should also be introduced? This all with the goal/intention to match the native browser API and enable composability.

otonashixav:
Closed due to discussion on discord.

PR #100, opened 2022-04-14
timer: simplify `createTimer`

otonashixav:
- Removes extra logic in `createTimer` leftover from the original react code
- Simplify `schedule` parameter - now accepts `setTimeout` or `setInterval` instead of `Schedule.Timeout` or `Schedule.Interval`.
- `Schedule` is now an object containing keys `Timeout` and `Interval` for backwards compatibility.
- Adds tests using uvu and solid-register

I prefer this approach to https://github.com/solidjs-community/solid-primitives/pull/99


otonashixav:
Closed due to discussion on discord.

Issue #104, opened 2022-04-15
add `omitProps` and `pickProps` primitives


thetarnav:
This proposes adding two new props-related primitives to the `props` package: `omitProps` and `pickProps`.
Their purpose would be to address specific uses of creating a subset of props object. Which is currently possible by utilising the `splitProps` exported by Solid, but there are some possible gains:
- usually we just want to remove some keys from props, that we don't want to pass further to some jsx component by doing `<div {...props} />`.  The rest of the props can still be accessed using the original props object.
- more compact api for those specific cases – no need to destructure a props tuple & picked/omitted keys could be passed as a `...rest` argument.
- improve performance – no need to create two new props objects, only create the one with the needed set of keys.
- maybe an option to pass default props as an optional argument to avoid calling another primitive just for that. We have a bit more room for additional args. *(this would mean that the keys cannot be a ..rest arg)*

### `omitProps`
```tsx
function omitProps<T, K extends keyof T>(props: T, ...keys: K[]): Omit<T, K>

const MyComp = props => {
   const others = omitProps(props, "dont", "pass", "this", "count", "further");
   props.count // <- the rest of props can still be used
   return <Comp {...others} count={0} />
}
```

### `pickProps`
```tsx
function pickProps<T, K extends keyof T>(props: T, ...keys: K[]): Pick<T, K>

const MyComp = props => {
   const attrs = pickProps(props, "style", "class", "onClick");
   props.count // <- the rest of props can still be used
   return <div {...attrs} />
}
```

davedbase:
I believe this is resolved in Solid 2.0 with new primitives or at least should be adapted for that new reality.

Issue #101, opened 2022-04-14
`utils` package cleanup


thetarnav:
Currently the shared `utils` package is chaotic and over-used for cases that are better implemented inline.

The goals would be to:

- [ ] Remove "typecasting" helpers from utils – typecast manually instead
Those functions would be: `tuple`, `definite`, `withFallback`, `withFallbackFn`, `includes`, `keys`, `entries` and `createProxy`

*`includes`, `keys` and `entries` could honestly be left, as long as they are implemented like so: (just targeting the types)*
```ts
const entries = Object.entries as <T extends AnyObject>(obj: T) => [keyof T, T[keyof T]][];
```

- [ ] Remove all `isNumber()` like type-checking functions – should be done inline

- [ ] Remove `Dispose`, `Clear`, `ClearListeners` & `Fn` types that are aliases to a `VoidFunction` – one need to figure out what they mean

- [ ] Extract `promiseTimeout` & `raceTimeout` to be in a separate package. Maybe `promise`?

- [ ] Extract `createTrigger` & `createTriggerCache` into separate package. Maybe `trigger`?

otonashixav:
> `includes`, `keys`, `entries`
Can also be done via declaration merging:

```ts
declare interface Array<T> {
  includes<U>(searchElement: T extends U ? U : T, fromIndex?: number): boolean;
}


const a: number[] = [];
a.includes("" as number | string); // no error
```

Which wouldn't require any imports. I'm not sure on best practices in this regard though, and there may not be a way to limit the effects to a single file.

PR #110, opened 2022-04-19
ui: initial setup + first primitive

fabien-ml:
- `ui` package setup (ts, jest, eslint).
- Moved all works from `@hope-ui/primitives`.
- Updated root `README.md`.
- Added `README.md` for the `visually-hidden` primitive.


thetarnav:
This is awesome! :)
Definitely seeing some utility functions that could be moved to more general home.

For example small utilities like `testPlatform` & `isMac` could be moved to the general `utils` package.

`combineProps` is a very interesting primitive that could be added to the `props` package along with [ones previously proposed](https://github.com/solidjs-community/solid-primitives/issues/104).
It seams it doesn't combine the `style` prop though, is it intentional? I know that it might be tricky to [combine string styles with object styles](https://discord.com/channels/722131463138705510/932856348855660544/965994371616997466), but as those are also DOM props, this could be useful. Also this is combining `class` but not `className` when SolidJS accepts two syntaxes.

We also had this [kind of assertions](https://github.com/solidjs-community/solid-primitives/blob/ui/packages/ui/src/utils/assertion.ts) in the `utils` package, but after some discussion, we are getting rid of them, to assert always inline: #101 
What do you think of that? Most of this kind of assertions would be a waste to have to call a function to do them. But some of more complicated ones (`isObject` only pretty much) might still be useful.

The `createControllableSignal` also seams like something to move, but... it also seams that it's usage is closely tied to ui, like checkboxes, that it might make sense to keep it here, for not at least.

I'm fine taking care of it, just wondering if you have some opinions on it :)

thetarnav:
Also, what do you think about providing an imperative helper of things like `createVisuallyHidden`? I mean that there could be a function like `hideVisually` that one can use like he pleases, if the primitive doesn't suit a more custom usecase.

fabien-ml:
What is the process for moving things in other packages ?

`combineProps` doesn't combine style because of the reason you mentioned (`style` can be string or object).
I can add combine `className` but it will be two separate props. `class` will not be combined with `className` for example.

For assertions maybe an `assertions` package ? `isNumber()` is simple but `isObject()` will be a pain to copy/paste.

`createControllableSignal` is heavily related to component state, can be renamed to `createControllableState` to avoid confusion.

For `createVisuallyHidden` the style is already returned by the primitive, but for simpler use case `getVisuallyHiddenStyles` can be useful.

What `hideVisually` would do ?

atk (commented):
First of all, all these primitives see to expect some flavor of styled components. This at least should be documented.

thetarnav:
Yeah, I didn't mean to combine `class` + `className`, just handle `className` as well. Also `classList`, would that be useful to combine?

`hideVisually` would just apply those "visually hidden" styles to passed HTML element. Does it makes sense? `getVisuallyHiddenStyles` would be good enough though.
```ts
hideVisually(el)
```

fabien-ml:
> First of all, all these primitives see to expect some flavor of styled components. This at least should be documented.

what do you mean by that ?

thetarnav:
`utils` are intended for internal usage primarily, so they don't require documentation. buttt... why not, some of the people (like @fabien-ml) are using them anyway. At least for some of the main functions it could be useful.

packages/ui/src/interactions/createFocusVisible.ts
```diff
@@ -0,0 +1,262 @@
...
+  return modality;
+}
+
+/**
+ * If this is attached to text input component, return if the event is a focus event (Tab/Escape keys pressed) so that
+ * focus visible style can be properly set.
+ */
+function isKeyboardFocusEvent(isTextInput: boolean, modality?: Modality, e?: HandlerEvent) {
+  return !(
+    isTextInput &&
```
atk, line 208: Can't you check the target to find out if it's a text input, i.e. an instance of HTMLInputElement with a textual type or HTMLTextArea or something inside an element that is contenteditable.

fabien-ml: Honestly I don't know. I've ported the react code as is because I don't know if major changes will break other primitives I port in the future. With your solution I'm not sure if i should check the target or currentTarget.

fabien-ml:
> `utils` are intended for internal usage primarily, so they don't require documentation. buttt... why not, some of the people (like @fabien-ml) are using them anyway. At least for some of the main functions it could be useful.

Things like `MaybeAccessor` and `access` will definitely be useful for library authors.

packages/ui/src/interactions/createFocusVisible.ts
atk: In input-mask, I've checked either.

fabien-ml: In react-spectrum the `isTextInput` is used in the `useFocusRing` hook to apply focus ring styles to the wrapping element of a "text field" or the "text field" itself.

fabien-ml:
All the works will be moved to https://github.com/solidjs-community/solid-aria

PR #112, opened 2022-04-21
new primitive: selection

atk:
Selection inside textual inputs is simple; in contentEditable elements, it gets complicated. So simplification is a worthwile endeavor.

### DIFF
--- a/packages/selection/README.md
+++ b/packages/selection/README.md
@@ -1,0 +1,69 @@
+<p>
+  <img width="100%" src="https://assets.solidjs.com/banner?type=Primitives&background=tiles&project=Selection" alt="Solid Primitives Selection">
+</p>
+
+# @solid-primitives/selection
+
+[![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg?style=for-the-badge)](https://lerna.js.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/selection?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/selection)
+[![version](https://img.shields.io/npm/v/@solid-primitives/selection?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/selection)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
+
+Primitive that helps reading/setting cursor position and selections both in text fields and contenteditable elements.
+
+## Installation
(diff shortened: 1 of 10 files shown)


packages/selection/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Usage
+
+
+```ts
+import { createSelection } from "@solid-primitives/selection";
+
+const [selection, setSelection] = createSelection();
+
+selection(); // [HTMLDivElement, 1, 3]
+setSelection([document.querySelector('input'), 3, 3]);
```
thetarnav, line 31: I was confused by this initially. "Where did those values came from?"
```suggestion
selection(); // T: [HTMLElement, number, number]

setSelection([document.querySelector('input'), 1, 3]);
selection(); // => [<input/>, 1, 3]

// when no node is selected:
selection(); // => [null, NaN, NaN]
```

packages/selection/README.md
```diff
@@ -0,0 +1,69 @@
...
+const [selection, setSelection] = createSelection();
+
+const ibanMask = anyMaskToFn('aa99999999999999999999');
+
+const inputMaskHandler = (ev) => {
+  const [node, start, end] = selection();
+  if (ev.currentTarget === node) {
+    const [value, selection] = ibanMask(node.innerHTML, [start, end]);
+    node.innerHTML = value;
+    setSelection(node, selection[0], selection[1]);
```
thetarnav, line 49: I believe you have to pass an array? *(I would prefer to have a ...rest option though)*
```suggestion
    setSelection([node, selection[0], selection[1]]);
```

packages/selection/src/index.ts
```diff
@@ -0,0 +1,96 @@
+import { Accessor, createEffect, createSignal, onCleanup, Setter } from "solid-js";
+
+export type HTMLSelection = [node: HTMLElement | null, start: number, end: number];
+
+export const getTextNodes = (node: Node) => {
```
thetarnav, line 5: If you are exporting this function, maybe would it be worth it to mention it in the readme, as an additional *vanilla* utility?

thetarnav:
Is there an easy way to clear the selection?
The only way is to do `setSelection([null, NaN, NaN])` right?

packages/selection/README.md
atk: Thanks, good catch.

packages/selection/src/index.ts
atk: I was exporting it mostly for testing purposes. If we're in need of such a function in other priitives, that's when I would move it to utils.

atk:
> Is there an easy way to clear the selection? The only way is to do `setSelection([null, NaN, NaN])` right?

That's the way, yes. I should document that, too.

packages/selection/README.md
atk: I added a few comments.

thetarnav (approved):
Looks great now :)

PR #115, opened 2022-04-29
rewrite lerna-scripts helpers in node

thetarnav:
This removes the lerna-script dev dependency and scripts written with it's help.
The scripts are replaced with plain node.js ones – with the same functionality
Executable like so:
```
yarn update-readme
// or
yarn new-package name-of-your-package
```
This would be the first step in moving away from lerna to turborepo

Edit: I've also added an gh action for it, it should run after PR merge and update the readme itself


scripts/new-package.js
```diff
@@ -0,0 +1,7 @@
+const shell = require("child_process").execSync;
+const { join } = require("path");
+const packageName = process.argv.pop();
+const src = join(__dirname, "../template");
+const dest = join(__dirname, "../packages", packageName);
+shell(`mkdir -p ${dest}`);
```
atk, line 6: Maybe we should expend a few more lines to make this work in Windows, too. fs has a mkdir and a copyFile API; used with path to take care of different separators, this should be fairly easy.

thetarnav: Yeah good idea.
Just to be clear (I'm not that familiar with node), are you proposing to use `path` + `fs` to copy files, instead of executing shell commands? I could try doing that.. also maybe do some search and replace on the readme and package.json, so that this helper is actually doing something useful besides copy-pasting :)

scripts/update-readme.js
```diff
@@ -0,0 +1,68 @@
...
+const markdownMagic = require("markdown-magic");
+const tablemark = require("json-to-markdown-table");
+
+const getPathTo = path => join(__dirname, path);
+
+console.log("updateReadme", "Updating README documentation");
+
+const githubURL = "https://github.com/solidjs-community/solid-primitives/tree/main/packages/";
+const sizeShield = "https://img.shields.io/bundlephobia/minzip/";
+const bundlephobiaURL = "https://bundlephobia.com/package/";
```
atk, line 12: We should probably use bundle.js.org instead as it is more precise.

thetarnav: Do they have size badges?

scripts/new-package.js
thetarnav: ok done, node is surprisingly easy, wow

scripts/update-readme.js
atk: Not sure. I think the maintainer is on our discord.

atk: Just checked, no, they don't support size badges.

PR #117, opened 2022-05-04
feat: fetch allows undefined request info

atk:
for if you do not want to start a request right away


thetarnav (changes requested):
The type definition in README hasn't been updated.
Also it would be nice to have a little mention of that mechanic in there as well.
It's a good addition overall.

thetarnav (approved):

PR #120, opened 2022-05-06
Allow createAudioPlayer with no handlers

TiagoCavalcante:


atk (approved):

davedbase:
@TiagoCavalcante thanks for the suggestion. I'm going to be implementing a slight refactor of these primitives shortly. I'll be adapting to the new `make` pattern that we're using with this project. I'll merge this in for now though

Issue #116, opened 2022-05-03
[intersection-observer] `createVisibilityObserver` does not work with `display: contents`


Enteleform:
# Expected

Elements with `display: contents` will trigger the visibility observer and render in their usual manner.

# Actual

Elements with `display: contents` do not trigger the visibility observer and they are not rendered at all.

https://user-images.githubusercontent.com/10906415/166436977-aa9e9fc9-e56a-4ccc-adef-6e601a16dfa1.mp4

# Reproduction
https://codesandbox.io/s/solid-primitives-intersection-observer-forked-7cc2vz?file=/src/index.tsx

davedbase:
Hmmm interesting. I too was a bit unaware of display: contents. I've never had a use case for that. I did a bit of research and other React based primitives have stumbled on the same issue: https://github.com/thebuilder/react-intersection-observer/issues/421

I'm not sure if there's a viable alternative to this issue. According to the definition contents: `Makes the container disappear, making the child elements children of the element the next level up in the DOM`. So the issue is likely that the container isn't in viewport at all and it's skipping to the next element and hence never displaying at all.

I kind of agree with the react hooks author that this is quite the edge case, but if there's an easy solution that doesn't include hammering the primitive with more LoCs, I'd certainly entertain it.

Ideas anyone?

Enteleform:
@thetarnav

Hey, haven't figured out a workaround. I encountered the issue while working on [astro-portal](https://github.com/Enteleform/lazy-portal/tree/main/packages/astro-portal), and I ended up just setting the portal's width & height to 100% to offload the sizing responsibility to the container where the portal is placed.

Dunno what the best way to approach this would be, I've encountered `display: contents` issues in several contexts (*portals, table components, drag-&-drop libraries*) and haven't seen any of them resolved. Is it that the DOM API doesn't have adequate means of interacting with elements with that property?

thetarnav:
> Is it that the DOM API doesn't have adequate means of interacting with elements with that property?

Yeah, the IO is just ignoring elements with that style. Similarly, you cannot put any CSS transform on that element successfully.
I also had some issues with astro islands with it. This is what I had to do:
```ts
const resolved = children(() => props.children)

onMount(() => {
	let _ref = resolved()
	let ref!: HTMLElement

	while (_ref instanceof HTMLElement && !ref) {
		if (getComputedStyle(_ref).display === "contents") _ref = _ref.firstChild
		else ref = _ref
	}
[...]

PR #122, opened 2022-05-07
Adapt to new `make` pattern + improve tests

davedbase:
@thetarnav before I move ahead with polishing this, can I get some feedback on the primitives API with the new `make` prim? If this makes sense I'll go ahead and polish up the package and adapt the tests if needed.

### DIFF
--- a/packages/intersection-observer/README.md
+++ b/packages/intersection-observer/README.md
@@ -11,7 +11,8 @@
 
 A range of IntersectionObserver API utilities great for different types of usecases:
 
-- `createIntersectionObserver` - Creates a basic intersection observer exposing methods to manage the observable.
+- `makeIntersectionObserver` - Creates a basic Intersection Observer exposing methods to manage the observable.
+- `createIntersectionObserver` - A reactive observer primitive.
 - `createViewportObserver` - More advanced tracker that creates a store of element signals.
 - `createVisibilityObserver` - Basic visibility observer using a signal.
 
@@ -35,8 +36,8 @@ const [add, { remove, start, stop, instance }] = createIntersectionObserver(els,
 add(el)
 
 // Directive usage:
-const [observer] = createIntersectionObserver()
-<div use:observer></div>
+const [intersectionObserver] = createIntersectionObserver()
+<div use:intersectionObserver></div>
 ```
 
 ### createViewportObserver
@@ -103,5 +104,9 @@ Updated to Solid 1.3
 1.2.2
 
 Minor improvements
+
+1.3.0
+
+Migrated to new `make` pattern and improved primitive structures.
 
 </details>
--- a/packages/intersection-observer/src/index.ts
+++ b/packages/intersection-observer/src/index.ts
@@ -1,4 +1,4 @@
-import { onMount, onCleanup, createSignal } from "solid-js";
+import { onMount, onCleanup, createSignal, createEffect } from "solid-js";
 import type { JSX, Accessor } from "solid-js";
 import { access, MaybeAccessor } from "@solid-primitives/utils";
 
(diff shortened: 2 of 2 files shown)


davedbase:
- [ ] Finalize API design
- [ ] Make functional minimal demo
- [ ] Port and validate tests

packages/intersection-observer/src/index.ts
```diff
@@ -57,31 +57,97 @@ export type E = JSX.Element;
...
+ * );
+ * ```
+ */
+export const createIntersectionObserver = (
+  elements: MaybeAccessor<Element[]>,
+  onChange: IntersectionObserverCallback,
+  options?: IntersectionObserverInit
+) => {
+  const { add, reset } = makeIntersectionObserver([], onChange, options);
+  createEffect(() => {
```
thetarnav, line 106: Effect could be created only if the `elements` arg in an accessor.

packages/intersection-observer/README.md
```diff
@@ -104,4 +105,8 @@ Updated to Solid 1.3
 
 Minor improvements
 
+1.3.0
```
thetarnav, line 108: Breaking changes should probably have a major release.

thetarnav (commented):
Looks good.
Have you seen this bug: #116 ? Might be worth to check it by the way. It doesn't look like something that we could do on our end, but maybe.

packages/intersection-observer/README.md
davedbase: Yup, I'll bump everything when I'm confident it's ready. Sorry just threw this together really fast.

packages/intersection-observer/src/index.ts
davedbase: Well if we do that it's just basically a `makeIntersectionObserver` so I've changed it to force an Accessor since that's the only thing that makes sense here. I feel like it's so stripped down now but perhaps this is indeed the basic reactive use?

thetarnav: That would be the point in using it I think.

packages/intersection-observer/src/index.ts
```diff
@@ -57,31 +57,97 @@ export type E = JSX.Element;
...
+export const createVisibilityObserver = (
+  element: MaybeAccessor<Element>,
+  options?: IntersectionObserverInit & {
+    initialValue?: boolean;
+    once?: boolean;
+  }
+): [Accessor<boolean>, { start: () => void; stop: () => void; instance: IntersectionObserver }] => {
+  const [isVisible, setVisible] = createSignal(options?.initialValue ?? false);
+  const { start, stop, instance } = makeIntersectionObserver(
+    [access(element)],
```
thetarnav, line 138: I believe this isn't reactive now, if the element in an accessor. It would have to be `createIntersectionObserver` for an accessor, and `makeIntersectionObserver` for just element.

packages/intersection-observer/src/index.ts
davedbase: Yes yes, lol. I agree.

packages/intersection-observer/src/index.ts
davedbase: Oops good point. I added an effect if an accessor is supplied.

davedbase:
> Looks good. Have you seen this bug: #116 ? Might be worth to check it by the way. It doesn't look like something that we could do on our end, but maybe.

Yeah I don't think we can handle that edge case without doing some JS acrobatics. Not worth the effort IMHO.

davedbase:
@thetarnav I think I'm going to call this a wrap and implement the tests + demo.

davedbase:
@thetarnav @atk ok this is ready for final-final review.

atk (approved):

PR #123, opened 2022-05-07
Audio: Adapt to new `make` pattern + improve tests

davedbase:
- [x] Finalize API design
- [x] Make functional minimal demo
- [x] Port and validate tests
- [x] Update README and package.json


thetarnav:
Looks good.
The event-listener pkg could reduce the logic even more here.
Also, what about returning a Store, instead of an object of accessors in the new `createAudio`?

davedbase:
> Looks good. The event-listener pkg could reduce the logic even more here. Also, what about returning a Store, instead of an object of accessors in the new `createAudio`?

Yup, good suggestion!

thetarnav:
I've made a [`createStaticStore`](#https://github.com/solidjs-community/solid-primitives/blob/027231ae2981f490740a5f1d1101c9539624a5a6/packages/utils/src/index.ts#L323) for such smaller usecases btw.

thetarnav (changes requested):
- [ ] fix tests
- [x] The docs could use some usage examples / types
It's hard to get an idea of what certain primitives will help you with, by just looking through the readme
- [x] There are few random console.logs in index.ts
- [x] expressions like these: `src instanceof Function ? src() : src` could be replaced with `access(src)`, since you already have utils package installed
- [x] Better types for event handlers
```ts
{
   [K in keyof HTMLMediaElementEventMap]?: (event: HTMLMediaElementEventMap[K]) => void;
}
// instead of 
{ [key: string]: EventListener }
```

davedbase:
@thetarnav I adjusted everything, just that pesky test that I have to track. Sigh.

davedbase:
@thetarnav tests are passing (wooohoo). Last review and we publish. Also there's a nice demo I put together

packages/audio/test/index.test.ts
```diff
@@ -1,83 +1,87 @@
...
-  test("test srcObject reactive value path", async () => {
-    const { player } = createAudioPlayer(() => new MediaSource(), []);
-    expect(player.srcObject).toBeInstanceOf(Object);
-  });
-  test("test player with no handlers", async () => {
-    const { player } = createAudioPlayer(() => new MediaSource());
-    expect(player.srcObject).toBeInstanceOf(Object);
-  });
-});
+const testPath = 'https://github.com/solidjs-community/solid-primitives/blob/audio/packages/audio/dev/sample1.mp3?raw=true';
```
thetarnav, line 9: Better not delete that branch haha :D

thetarnav (approved):
Alright, the `server.ts` needed some corrections, but I took care of it. Looks good otherwise.

davedbase:
Shoot good point, I changed things last minute. Thanks for adjusting it! Going to merge and publish now.

PR #124, opened 2022-05-07
Clipboard: Adapt to new `make` pattern + improve tests

davedbase:
- [x] Finalize API design
- [x] Make functional minimal demo
- [x] Port and validate tests
- [x] Update README and package.json


packages/clipboard/src/index.ts
```diff
@@ -26,11 +26,11 @@ declare module "solid-js" {
...
  * ```ts
- * const [ write, read ] = createClipboard();
+ * const [ write, read ] = makeClipboard();
  * console.log(await read());
  * ```
  */
-const createClipboard = (): [
+export const makeClipboard = (): [
   write: ClipboardSetter,
   read: () => Promise<string>,
```
thetarnav, line 35: Should we take this opportunity to switch [write, read, something_else] to [read, write, something_else]?
Maybe writing is more common operation, but [read, write] is a more common api.

thetarnav (commented):
`makeClipboard` seams weird... It doesn't actually do anything, functions that it provides could be exported separately.
But then again... Those functions separately, wouldn't do anything special either. Just alias the browser API. Maybe this pattern just doesn't fit here? Don't know if we should export functions that don't have a clear purpose.

packages/clipboard/src/index.ts
```diff
@@ -44,6 +44,31 @@ const createClipboard = (): [
   return [write, read, { newItem }];
 };
 
+/**
+ * Creates a new reactive primitive for managing the clipboard.
+ *
+ * @return write - A method to write to the clipboard
+ * @return read - Read from the clipboard
+ * @return helpers - helper functions:
```
thetarnav, line 52: old

davedbase: Yeah I haven't adjusted ANY of these yet. Just focusing on the API first.

packages/clipboard/src/index.ts
davedbase: Yup I'll be sure to do that, good point.

davedbase:
@thetarnav I had another look at makeClipboard and I think even though it doesn't do any clean-up I think it's important to wrap this base functionality. It's used in the directive and the reactive primitive. I like that we're wrapping ClipboardItem as well with newItem.

davedbase:
I think I've taken this as far as I can at this point. I'd like to get it up and improve/finish the tests at another date. Would love a look and approval to merge @thetarnav @atk :-)

packages/clipboard/README.md
```diff
@@ -21,27 +21,119 @@ yarn add @solid-primitives/clipboard
...
+];
+```
+
 ### createClipboard
 
-Clipboard exports a read and write function. Note the write function is exported first for convenience as the most common use case for this primitive.
+This primitive provides full facilities for reading and writing to the clipboard. It allows for writing to clipboard via exported function or input signal. It wraps the Clipboard Async API with a resource and supplies reactive helpers to make pulling from the clipboard easy.
+
+```tsx
+const [data, setData] = createSignal('Hello);
```
thetarnav, line 47: Lol must have missed the notification. ehh
Found a typo though. The missing `'` is messing the code highlighting.

PR #125, opened 2022-05-07
Geolocation: Minor improvements and test updates

davedbase:
- [x] Complete writing tests
- [x] Validate reactivity setup and additional edge case needs are covered
- [x] Adjusted README and package.json


davedbase:
Coming to the conclusion that this package likely provides everything you'd need for geolocation. My only thought is that we could probably have createGeolocationWatcher output a proxy so that you can use:

```ts
const location = createGeolocationWatcher(true);
location; // location signal
location.error; // error is accessed separately
```

Also thinking it would be neat to have createGeolocationDistance which lets you provide a lat/lng and it calculates the distance between the current user and the supplied lat/lng/

davedbase:
@thetarnav ok I think this one's ready to go as well. Let me know what you think.

packages/geolocation/README.md
```diff
@@ -23,7 +23,7 @@ yarn add @solid-primitives/geolocation
 
 ### createGeolocation
 
-Used to fetch current geolocation data as a resource.
+Used to fetch current geolocation data as a resource. This primitive uses `createResource` to return a location, so `loading`, `error`
```
thetarnav, line 26: This sentence seems incomplete, and the code snippet below doesn't have `loading` or `error` used.

packages/geolocation/package.json
```diff
@@ -36,52 +36,47 @@
...
-    "tsup": "^6.1.2",
+    "tsup": "^5.10.1",
     "typescript": "^4.7.4",
-    "wait-for-expect": "^3.0.2"
+    "unocss": "^0.39.1",
+    "uvu": "^0.5.3",
+    "vite": "^2.9.12",
+    "vite-plugin-mkcert": "^1.7.2",
+    "vite-plugin-solid": "^2.2.6",
+    "vitest": "^0.15.1"
```
thetarnav, line 74: Some dev deps here are duplicated, and some are not needed at all.
vitest, babel-preset-solid, @types/jest and @babel/preset-env could be safely removed I think

thetarnav (approved):
👍

packages/geolocation/package.json
davedbase: Thanks for pointing that out! :-)

PR #131, opened 2022-05-17
fetch: rewrite with composability in mind

atk:
Fresh from the discussion here https://github.com/solidjs-community/solid-primitives/discussions/129, this is the first draft of a simplified, more composable fetch primitive.

This is still WIP; the retry and cache modifier are not yet done, the internal documentation and the README are not yet finished.


packages/fetch/src/modifiers.ts
```diff
@@ -0,0 +1,146 @@
...
+    throw new Error("could not read resource fetcher");
+  }
+  requestContext.fetcher = wrapper(originalFetcher) || originalFetcher;
+};
+
+export const wrapResource = <T>(
+  requestContext: RequestContext<T>,
+  wrapper: (requestContext: RequestContext<T>) => [
+    props?: { [key: string]: any },
+    actions?: { [key: string]: any }
```
thetarnav, line 23: I don't like the types here, they do not guide the user to provide correct values.
```suggestion
    props?: PropertyDescriptorMap,
    actions?: Record<PropertyKey, (...args: unknown[]) => unknown>
```
These would be better.
Also on another thought... if the props are being assigned via `Object.defineProperties`, then:
1. You have to use this lengthy syntax to have the getter assignad.
```ts
{
   aborted: { get: () => my_aborted_value },
   status: { get: () => my_status_value },
   response: { get: () => my_response_value }
}
```
2. Users can easily assign setters as well, which breaks the read/write separation. Actions would be for that.

So maybe the type for props should be:
```ts
Record<PropertyKey, Accessor<unknown>>
```
Used like this:
```ts
{
   aborted: () => my_aborted_value,
   status: () => my_status_value,
   response: () => my_response_value
}
```
And assigned like this:
```ts
for (const prop in props) Object.defineProperty(ctx, prop, { get: props[prop] })
```

atk: Actually, the types are still on my TODO list. I wanted to find a way to make sure the return value type of createFetch can be changed by the modifiers.

thetarnav: Ah right. well, that sounds like fun :D
It should work similarly to mergeProps actually.

atk: yes, but the difficulty is that we have no return type to the modifiers, so I need to create some utility types.

thetarnav: Why not have a `createModifier` then, that would provide the `wrapRequest` and `wrapFetcher` with types limited to stuff passed to generics?

atk: I don't want to add code just to satiate the type checker, so I want to try without first.

davedbase:
Re: Axios I have a lot of feelings about supporting it. This 3 year old post comes to mind: https://www.reddit.com/r/reactjs/comments/dcte0e/psa_axios_is_mostly_dead/. If you are going to support it perhaps consider using redaxios instead? It's an 800 byte alternative written by Jason Miller: https://github.com/developit/redaxios

atk:
I'll support anything with the same API.

atk:
Also, I plan on supporting TRPC in the near future.

davedbase:
I'm wondering if using Redaxios out of the box as a dependency is better than Axios. The user may not care which is used so opting for the smaller is an advantage (6.7kb vs 1kb). The user can install Axios proper if they need to override it for whatever reason. Just a thought/suggestion :-)

thetarnav:
@davedbase Wouldn't all of them be optional dependencies that user has to explicitly install if he want to use it?

davedbase:
Yah you're right, sorry I thought axios was marked as a dependency in the package.json but I read it wrong. Maybe some instructions/suggestions in the readme for the axios modifier would be best?

atk:
We should definitely document this thoroughly.

atk:
I've added the first stub for caching. The reload mechanism is still missing. I've put cache into its own file so modifiers is not blown up out of proportion.

atk:
refetch mechanism is also added now. I'm closer to finishing the main functionality and can then turn to types and tests.

packages/fetch/src/modifiers.ts
```diff
@@ -0,0 +1,204 @@
...
+      }
+    ]);
+  };
+
+export const withTimeout: RequestModifier = (timeout: number) => requestContext => {
+  wrapFetcher(
+    requestContext,
+    originalFetcher => (requestData, info) =>
+      new Promise((resolve, reject) => {
+        window.setTimeout(() => {
```
thetarnav, line 89: What's the point of `window.setTimeout()` instead of just `setTimeout()`? Won't accessing window throw on node?

packages/fetch/src/modifiers.ts
```diff
@@ -0,0 +1,204 @@
...
+          // @ts-ignore
```
thetarnav, line 90:
```suggestion
```
It doesn't seem to be needed. It's probably better to use @ts-expect-error to not over ignore stuff.

packages/fetch/src/fetch.ts
```diff
@@ -0,0 +1,244 @@
...
+  responseHandler?: (response: Response) => any;
+  disable?: boolean;
+};
+
+export type FetchReturn<T, I> = [
+  {
+    (): T | I;
+    /** if you are using withAbort(), this will contain a boolean to check if the request was aborted */
+    aborted?: boolean;
+    error: any;
```
thetarnav, line 38: Ah, I see what are you unhappy about with the types :D
It's probably worth mentioning in the docs, and what's the best path to set the correct types if needed.

packages/fetch/src/server.ts
```diff
@@ -6,14 +6,29 @@ if (!globalThis.fetch) {
...
+  withAbort,
+  withCatchAll,
+  withTimeout,
+  withRetry,
+  withRefetchEvent,
+  wrapFetcher,
+  wrapResource
+} from "./modifiers";
+export { withCache, withCacheStorage } from "./cache";
+export { fetchRequest, axiosRequest } from "./request";
```
thetarnav, line 34: 👀 

I advise you to set up this test suite to catch unmatched exports: https://github.com/solidjs-community/solid-primitives/blob/rework-visibility-observer/packages/intersection-observer/test/exports.test.ts

packages/fetch/src/modifiers.ts
```diff
@@ -0,0 +1,204 @@
...
+        lastRequestRef.current &&
+        !(ev.type === "visibilitychange" && document.visibilityState !== "visible") &&
+        filter(...lastRequestRef.current, ev)
+      ) {
+        requestContext.resource?.[1].refetch();
+      }
+    };
+    events.forEach(name => window.addEventListener(name, handler));
+    getOwner() &&
+      onCleanup(() => events.forEach(name => window.removeEventListener(name, handler)));
```
thetarnav, line 203: +1 on the `window.` thing. Also, this should probably be wrapped with `if (!isServer) {}` anyway.
You are accessing `document.visibilityState` higher up as well.

packages/fetch/src/cache.ts
```diff
@@ -0,0 +1,111 @@
...
+          });
+        }
+    );
+    requestContext.wrapResource();
+    Object.assign(requestContext.resource![1], {
+      invalidate: (requestData: FetcherArgs) => {
+        try {
+          delete requestContext.cache[serializeRequest(requestData)];
+        } catch (e) {
+          DEV &&
```
thetarnav, line 85:
```suggestion
          DEV && !isServer &&
```
On the server, `DEV` is an empty object, which will be truthy. `DEV && !isServer` should do the trick, or just use [warn from utils](https://github.com/solidjs-community/solid-primitives/blob/main/packages/utils/src/index.ts#L42).

thetarnav (commented):
Here are a couple of things I've caught. Generally, the amount of work put into this package is insane. Really good stuff.

packages/fetch/src/fetch.ts
```diff
@@ -0,0 +1,244 @@
...
+        : args[0];
+    if (!info) {
+      return undefined;
+    }
+    const init =
+      typeof args[1] === "function"
+        ? (args[1] as Accessor<FetcherArgs[1]>)()
+        : isOptions(args[1])
+        ? undefined
+        : (args[1] as RequestInit);
```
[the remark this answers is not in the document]
atk: While I wasn't sure this would make a difference since in this case the memo would be static so it would basically only use a little more memory, it's still something that can be optimized. I'll rewrite this part.

packages/fetch/src/modifiers.ts
atk: You're right. Fixed.

packages/fetch/src/modifiers.ts
atk: That's a remnant from before I fixed the types.

packages/fetch/src/fetch.ts
atk: I think the JSDoc comment will suffice.

packages/fetch/src/server.ts
atk: That's great! It should be added to our template!

packages/fetch/src/modifiers.ts
atk: It won't work on the server anyways, so I'll just replace it with a shim there.

packages/fetch/src/cache.ts
atk: I think we're not really interested in this warning on the server, and don't want to spam log lines.

thetarnav: Yeah, that's the point. I think that with `DEV &&` alone, it'll log on the server too.

thetarnav (approved):
Looks good, let's ship it :)

PR #134, opened 2022-05-21
Merge throttle & debounce into scheduled package

### LINKED ISSUE #64: createThrottle does not call on the leading edge
Thanks for your awesome repository! 

I'm a little confused by `createThrottle`, as I expect a throttled function to fire immediately when it's the first invocation (in the throttle interval). [Here's a little example of what I mean](http://demo.nimius.net/debounce_throttle/).

Current behavior from `createThrottle`: the throttled function waits the full delay even if it's the first invocation. I would expect debounce to behave that way, but not throttle. 

To reproduce: use the CodeSandbox in the [`createThrottle` README](https://github.com/davedbase/solid-primitives/tree/main/packages/throttle).

<details><summary>Caveat: I figure it's "by design," but in that case I think the docs should be updated to clarify the behavior.</summary>

I'm aware sometimes people use the word "throttle" more generally; and/or, sometimes there's a flag like `{leading: true}` that enables/disables the leading-edge invocation. ([An example of that can be seen in lodash](https://github.com/lodash/lodash/blob/master/throttle.js#L35). Optionally, see also [p-debounce's `{before: true}` option](https://github.com/sindresorhus/p-debounce#before))

But I think in the general usage, "throttle" is contrasted against "debounce" — some references: [1](https://css-tricks.com/debouncing-throttling-explained-examples/), [2](https://css-tricks.com/the-difference-between-throttling-and-debouncing/), [3](https://stackoverflow.com/a/52192108/884640), [4](https://stackoverflow.com/a/62017005/884640), [5](https://dev.to/rxjs/debounce-vs-throttle-vs-audit-vs-sample-difference-you-should-know-1f21), [6](https://towardsdev.com/debouncing-and-throttling-in-javascript-8862efe2b563)  

So if `createThrottle` is _not_ going to call on the leading edge by default, perhaps to keep backwards-compatibility, I think the README/demo could be updated to emphasize the behavior.

</details>

### LINKED ISSUE #93: requrest to lower throttle, debounce to stage 1
And for any similar APIs, the same.

They have different shape APIs, but they should be the same for consistency. I think they should be stage 1, based on the [proposed staging process](https://github.com/solidjs-community/solid-primitives/discussions/92).

If we move too fast, we're going to end with inconsistent APIs like throttle and debounce currently are, or maybe the wrong APIs.

We need a way to publish pre-stage-3 packages with clear labeling that indicates consensus isn't reached on the API shape or features yet. Maybe we do it with certain addition to the package name, or use only `0.0.0-*` pre-release versioning, or something.

### LINKED ISSUE #133: createThrottle: throttled function is called with outdated value
I had to switch an app to lodash `_.throttle`, for now, because `createThrottle`'s trailing call happens with an outdated value instead of the most recent.

thetarnav:
Removes `throttle` & `debounce` packages in favor of new combined `scheduled` package.

Fixes #133
Fixes #93
Fixes #64

[THE README](https://github.com/solidjs-community/solid-primitives/tree/schedule/packages/scheduled#readme)

Adds new primitive: `leading`. For debouncing/throttling with callback called on leading edge.
```ts
// with debounce
import { leading, debounce } from "@solid-primitives/scheduled";

const trigger = leading(debounce, (message: string) => console.log(message), 250);
trigger("Hello!");
trigger.clear(); // clears a timeout in progress

// with throttle
import { leading, throttle } from "@solid-primitives/scheduled";

const trigger = leading(throttle, (message: string) => console.log(message), 250);
trigger("Hello!");
trigger.clear(); // clears a timeout in progress
```


davedbase:
@trusktr thoughts on this before we merge it?

packages/scheduled/README.md
```diff
@@ -0,0 +1,116 @@
...
+const trigger = leading(throttle, (message: string) => console.log(message), 250);
+trigger("Hello!");
+trigger.clear(); // clears a timeout in progress
+```
+
+## Scheduling explanation
+
+This package provides 4 different methods for scheduling a callback. Pick one that suits your application.
+
+```
```
atk, line 90: loving this little schematic here, this is really great.

packages/scheduled/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  const func: typeof callback = (...args) => {
+    if (!isScheduled) callback(...args);
+    isScheduled = true;
+    scheduled();
+  };
+  const clear = () => {
+    isScheduled = false;
+    scheduled.clear();
+  };
+  if (getOwner()) onCleanup(() => (isScheduled = false));
```
atk, line 110: why not use clear() here to also clear any timeout?

packages/scheduled/test/debounce.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const deb = suite("debounce");
+
+deb("setup and trigger debounce", () =>
+  createRoot(async dispose => {
+    let val = 0;
+    const fn = (current: number) => {
+      val = current;
+    };
+    const trigger = debounce(fn, 150);
```
atk, line 18: Please use smaller delays, so the tests don't take unneccessary long.

atk (approved):

packages/scheduled/src/index.ts
thetarnav: I think my reasoning was that whatever schedule method is used, will already clear itself. But does it mean that the clear() shouldn't be used instead? Probably not...

atk: Would clearing twice have any unwanted side effect? I guess not.

PR #137, opened 2022-05-22
Websocket: Improved composability

davedbase:
- [ ] Complete writing tests
- [ ] Validate reactivity setup and additional edge case needs are covered
- [ ] Adjusted README and package.json


davedbase:
@atk I really love the work you're doing on fetch and want to refactor Socket to have a similar structure. Some of the things Fetch has aren't relatable to Socket so I did my best to factor it. Before I continue with it I was wondering if you have an ideas or thoughts on this structure?

packages/websocket/src/socket.ts
```diff
@@ -0,0 +1,78 @@
...
+    socket.onmessage = onData;
+  };
+  const ctx = {
+    url,
+    disconnect,
+    get socket() {
+      return socket;
+    },
+    get disconnecting() {
+      if (disconnecting) disconnecting;
```
atk, line 70: Did you mean

```ts
return disconnecting || () => {};
```
?

packages/websocket/src/socket.ts
```diff
@@ -0,0 +1,78 @@
...
+      return () => {};
+    },
+    connect
+  } as SocketContext;
+  (modifiers || []).map((modifier) => modifier(ctx));
+  onCleanup(() => disconnect);
+  return [connect, disconnect, send, state, () => socket];
```
atk, line 77: I know it's a heavy breaking change, but I don't like this interface. 5 values in an array are too easy to confuse, also we have connect, disconnect and state, I would like to establish the pattern [read, write, actions]; read can be a signal with optional properties, write should be the main writing method with optional extra properties and actions should encompass everything that doesn't fit neither here not there.

In this case, I would propose the following return value:

```ts
type SocketReturnValue = <T>[
  data: () => T | undefined & { 
    state: 
      | WebSocket.OPEN
      | WebSocket.CONNECTING
      | WebSocket.CLOSING
      | WebSocket.CLOSED,
    socket: WebSocket,
  },
  send: (data: T) => void & {
    requestState(state: WebSocket.OPEN | WebSocket.CLOSED)
  }
];
```

davedbase:
@atk just looking back at this now and yeah sorry that you couldn't do a proper review. It was really messy. What I really wanted was an opinion on the API before I got further. I'll jump back on it now, thanks for the suggestions so far.

packages/websocket/src/socket.ts
davedbase: Hmmm mind talking me through the requestState part within send?

mur2zab: Also was just curious, why return an array ? where we have to make sure about the sequence in which the items are exported/returned. Can we simply return an object from it? Would that make it less error prone?
Suggestions? @davedbase @atk

atk: Using a tuple array is an established pattern in modern front end libraries, so there's nothing wrong with that, especially if you have TS/JSDoc.

PR #139, opened 2022-05-24
props: add `omitProps` and `pickProps` primitives

### LINKED ISSUE #104: add `omitProps` and `pickProps` primitives

thetarnav:
Closes #104 

[THE README](https://github.com/solidjs-community/solid-primitives/tree/pick-omit-props/packages/props#readme)


thetarnav:
Gonna close this for now as the current implementation of `omitProps` and `pickProps` doesn't account for dynamic props.
Will open this again once I get back to it and figure out what I want to do here.

PR #142, opened 2022-06-03
Bounds

thetarnav:
Want to add a primitive for getting the updated position and size of the element on the screen reactively.
Need to work on some dependency packages first.

TODO:
- [x] Refactor `resize-observer` package. [README](https://github.com/solidjs-community/solid-primitives/tree/bounds/packages/resize-observer#readme)
- [ ] Refactor `scroll` package
- [ ] Add `bounds` package

### DIFF
--- a/packages/resize-observer/README.md
+++ b/packages/resize-observer/README.md
@@ -9,7 +9,10 @@
 [![size](https://img.shields.io/npm/v/@solid-primitives/resize-observer?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/resize-observer)
 [![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-3.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
 
-Provides a reactive resize observer wrapper.
+Reactive primitives for observing resizing of HTML elements.
+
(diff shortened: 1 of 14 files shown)


packages/bounds/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  const updateBounds = () => setBounds(getElementBounds(access(target)));
+  const updateBoundsOf = (el: Element | undefined | null | false) =>
+    setBounds(getElementBounds(el));
+
+  if (typeof target === "function") {
+    onMount(() => updateBoundsOf(target()));
+    createComputed(on(target, updateBoundsOf, { defer: true }));
+  }
+
+  if (trackResize)
```
atk, line 125: By using a predefined options object, if the user tries `{ trackScroll: false }`, he'll be surprised that nothing will be tracked. It would be better to use an empty object by default and test whether `trackScroll` (or `scroll`, see earlier comment) was === false.

packages/bounds/src/server.ts
```diff
@@ -0,0 +1,13 @@
+import type * as API from "./index";
```
atk, line 1: Since you cannot possibly get an element, wouldn't it be better to use nulledBounds from before for the server? Or have an option for that?

atk (approved):
That's some nice work, Damian! I found nothing that I would request a change, I merely added some thoughts of mine.

packages/bounds/src/index.ts
thetarnav: I'm pretty sure it works properly...
The options are initialized as an empty object and destructured with some default values. So `{ trackScroll: false }` would only flip one switch.

PR #144, opened 2022-06-08
Upload: export types

ivteplo:
I made the `index.ts` export all the types of `@solid-primitives/upload` from `types.ts`.

I don't know whether all the types should be exported, or some should be left accessible only inside the package, so maybe some changes should be made


atk (approved):
Looks good to me.

PR #145, opened 2022-06-09
Refactor visibility-observer

thetarnav:
- Rename `createPageVisibilityObserver` to `createPageVisibility` _(no longer exported as default)_

- Add `usePageVisibility`

- Add tests

[README](https://github.com/solidjs-community/solid-primitives/tree/visibility-observer-refactor/packages/visibility-observer#readme)


davedbase:
Make sure you add yourself to the list of contributors or just change the author to your name :). Looking good, much simpler :)

thetarnav:
Ah thanks for reminding me :)
Will remove the dependency too, not worth it as you said.

davedbase:
Since we removed observer out of the name, maybe the package should be renamed to `page-visibility`?

thetarnav:
Yeah, it makes more sense. In interaction-observer pkg, there is an element visibility observer primitive too. `page-visibility` would be more clear.

PR #159, opened 2022-07-08
refactor Keyboard package

thetarnav:
- Refactors the keyboard package to heavily rely on shared roots, for reusing of signals/event-listeners/computations, and every other primitive would be only quering a set of data in signals provided by them.
- Adds a primitive for handling keyboard shortcuts.
- Bumps the version to `1.0.0`

[README](https://github.com/solidjs-community/solid-primitives/tree/keyboard/packages/keyboard#readme)

TODO:
- [ ] move from uvu to jest for now for testing — cannot properly test primitives that use a shared root, without the solid in dependencies not working properly.
- [ ] add tests for `createKeyHold` and `createShortcut`
- [x] update docs


atk:
The vitest problem is solved, you just need to inline all solid-using dependency packages (including internal if imported through node).

thetarnav:
Still can't get it to work...
This still fails:
```ts
import { isServer } from "@solid-primitives/utils";
test("...", () => {
  expect(isServer).toBe(false);
});
```

atk:
You need to inline `@solid-primitives/utils`, too, since it is otherwise imported by node.

thetarnav:
I think I'm doing it here if it's correct: https://github.com/solidjs-community/solid-primitives/blob/320d05f74e8894e45a6af6c6bcad419e287db6c1/packages/keyboard/vite.config.ts

atk:
I'll have a look later.

thetarnav:
Okay, I'm gonna merge it as it is.
@atk Hopefully someday this daunting problem will finally go away :D

PR #162, opened 2022-07-21
Props: add `filterProps` primitive

thetarnav:
Adds `filterProps` and `createPropsPredicate` primitives to the props package.

The idea of `filterProps` comes from the [`filterDOMProps`](https://github.com/solidjs-community/solid-aria/blob/main/packages/utils/src/filterDOMProps.ts#L58) helper in solid-aria, and will be used to update it.

[README](https://github.com/solidjs-community/solid-primitives/tree/filter-props/packages/props#filterprops)

### DIFF
--- a/packages/props/package.json
+++ b/packages/props/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@solid-primitives/props",
-  "version": "2.1.8",
+  "version": "2.2.0",
   "description": "Library of primitives focused around component props.",
   "author": "Alex Lohr <alex.lohr@logmein.com>",
   "contributors": [
@@ -17,6 +17,7 @@
     "stage": 3,
     "list": [
       "combineProps",
+      "filterProps",
       "createControlledProps"
     ],
     "category": "Utilities"
@@ -69,7 +70,8 @@
     "unocss": "^0.39.3",
     "uvu": "^0.5.6",
     "vite": "^2.9.14",
-    "vite-plugin-solid": "2.2.6"
+    "vite-plugin-solid": "2.2.6",
+    "solid-js": "^1.4.7"
   },
   "peerDependencies": {
     "solid-js": "^1.3.0"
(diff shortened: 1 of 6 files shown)


packages/props/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@solid-primitives/props",
-  "version": "2.1.8",
+  "version": "2.2.0",
   "description": "Library of primitives focused around component props.",
   "author": "Alex Lohr <alex.lohr@logmein.com>",
```
atk, line 5: Please add yourself to the author(s) field.

packages/props/package.json
```diff
@@ -17,6 +17,7 @@
     "stage": 3,
     "list": [
```
atk, line 18: Why not add the predicate helper?

packages/props/src/filterProps.ts
```diff
@@ -0,0 +1,75 @@
...
+ * ```
+ */
+export function filterProps<T>(props: T, predicate: (key: keyof T) => boolean): T {
+  return new Proxy(
+    {
+      get(property: string | number | symbol) {
+        return predicate(property as keyof T) ? props[property as keyof T] : undefined;
+      },
+      has(property: string | number | symbol) {
+        return predicate(property as keyof T) ? property in props : false;
```
atk, line 26: Maybe you don't want to needlessly run the predicate if the property is not even existing in the original props?

packages/props/src/filterProps.ts
```diff
@@ -0,0 +1,75 @@
...
+    (): Partial<Record<keyof T, boolean>> => {
+      // track prop names — adding/removing keys to dynamic props will trigger this
+      Object.keys(props);
+      return {};
+    },
+    undefined,
+    { equals: false }
+  );
+  return key => {
+    const _cache = cache();
```
atk, line 68: I would probably prefer the name `cacheRef` instead of `_cache` to better show the meaning.

atk (changes requested):

PR #163, opened 2022-07-24
chore(fetch): add defer stream to FetchOptions

heartistizio:
# Summary
Since v1.4.0 you can use deferStream so Solid waits for resource before flushing to stream, however this package doesn't support this with its types.


thetarnav:
@atk Should we add this before you finish #131?

thetarnav:
Alright I'll publish it then. Thanks @heartistizio for the fix :)

PR #177, opened 2022-08-12
stream: add mute method

atk:


packages/stream/src/index.ts
```diff
@@ -70,7 +79,15 @@ export const createStream = (streamSource: StreamSourceDescription): StreamRetur
   );
 
   onCleanup(() => stop(stream()));
-  return [stream, { mutate, refetch, stop: () => stop(stream()) }];
+  return [
+    stream,
+    {
+      mutate,
+      refetch,
+      mute: (muted?: boolean) => mute(stream(), muted),
```
thetarnav, line 87: Calling `mute()` will track `stream`. Is that desirable?
Same with `stop()` it seems.
The rest looks 👍

atk: In most cases, these will be called inside events, which are outside of the reactive scope. I hadn't considered other cases yet.

thetarnav: Sure, you can leave it if you think that's fine. But I would wrap it with `untrack` to avoid some weird edge-cases. I think it would be a hard issue to track down.

atk: You're right, and untrack is cheap. I'll add it once I get to it. Too tired today.

PR #180, opened 2022-08-14
feat(graphql): add support for `DocumentNode` and `TypedDocumentNode`

teenjuna:
This PR adds support `DocumentNode` and `TypedDocumentNode` types as arguments for the GraphQL query. With this change, it's now possible to use `graphql` primitive with the `graphql-code-generator`, which improves the TypeScript story for medium and large sized projects. 

This PR adds dependencies on the `DocumentNode` type from `graphql` package, `TypedDocumentNode` from `@graphql-typed-document-node/core` package (this type is just a thin wrapper around `DocumentNode` from `graphql`) and the `print` function from the `graphql` package (which is the only way to turn `DocumentNode` and `TypedDocumentNode` types into a string suitable for an API request). 

This PR also adds an example usage of such type inside the `dev` directory (adds a `pnpm gqlgen` command for generating the `TypedDocumentNode` types and uses generated type in one the two queries that already existed).


teenjuna:
I guess I didn't add changeset properly?

packages/graphql/src/index.ts
```diff
@@ -44,10 +46,10 @@ export const createGraphQLClient =
     headers?: RequestHeaders,
     fetchFn?: typeof fetch
   ): GraphQLClientQuery =>
-  (query, variables = {}, initialValue) =>
+  (query, variables, initialValue) =>
     createResource(
-      () => access(variables),
-      vars => {
+      () => access(variables == undefined ? {} : variables),
+      (vars: any) => {
```
thetarnav, line 52: Are changes here necessary?
here `variables == undefined ? {} : variables` will be executed every time createResource calls the `source` function, although the `variables` argument once provided, cannot change, so there is no point checking if it's undefined multiple times.

packages/graphql/src/index.ts
```diff
@@ -1,24 +1,26 @@
...
-  <T = unknown>(
-    query: string,
-    variables: MaybeAccessor<QueryVariables | boolean> | undefined,
+  <T = unknown, V = unknown>(
+    query: string | DocumentNode | TypedDocumentNode<T, V>,
+    variables: MaybeAccessor<V | boolean> | undefined,
     initialValue: T
   ): ResourceReturn<T, { initialValue: T }>;
-  <T = unknown>(
+  <T = unknown, V = unknown>(
```
thetarnav, line 21: The `V` generic should extend `{ [key: string]: string | number }`, right? Or at least an object.
```ts
V extends Record<string, string | number> = {}
```

packages/graphql/src/index.ts
```diff
@@ -1,24 +1,26 @@
...
-    variables?: QueryVariables;
+    variables?: V;
     fetcher?: typeof fetch;
   }
 >;
 
 export type GraphQLClientQuery = {
-  <T = unknown>(
-    query: string,
-    variables: MaybeAccessor<QueryVariables | boolean> | undefined,
```
thetarnav, line 16: I know that's not your doing, but allowing for `boolean` means that `true` could also be passed to bypass passing variables. So a `FalsyValue` from [utils](https://github.com/solidjs-community/soliprimitives/blob/main/packages/utils/src/types.ts#L116) should be a better fit.

thetarnav (approved):
Thank you for bringing this!
As you said, this improves the typescript support significantly.
Here are just some things I wanted to ask about:

packages/graphql/dev/gqlgen.ts
```diff
@@ -0,0 +1,146 @@
...
+export const Document = {"kind":"Document","definitions":[{"kind":"OperationDefinition","operation":"query","selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"countries"},"selectionSet":{"kind":"SelectionSet","selections":[{"kind":"Field","name":{"kind":"Name","value":"name"}},{"kind":"Field","name":{"kind":"Name","value":"code"}}]}}]}}]} as unknown as DocumentNode<Query, QueryVariables>;
```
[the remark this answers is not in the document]
teenjuna: That's weird, but I generated the file before I removed `QueryVariables` from `types.d.ts` file, so maybe this caused this weird behaviour. Need to check

packages/graphql/src/index.ts
teenjuna: You're right, I just tried fixing type error that was caused by `variables = {}`

packages/graphql/src/index.ts
teenjuna: The type of filed may be other than string or number. It can be a custom GraphQL scalar type

packages/graphql/src/index.ts
teenjuna: I'll check it

packages/graphql/dev/gqlgen.ts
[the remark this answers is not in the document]
teenjuna: This was caused by the fact that the `countries` query was unnamed

teenjuna:
@thetarnav I fixed most of the issues, except this one:
```ts
export const createGraphQLClient =
  (
    url: MaybeAccessor<string>,
    headers?: RequestHeaders,
    fetchFn?: typeof fetch
  ): GraphQLClientQuery =>
  (query, variables, initialValue) =>
    createResource(
      () => access(variables == undefined ? {} : variables),
      (vars: any) => {
        const variables = typeof vars === "boolean" ? {} : vars;
        return request(access(url), query, { headers, variables, fetcher: fetchFn });
      },
      { initialValue }
    );
```

I can't just write `variables = {}`, because `Type '{}' is not assignable to type 'MaybeAccessor<V | FalsyValue>'`. Not sure why, given that `V extends object = {}` in the signature of `GraphQLClientQuery`

teenjuna:
Also, is it really better to use `FalsyValue` instead of just `false`?
```ts
export type GraphQLClientQuery = {
  <T = unknown, V extends object = {}>(
    query: string | DocumentNode | TypedDocumentNode<T, V>,
    variables: MaybeAccessor<V | false> | undefined,
    initialValue: T
  ): ResourceReturn<T, { initialValue: T }>;
  <T = unknown, V extends object = {}>(
    query: string,
    variables?: MaybeAccessor<V | false>,
    initialValue?: undefined
  ): ResourceReturn<T, undefined>;
};
```

Also, it's not clear for me how we actually need to call `refetch` function if we init the query with `false` instead of variables:
```ts
const [countryData, { refetch }] = query(CountryQueryDocument, false, {
  country: { name: "loading..." }
});
refetch(/* info: unknown ???? */); // how do I pass variables?
```

thetarnav:
> I can't just write variables = {}, because Type '{}' is not assignable to type 'MaybeAccessor<V | FalsyValue>'.

All squares are rectangles but not every rectangle is a square ;)
You can just do `variables: any` in the args, and then `variables = variables ?? {}` before createResource

> is it really better to use `FalsyValue` instead of just `false`

createResource works the same way with the truthiness of the source arg—anything falsy will prevent fetch. So we are just reusing that behavior directly. The types should express that if they can.

> it's not clear for me how we actually need to call refetch function if we init the query with false instead of variables

The docs might might give a wrong example. Passing `false` will successfully prevent the fetcher from being executed at all. To properly use the falsy values, they have to be passed as a reactive function.
e.g.

```ts
const [queryEnabled, setQueryEnabled] = createSignal(false);

const [countryData, { refetch }] = query(
  CountryQueryDocument,
  () => queryEnabled() && { code: "BR" },
  { country: { name: "loading..." }}
);

setQueryEnabled(true) // will fetch the first time

refetch() // will refetch the second time
```

teenjuna:
I guess everything is done. Waiting for the @davedbase approval

PR #183, opened 2022-08-14
added webrtc demo for @solid-primitives/stream

mohanavel15:


atk:
Thank you for your contribution! I'll make a few minor changes before we'll add this.

packages/stream/demo/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  iceServers: [
+    {
+      urls: ['stun:stun1.l.google.com:19302', 'stun:stun2.l.google.com:19302'],
+    },
+  ],
+  iceCandidatePoolSize: 10,
+}
+
+const App: Component = () => {
+  const [stream, { mutate, stop }] = createStream({ video: true, audio: true })
```
atk, line 23: Maybe call it localStream to avoid confusion.

packages/stream/demo/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  async function StartCall() {
+    if (stream() === undefined) return alert("local Stream not ready yet")
+
+    if(PeerConnection.localDescription !== null) return
+    const offer = await PeerConnection.createOffer()
+    await PeerConnection.setLocalDescription(offer)
+  }
+
+  async function AnswerCall() {
+    if (stream() === undefined) return alert("local stream not ready yet")
```
atk, line 54: Wouldn't it be better to disable the button if the stream is not available instead of blocking the UI with an alert?

atk:
Have a look at the stream-demo branch.

packages/stream/demo/index.tsx
```diff
@@ -0,0 +1,135 @@
+import { render } from "solid-js/web";
+import { Component, type JSX, createEffect, createSignal } from 'solid-js';
```
atk, line 2: Component is a type, too.

packages/stream/demo/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+    },
+  ],
+  iceCandidatePoolSize: 10,
+}
+
+const App: Component = (): JSX.Element => {
+  const microphones = createMicrophones();
+  const cameras = createCameras();
+
+  const microphone = createPermission("microphone")
```
atk, line 30: using `microphone` as a name for the permission seems confusing, same for camera.

packages/stream/demo/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const camera = createPermission("camera")
+
+  const constraints: MediaStreamConstraints = {}
+  if (microphones().length > 0) {
+    constraints.audio = { deviceId: microphones()[0].deviceId }
```
atk, line 35: This should probably be made reactive.

atk (changes requested):
I'll help with the changes later.

atk (changes requested):

packages/stream/demo/index.tsx
```diff
@@ -31,20 +31,24 @@ const App: Component = (): JSX.Element => {
...
       setContraints("video", { deviceId: cameras()[0].deviceId })
+      setFoundDevice(true)
     }
   })
   
-  const [localStream, { mutate, stop }] = createStream(constraints)
+  const [localStream, { mutate, stop }] = createStream(() => foundDevice() ? constraints : undefined)
   const [remoteStream, setRemoteStream] = createSignal<MediaStream>()
 
+  const [CallState, setCallState] = createSignal(0)
```
atk, line 51: Just to make this more readable, you can have some magic number constants:
```ts
const CALL_IDLE = 0;
const CALL_INITIATED = 1;
const CALL_ACTIVE = 2;
const [callState, setCallState] = createSignal<0 | 1 | 2>(CALL_IDLE);
setCallState(CALL_ACTIVE);
```

atk (approved):

atk (approved):

PR #184, opened 2022-08-14
fetch: remove side-effects

atk:
The side effect was impacting tree shaking. Thanks to the composable update, we can now easily remove the global polyfill.


thetarnav (approved):
👍

PR #190, opened 2022-08-20
Created `createAmplitudeFromStream`

mohanavel15:
Moved amplitude logic from `createAmplitudeStream` to `createAmplitudeFromStream` So it can be used with third party MediaStream. Also removed audio sample rate so it will defaults to device's preferred sample rate.

### DIFF
--- a/packages/stream/src/index.ts
+++ b/packages/stream/src/index.ts
@@ -121,8 +121,31 @@ export const createAmplitudeStream = (
     stop: () => void;
   }
 ] => {
-  const [amplitude, setAmplitude] = createSignal(0);
   const [stream, { mutate, refetch, stop }] = createStream(streamSource);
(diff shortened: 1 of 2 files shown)


packages/stream/src/index.ts
```diff
@@ -134,7 +168,7 @@ export const createAmplitudeStream = (
 
   let source: MediaStreamAudioSourceNode;
   createEffect(() => {
-    const currentStream = stream();
+    const currentStream = stream instanceof MediaStream ? stream : stream ? stream() : undefined;
```
atk, line 171: I'd rather test if typeof stream is function, makes it more legible.

mohanavel15: Made a new commit with that change.

atk (approved):

atk:
Before merging, please run `pnpm run changeset` for a minor update with a change log entry.

PR #198, opened 2022-09-03
Implemented `createAccelerometer` and `createGyroscope`

mohanavel15:


atk (approved):

packages/devices/src/index.ts
```diff
@@ -73,3 +74,53 @@ export const createCameras = () => {
...
+ * @returnValue { alpha: 0, beta: 0, gamma: 0 }
+ */
+export const createGyroscope = (interval: number = 100) => {
+  const [orientation, setOrientation] = createStore({ alpha: 0, beta: 0, gamma: 0 });
+  let throttled = false;
+
+  const orientationEvent = (e: DeviceOrientationEvent) => {
+    if (throttled) return;
+    throttled = true
+    setTimeout(() => { throttled = false }, interval)
```
thetarnav, line 114: Shouldn't throttling be something you implement in a composable way, outside of the primitive?
If throttling is an important feature here, then maybe the user should be passing a scheduling callback to control the behavior. You can see this as an example of this pattern: https://github.com/solidjs-community/solid-primitives/tree/main/packages/bounds#throttling-updates

thetarnav (commented):
Hey, thanks for another contribution :)
The implementations look good, I only have one question about the throttling API below.
Could you add yourself as a contributor to package.json and the primitives to the list there?

packages/devices/src/index.ts
atk: Using a simpler solution is better for performance in this case.

PR #203, opened 2022-09-09
refetch on expiry

atk:
+ fix failing tests

### DIFF
--- a/packages/fetch/README.md
+++ b/packages/fetch/README.md
@@ -76,6 +76,10 @@ withRefetchEvent({ on: keyof HTMLWindowEventMap[], filter: (...args, data, event
 // caches requests
 withCache({ cache?: Record<string, CacheEntry>, expires?: number | ((entry: CacheEntry) => boolean); })
 
+// refetch on cache expiry
+// (expiry control function requires polling; you can set the delay; 0 = raf; default is 100ms)
+withRefetchOnExpiry(pollDelayMs: number)
+
 // makes cache persistent in storage, defaults = [localStorage, 'fetch-cache']
 withCacheStorage(storage?: Storage, key?: string)
 ```
--- a/packages/fetch/src/cache.ts
+++ b/packages/fetch/src/cache.ts
@@ -50,6 +50,7 @@ export const withCache: RequestModifier =
   ) =>
   (requestContext: RequestContext<Result, FetcherArgs>) => {
     requestContext.cache = options.cache || requestContext.cache;
+    requestContext.expires = options.expires;
     const isExpired = (entry: CacheEntry) =>
       typeof options.expires === "number"
         ? entry.ts + options.expires < new Date().getTime()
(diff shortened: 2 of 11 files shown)


packages/fetch/src/cache.ts
```diff
@@ -89,6 +91,37 @@ export const withCache: RequestModifier =
...
+          const cached: CacheEntry | undefined = requestContext.cache[serializedRequest];
+          if (!cached) {
+            if (typeof requestContext.expires === 'number') {
+              setTimeout(() => requestContext.resource?.[1].refetch(), requestContext.expires + 10);
+            } else if (typeof requestContext.expires === 'function') {
+              const poll = () => {
+                if (requestContext.expires(cached)) {
+                  requestContext.resource?.[1].refetch();
+                } else if (pollDelayMs === 0) {
+                  requestAnimationFrame(poll);
```
thetarnav, line 111: Why raf and not setTimeout? Is it because it'll happen sooner?

packages/fetch/src/cache.ts
```diff
@@ -1,4 +1,5 @@
 import { DEV } from "solid-js";
+import { urlToHttpOptions } from "url";
```
thetarnav, line 2: Is it fine to import from "url" on the browser?

packages/fetch/README.md
```diff
@@ -76,6 +76,9 @@ withRefetchEvent({ on: keyof HTMLWindowEventMap[], filter: (...args, data, event
 // caches requests
 withCache({ cache?: Record<string, CacheEntry>, expires?: number | ((entry: CacheEntry) => boolean); })
 
+// refetch on cache expiry (expiry control function requires polling; you can set the delay; 0 = raf)
+withRefetchOnExpiry(pollDelayMs: number)
```
thetarnav, line 80:
```suggestion
// refetch on cache expiry (expiry control function requires polling; you can set the delay; 0 = raf; defaults to 100ms)
withRefetchOnExpiry(pollDelayMs: number = 100)
```

packages/fetch/src/cache.ts
```diff
@@ -89,6 +91,37 @@ export const withCache: RequestModifier =
...
+          const serializedRequest = serializeRequest(requestData);
+          const cached: CacheEntry | undefined = requestContext.cache[serializedRequest];
+          if (!cached) {
+            if (typeof requestContext.expires === 'number') {
+              setTimeout(() => requestContext.resource?.[1].refetch(), requestContext.expires + 10);
+            } else if (typeof requestContext.expires === 'function') {
+              const poll = () => {
+                if (requestContext.expires(cached)) {
+                  requestContext.resource?.[1].refetch();
+                } else if (pollDelayMs === 0) {
```
thetarnav, line 110: Would it make sense to pull those `pollDelayMs` above the fetcher? The pollDelayMs won't change so it would be more readable.

thetarnav (approved):
Looks lika a great addition :)

packages/fetch/src/cache.ts
atk: The minimal resolution is ~16ms anyways, but raf will have slightly better performance, because it will always trigger between the client render cycles.

packages/fetch/src/cache.ts
atk: Removed the stray auto-import.

packages/fetch/src/cache.ts
atk: Found a better compromise between code size and performance.

packages/fetch/README.md
atk: fixed.

PR #207, opened 2022-09-14
fetch: new aggregation modifier

atk:
In order to have infinite scrolling or similar solutions, you can use the new `withAggregation`modifier.


thetarnav (approved):
Alright this looks solid, you can merge anytime.
The only thing I'm thinking about is maybe a callback for a custom way to aggregate resources.
Like if you have a response structure like this:
```ts
type Response = {
   // field you'd want to overwrite
   tabId: string
   // field you'd want to concat
   posts: Post[]
}
```

thetarnav (changes requested):
Oh, actually, I think you forgot about updating the server.ts export file :)

atk:
> Oh, actually, I think you forgot about updating the server.ts export file :)

Fixed that.

thetarnav (approved):

PR #212, opened 2022-09-15
fetch: expose cache serializer

atk:


thetarnav (approved):
👍
missed this one :)
should I be just merging them in instead of just "approving"? Not sure if it has any value..

PR #216, opened 2022-09-25
permission: fix in firefox

atk:


packages/permission/src/index.ts
```diff
@@ -4,19 +4,43 @@ import { Accessor, createSignal, onCleanup } from "solid-js";
...
-      onCleanup(() => status.removeEventListener("change", listener));
-    });
+  if (typeof navigator !== "undefined") {
+    navigator.permissions
+      .query(typeof name === "string" ? { name: name as PermissionName } : name)
+      .then(status => {
+        setPermission(status.state);
+        const listener = () => setPermission(status.state);
+        status.addEventListener("change", listener);
+        onCleanup(() => status.removeEventListener("change", listener));
```
thetarnav, line 20: Can't really get it to work in the playground. I would expect that I should get a prompt asking for permissions immediately, no? (silence both on chrome and firefox)
So I'm just gonna trust you that it works :P
or better, @davedbase should review this, as he's using it on the solid site.

I can only point to that `onCleanup`, which is getting called asynchrounously—after the promise resolves—not getting attached to the owner.
And causing this:
image

atk: As I said on discord, you need to use

```js
navigator.mediaDevices.getUserMedia({ 'audio': true })
```

In order to request permissions to get a stream in Firefox.

I'll have a look at the cleanup later.

atk: I fixed the warnings and added a few buttons for easier testing.

PR #219, opened 2022-09-29
fetch: fix cache invalidation

atk:


thetarnav (approved):
👍 💯 😄 
there is no changeset, but imma add one based on the previous PR

PR #226, opened 2022-10-08
quickfix: config object is supposed to be optional

AyloSrd:
Configuration argument is supposed to be optional, however calling `createIdleTimer()` without any argument would result in error. This shall fix it.


thetarnav:
You're right, thank you!

PR #228, opened 2022-10-09
Updated the reconnectLimit to avoid reconnection on calling disconnect function

mur2zab:
packages/websocket/src/index: Updated the reconnectLimit  to NEGATIVE_INFINITY to avoid reconnection on calling disconnect function


davedbase:
Thanks for adding that. :-)

PR #242, opened 2022-10-30
Add `scheduleIdle` to `scheduled` package

thetarnav:
Adds `scheduleIdle` scheduler based on `window.requestIdleCallback()`.

### DIFF
--- a/packages/scheduled/README.md
+++ b/packages/scheduled/README.md
@@ -64,7 +64,7 @@ The throttled callback is called on **trailing** edge.
 
 The timeout will be automatically cleared on root dispose.
 
-> **Note:** `requestIdleCallback` is not available in all browsers. If it's not available, `scheduleIdle` will fallback to `throttle` with default timeout.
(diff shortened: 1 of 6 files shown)


packages/scheduled/README.md
```diff
@@ -55,6 +56,30 @@ trigger("Hello!");
...
 
+## `scheduleIdle`
+
+Creates a callback throttled using `window.requestIdleCallback()`. ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback))
+
+The throttled callback is called on **trailing** edge.
+
+The timeout will be automatically cleared on root dispose.
+
+> **Note:** `requestIdleCallback` is not available in all browsers. If it's not available, `scheduleIdle` will fallback to `throttle` with default timeout.
```
atk, line 67: A bit more documentation of the consequences would be nice. Is there supposed to be a substantial difference between the fallback and the ideal solution?

atk (approved):

packages/scheduled/README.md
thetarnav: hmm well, the difference is very subtle, but that probably depends on the use case. Usually, `requestIdleCallback` will happen later than a `setTimeout` with no ms set, batching more callbacks.
How do you think I should expand on that?

PR #259, opened 2022-11-23
Fix vite config path on new-package script

aquaductape:
Fix vite config path on new-package script. The current path is incorrect and causes dev script on created package to fail.


atk (approved):

thetarnav:
Thanks 🙏

PR #260, opened 2022-11-24
Updated the link to mozilla.org

Evertt:
They changed the header in the article, so the hash part wasn't correct anymore. Although I just noticed that if you open the link, it doesn't scroll to the header anyway, so maybe this change is useless.


thetarnav:
Thank you :)

PR #263, opened 2022-11-28
add new package: broadcast-channel

aquaductape:
# Add Package: broadcast-channel

Proposing a primitive to use [BroadcastChannel API](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API)

Screen Shot 2022-11-28 at 1 06 41 AM

### makeBroadcastChannel

```ts
const { postMessage } = makeBroadcastChannel("test_channel");

postMessage({ id: 2, message: "hi" });

// Another browsing context
const { onMessage } = makeBroadcastChannel("test_channel");

onMessage(({ data }) => {
  console.log(data); // { id: 2, message: "hi" }
});
```

You can use the same channel easily across different components in the same context

```ts
const Component_1 = () => {
  const { postMessage } = makeBroadcastChannel("river");

  const onClick = () => {
    postMessage("hi");
  };

  return <button onClick={onClick}>Send Message</button>;
};

const Component_2 = () => {
  const { onMessage } = makeBroadcastChannel("river");
  const [message, setMessage] = createSignal("");

  onMessage(({ data }) => {
    setMessage(data);
  });

  return <div>{message()}</div>;
};

const App = () => {
  const { onMessage } = makeBroadcastChannel("river");

  onMessage(({ data }) => {
    console.log(data);
  });

  return (
    <>
      <Component_1 />
      <Component_2 />
    </>
  );
};
```

### createBroadcastChannel

Access the reactive `message()` signal that updates when postMessage is fired from other contexts

```ts
const { postMessage } = createBroadcastChannel("test_channel");

postMessage({ id: 2, message: "hi" });

// Another browsing context
const { message } = createBroadcastChannel("test_channel");

createEffect(
  on(
    message,
    data => {
      console.log(data); // { id: 2, message: "hi" }
    },
    { defer: true }
  )
);
```

## Stackblitz Demo
https://stackblitz.com/edit/vitejs-vite-5xren3?file=src%2Fmain.tsx
Copy url in stackblitz container and open additional tabs to properly test
Screen Shot 2022-11-28 at 1 02 36 AM

##  Or run Demo Locally
Check out `add-package-broadcast-channel` branch and look at dev directory.
```
cd ./packages/broadcast-channel
pnpm dev
```


atk:
This could be making much more use of the reactive system, i.e. have a "createBroadcastChannel(name)" that returns an accessor "message()" along with "sendMessage(msg)".

davedbase:
@aquaductape  this is awesome! Great idea. I agree with @atk next step are more reactive primitives. :-)

aquaductape:
@atk I added `createBroadcastChannel` primitive.

It returns signal `message()` where it's default value is `null`, which is why I'm using createEffect with `on` with defer `true`.

```ts
const { postMessage } = createBroadcastChannel("test_channel");

postMessage({ id: 2, message: "hi" });

// Another browsing context
const { message } = createBroadcastChannel("test_channel");

createEffect(
  on(
    message,
    data => {
      console.log(data); // { id: 2, message: "hi" }
    },
    { defer: true }
  )
);
```

aquaductape:
I can't think of anymore reactive primitives, broadcast channel is a pretty simple API

packages/broadcast-channel/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+  onMessage(({ data }) => {
+    setMessage(data);
+  });
+
+  return { message, postMessage: postMessage.bind(instance), channelName, instance, close };
+}
+
+// This ensures the `JSX` import won't fall victim to tree shaking before
+// TypesScript can use it
+export type E = JSX.Element;
```
thetarnav, line 133: This seems like a leftover from the template. From my understanding, it is needed only when extending JSX types to declare directives.

packages/broadcast-channel/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+    instance
+  } as TBroadcastChannelInstance["instance"];
+  newInstance.instance = result;
+  map[name] = newInstance;
+
+  return result;
+}
+
+export function createBroadcastChannel(name: string) {
+  const [message, setMessage] = createSignal<any>(null);
```
thetarnav, line 121: `unknown` might be a better option, as it requires actually checking the shape.
```suggestion
  const [message, setMessage] = createSignal<unknown>(null);
```

packages/broadcast-channel/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    const bcObjectInstance = currentInstance.instance.instance;
+    bcObjectInstance.close();
+    bcObjectInstance.removeEventListener("message", cbRef);
+    map[name] = null as unknown as any;
+  };
+
+  onCleanup(() => {
+    close();
+  });
```
thetarnav, line 83: No need to make another closure
```suggestion
  onCleanup(close);
```

packages/broadcast-channel/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "build": "jiti ../../scripts/build.ts --ssr",
+    "test": "vitest -c ../../configs/vitest.config.ts",
+    "test:ssr": "pnpm run test --mode ssr"
+  },
+  "peerDependencies": {
+    "solid-js": "^1.6.0"
+  },
+  "dependencies": {
+    "@solid-primitives/event-listener": "workspace:^2.2.4"
+  }
```
thetarnav, line 74: @solid-primitives/event-listener isn't actually used
```suggestion
  }
```

packages/broadcast-channel/src/index.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (process.env.SSR)
+    return {
+      onMessage: () => void 0,
+      postMessage: () => void 0,
+      close: () => void 0,
+      channelName: name,
+      instance: {} as unknown as BroadcastChannel
+    };
+
+  const id = createId();
```
thetarnav, line 35: I guess a symbol could be used in the same way, without the need to increment a counter.
```suggestion
  const id = Symbol()
```

thetarnav (commented):
Thank you, this looks like a nice API
I still have to look more into it, but here are a couple of comments from looking through the pr

Also, a couple of random thoughts

What about type safety? I'm guessing that there could be an interface with available channels and messages that one can send. Not sure if this is a good idea though, considering serialization and that some other messages could always sneak in. But I'm always thinking about ways of making something typesafe, so... I'm down to help :)

Also, I think it would be nice to have a primitive for sharing and synchronizing a signal state between contexts.
Pretty much like `createBroadcastChannel`, but has the same interface as `createSignal`.

```ts
const [selectedId, setSelectedId] = createBroadcastSignal("name-of-channel", initialValue);
```

But it's those just random ideas, the pr is good as is I think :)

packages/broadcast-channel/src/index.ts
aquaductape: Should I remove it?

packages/broadcast-channel/src/index.ts
aquaductape: updated with [d14d67c](https://github.com/solidjs-community/solid-primitives/pull/263/commits/d14d67c27f872e22cfe4f99a78e747df31d3e283)

aquaductape:
> What about type safety? I'm guessing that there could be an interface with available channels and messages that one can send. Not sure if this is a good idea though, considering serialization and that some other messages could always sneak in. But I'm always thinking about ways of making something typesafe, so... I'm down to help :)

I could allow user to pass basic generic, where postMessage must match that generic and message data is that generic. Any better ideas/implementation feel free to change

aquaductape:
> Also, I think it would be nice to have a primitive for sharing and synchronizing a signal state between contexts.
Pretty much like createBroadcastChannel, but has the same interface as createSignal.
> ```js
> const [selectedId, setSelectedId] = createBroadcastSignal("name-of-channel", initialValue);
> ```

I don't understand this one 😅, can you provide an example?

ruler501:
An easy mistake I see here is passing different types as the generic parameter in different components. I'm not sure how that would work though since it implies there's a single source of truth for the message type.

aquaductape:
I imagined users would make Types tied to the channelname, and pair them correctly throughout the application.

thetarnav:
@ruler501 nothing is stopping you from wrapping the primitives with your own typesafe interface.
There is a way to hold a type contract by extending a global interface, but that could be added later.

PR #265, opened 2022-12-01
feat: web share.

creampnx-x:
### Support

+ firefox
    + mac os: can be enable, default false.
    + windows: can be enable, default false.
    + android: true, but not `files`.
+ chrome, edge
    + mac os: `false`.
    + windows: true, but not `files`.
    + android: true.
+ safari
    + mac os, ios: true.

details: 
+ https://caniuse.com/?search=navigator.canShare
+ https://caniuse.com/mdn-api_navigator_canshare_data_files_parameter

#### screen shot
Test on Mac os Safari. 
1. Initial Appearance
截屏2022-12-01 15 59 01

2. click share button.
截屏2022-12-01 15 53 17

3. after sharing, the info change: `successful sharing.`
截屏2022-12-01 15 54 04


davedbase:
This is a wonderful contribution. Thank you so much! I have a couple suggestions:

- [ ] I'd prefer not to put this in the `browser` package. I recommend consolidating it in the `share` package that exists so that createWebShare lives beside createSocialShare. Another option is a new package called `webshare`
- [ ] I would simplify the demo and use the signal from the primitive directly instead of a createEffect that updates other signals.
- [ ] Would it make sense to have a makeWebShare and compose it with createWebShare? or is this API simple enough that producing either types of sharing primitives wont be necessary?

creampnx-x:
Thank you very much for your review. I'm going to try this done this weekend.

creampnx-x:
## Change
+ I added it to the `web-share` package, because the `share` package is exported by default.
+ I refactored the `createWebShare` using `createEffect` and `from`, so we can use the store of primitives directly. Besides, the return of primitives is not a signal but a store which has `status` and `failing message` properties.
+ I added `makeWebShare` which hasn't reactivity and compose it with `createWebShare`.

## Screen shots
截屏2022-12-03 21 43 35截屏2022-12-03 21 44 02

packages/web-share/src/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        data,
+        () => {
+          share(data())
+            .then(() => setStatus("status", true))
+            .catch(e => {
+              console.error(e);
+              setStatus({ status: false, message: e.toString() });
+            });
+        },
+        { defer: deferInitial || true }
```
atk, line 82: `deferInitial` will make no difference here. What you want is `deferInitial !== false`.

atk (changes requested):

packages/web-share/src/index.ts
creampnx-x: Done.

atk (approved):

thetarnav (commented):
Thank you, this is a very nice addition.
Here are a couple of suggestions from looking through the code.

packages/web-share/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  deferInitial: boolean = false
+): shareStatus => {
+  const [status, setStatus] = createStore<shareStatus>({});
+
+  if (process.env.SSR) {
+    return status;
+  }
+
+  const share = makeWebShare();
+  if (data)
```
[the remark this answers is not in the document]
creampnx-x: I have changed it to just be required, because web share can not be actively paused.

thetarnav (approved):

PR #267, opened 2022-12-07
Demo pages

atk:


thetarnav (approved):
wish there was an easy way to run it locally. (just opening index.html files gives me a cors error)
but it's not important

atk:
you mean run demos locally? `npx vite serve [path]`.

PR #274, opened 2022-12-25
fix: specify createTween return type

jrhicks:
Type declaration file in build had incorrect type:

@solid-primitives/tween/dist/index.d.ts
```
declare function createTween<T extends number>(target: () => T, { ease, duration }: {
    ease?: ((t: T) => T) | undefined;
    duration?: number | undefined;
}): () => void;
```

### DIFF
--- a/.changeset/witty-buses-hug.md
+++ b/.changeset/witty-buses-hug.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/tween": patch
+---
+
+fix: specify createTween return type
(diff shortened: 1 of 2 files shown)


thetarnav:
Thanks for the PR
Shouldn't the return type be `() => T` though? I see that `void` is returned on the server, but that probably shouldn't be the case... would returning `target` directly on the server make sense, just to keep the return type as `T`?

atk (approved):

PR #276, opened 2022-12-31
feat: solid-primitives-parser

bigmistqke:
A primitive to tokenize your solid-components to enable custom parsing/custom renderers.

It exists out of two parts, `tokenize()` and `childrenTokens()`:
- `const token = tokenize(callback, meta)` 
  - takes two parameters: 
    - a callback-function that takes in props just like a regular Solid-component
    - an object that can contain any additional metadata
  - returns a function token
    - this function spoofs to be a JSX.Element so it can be used in JSX without type-errors
    - its props are `Object.assigned()` to this spoof-function, so they can be accessed from their parent
    - a symbol `$TOKEN` for validation
- `const tokens = childrenTokens(() => props.children)`
  - takes as parameter:
    - its children in a callback, similar to `children()`-api
  - returns all valid tokens as a memo
    - they are filtered by checking for the `$TOKEN`-symbol

I made a very simple example in the repo with a calculator, and played around with a [threejs-renderer](https://codesandbox.io/p/github/bigmistqke/solid-three-parser/draft/funny-rgb)

I originally started playing with these type of hacks for [this cms-experiment](https://github.com/bigmistqke/plasticinecms) and @thetarnav mentioned that they also used something similar for `solid-aria` so it would be sweet to have this standardized. I think I will try to make a simplified version of the cms since it is quite a different usecase then the threejs-parser (callback() does return JSX which is being rendered, and data is inject from the parser).

The `Object.assign()`-tip came from someone in the typescript-discord, either liquid or foolswisdom. The remark they gave back then was that it's impossible to type it in a way to disallow these spoofed JSX-elements to be used outside the Parser, this can only be done in runtime. 

talking points:
-  Right now the `token.callback` does not take any parameters, but instead the props of the component are being passed into the callback. you can inject or manipulate data with `Dynamic` in the case of when you would use the callback to render, and I think you might even be able to set them directly on the token-function with another Object.assign, but haven't tested this. Another option possible would be to have the callback-function inside token to take besides the component-props also an object coming from the parent-parser: so something like `tokenize((props: {...}, parserParams: {...}) => {...})`.
- maybe instead of having tokenize pass in the component-props when you call `callback()` it should be done explicitly `callback(token.props, {... parserParams ...})` to demistify the inner-workings.
- maybe instead of the props being inferred from the callback-function it could be done by a generic: `tokenize<Props>(() => {...}, {})`. that way the callback is completely decoupled from its props, which feels a bit less spoofy and then you could type callback independently from the component's props. I imagine that in most of the usecase you do want those props though, so it will be a lot of `token.callback(token.props)`, but that's maybe better then doing it automagically. 
- mix and matching multiple different parsers could get buggy since they are checking for the same symbol. 

TODOS: 
- [ ] the whole type-story can use some assistance, I am still a ts-noob
- [ ] tests
- [x] docs

Also very open for API-suggestions.
I think this can become a very powerful tool in the solid toolchain, if done correctly. Since there is some spoofing going on to trick typescript to accept the JSX, the API-choices feel extra important.


packages/parser/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "url": "git+https://github.com/solidjs-community/solid-primitives.git"
+  },
+  "bugs": {
+    "url": "https://github.com/solidjs-community/solid-primitives/issues"
+  },
+  "primitive": {
+    "name": "parser",
+    "stage": 0,
+    "list": [
+      "createPagination"
```
atk, line 20: Please add the exported items to the list.

bigmistqke: done!

bigmistqke:
I continued my experimentations and added a cms-implementation as mentioned above: [it can be found here](https://codesandbox.io/p/github/bigmistqke/solid-three-parser/master?file=%2Fsrc%2FCMSParser.tsx&workspace=%257B%2522activeFileId%2522%253A%2522clcgs0y04000h7pg55mqpd8zm%2522%252C%2522openFiles%2522%253A%255B%2522%252FREADME.md%2522%255D%252C%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522gitSidebarPanel%2522%253A%2522COMMIT%2522%252C%2522spaces%2522%253A%257B%2522clcgs11hs00152e680q6vcvrf%2522%253A%257B%2522key%2522%253A%2522clcgs11hs00152e680q6vcvrf%2522%252C%2522name%2522%253A%2522Default%2522%252C%2522devtools%2522%253A%255B%257B%2522key%2522%253A%2522clcgs11hs00162e68hvq8lovw%2522%252C%2522type%2522%253A%2522PROJECT_SETUP%2522%252C%2522isMinimized%2522%253Atrue%257D%252C%257B%2522type%2522%253A%2522PREVIEW%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522port%2522%253A3000%252C%2522key%2522%253A%2522clcgs165t006y2e685oh9xzoh%2522%252C%2522isMinimized%2522%253Afalse%257D%252C%257B%2522type%2522%253A%2522TASK_LOG%2522%252C%2522taskId%2522%253A%2522dev%2522%252C%2522key%2522%253A%2522clcgs13t800562e68rc3z4zix%2522%252C%2522isMinimized%2522%253Atrue%257D%255D%257D%257D%252C%2522currentSpace%2522%253A%2522clcgs11hs00152e680q6vcvrf%2522%252C%2522spacesOrder%2522%253A%255B%2522clcgs11hs00152e680q6vcvrf%2522%255D%257D) 
A schema written in JSX generates a cms to enter data and typescript-types describing the schema. These types could be consumed hypothetically by a front-end who would fetch from the cms.

In these experiments I decided to remove the first `callback`-parameter from tokenize and only remain the second parameter, the object I called `meta` (naming suggestions welcome).

so 
```ts
tokenize((props: Props, parserParams: ParserParams) =>  ..., { ...meta... })
```
becomes
```ts
tokenize({ ...meta... })
```
instead of inferring the types from the callback-parameters they are now typed in a generic. It takes a type of type Token that looks like `Meta & {props: Props}`. Originally I wanted to keep `meta` and `props` in two different keys, like `{meta: Meta, props: Props}`, but this turned out to make typeguarding impossible

to illustrate 
```ts
type Green = Token<{id: 'green-id'}, {color: 'green'}> 
// {meta: {id: 'green-id'}, props: {color: 'green'}};
type Red = Token<{id: 'red-id'}, {color: 'red'}> 
// {meta: {id: 'red-id'}, props: {color: 'red'}};
type Color =  Green | Red;

const x: Color;
if(x.meta.id === 'green-id')
   x.props.color // Green | Red
```

while now
```ts
type Green = Token<{id: 'green-id'}, {color: 'green'}> 
  // {id: 'green-id', props: {color: 'green'}};
type Red = Token<{id: 'green-id'}, {color: 'green'}> 
  // {id: 'red-id', props: {color: 'red'}};
type Color =  Green | Red;

const x: Color;
if(x.id === 'green-id')
   x.props.color // Green
```

there is still some friction with types, for example:
```ts

type PropsGreen = {color: 'green'}
type Green = Token<PropsGreen, {callback: (props: PropsGreen) => void}> 
  // {callback: (props: PropsGreen) => void, props: {color: 'green'}}
type PropsRed = {color: 'red'}
type Red = Token<PropsRed, {callback: (props: PropsRed) => void}> 
  // {callback: (props: PropsRed) => void, props: {color: 'red'}}
type Color =  Green | Red;

const x: Color;
x.callback(x.props) 
  // typerror x.props: PropsGreen | PropsRed is not assignable to parameter of type PropsGreen & PropsRed 
```

I would have preferred to keep `meta` and `props` seperated, but this way we get a lot more type-information during parsing and there are far less typecasts needed.

At last I decided to scope $TOKEN, the symbol with which we identify a valid token, behind a createParser-closure. createParser returns `const {childrenTokens, tokenize}`: `childrenTokens` will only return children which are created by that specific `tokenize`. These could be exported by library-maintainers to extent the parser.

I updated this repo accordingly.

packages/parser/README.md
```diff
@@ -0,0 +1,27 @@
+<p>
+  <img width="100%" src="" alt="Solid Primitives parser">
+</p>
+
+# @solid-primitives/parser
```
thetarnav, line 5: `jsx-parser`?
Maybe `jsx` doesn't make sense if you put on the "well actually" glasses, but `parser` is wayyy too general

packages/parser/src/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  function tokenize<
+    TToken extends Meta & {
+      props: unknown;
+    }
+  >(meta: Omit<TToken, "props">): (props: TToken["props"]) => JSXElement {
+    return (props: any) => {
+      return Object.assign(
+        () => {
+          console.info("tokens can only be rendered inside a Parser");
```
thetarnav, line 14: warnings we usually put behind a dev check so that it's only included in the dev build
```suggestion
          process.env.DEV && console.info("tokens can only be rendered inside a Parser");
```

A `--dev` flag needs to be included in the build script to make it work though
-> `jiti ../../scripts/build.ts --dev`

bigmistqke:
API-idea

```
type CustomToken = {id: 'token', props: {type: string}}
const token = tokenize<CustomToken>({id: 'token'});
token // {id: 'token', props: {type: string}}
```
becomes
```
type CustomToken = {id: 'token', props: {type: string}}
const token = tokenize<CustomToken>((props) => ({id: 'token', props}));
token // {id: 'token', props: {type: string}}
```
?

so customer-props are being added to the token becomes an explicit action done by the dev, instead of something we do behind the scenes. then the dev themselves can also decide for themselves what to do with it.

packages/parser/src/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  function childrenTokens<T>(children: () => JSXElement | JSXElement[] | T | T[]) {
+    return createMemo(() =>
+      children
+        ? ([] as any[])
+            .concat(children())
+            .filter(child => child && $TOKEN in child)
+            .map((a: T) => a as T)
+        : []
+    );
+  }
```
thetarnav, line 35: This probably should use the `children` helper from solid to resolve children recursively from functions/arrays
```suggestion
function childrenTokens<T>(fn: () => JSXElement | JSXElement[] | T | T[]) {
  const resolved = children(fn)
  return createMemo(() => resolved.asArray().filter(child => child && $TOKEN in child))
}
```

packages/parser/src/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+        {
+          [$TOKEN]: true,
+          ...meta,
+          props
+        }
+      );
+    };
+  }
+
+  function childrenTokens<T>(children: () => JSXElement | JSXElement[] | T | T[]) {
```
thetarnav, line 26: Not sure about the generic here
I guess I can help with typings once it's finished
```suggestion
  function childrenTokens<T = {}>(children: () => JSX.Element): T[] {
```

packages/parser/src/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    TToken extends Meta & {
+      props: unknown;
+    }
+  >(meta: Omit<TToken, "props">): (props: TToken["props"]) => JSXElement {
+    return (props: any) => {
+      return Object.assign(
+        () => {
+          console.info("tokens can only be rendered inside a Parser");
+          return <></>;
+        },
```
thetarnav, line 16: This could be the default rendering method (just warn), but it could probably be customizable. You might have a component that is valid to use inside and outside the parser.
Although I would mostly parser-only so I like the default

packages/parser/src/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  function tokenize<
+    TToken extends Meta & {
+      props: unknown;
+    }
+  >(meta: Omit<TToken, "props">): (props: TToken["props"]) => JSXElement {
+    return (props: any) => {
+      return Object.assign(
+        () => {
+          console.info("tokens can only be rendered inside a Parser");
+          return <></>;
```
thetarnav, line 15: Can just return an empty string. No need to make it a tsx file and compile it.
```suggestion
          return '';
```

thetarnav (commented):
Looks good.
Here are a couple of things that I bought
Need to looks through this one more time later

packages/parser/src/index.tsx
thetarnav: On second thought the `children` helper would also resolve tokenized functions...
I guess this needs to resolve children manually to call normal functions and arrays, but leave the tokens.

bigmistqke: this is what I initially tried as well, but then it would try to render it out and we get `tokens can only be rendered inside a Parser`

packages/parser/src/index.tsx
bigmistqke: I like that! An optional second parameter in tokenize ===  the component used when u try to render the token directly?

packages/parser/src/index.tsx
bigmistqke: this is what I initially tried as well, but then it would try to render it out and we get `tokens can only be rendered inside a Parser`

packages/parser/README.md
bigmistqke: good by me. 

changed 
`parser` to `jsx-parser`, 
`createParser` to `createJSXParser` and
`tokenize` to `createToken`

packages/parser/src/index.tsx
bigmistqke: a weird, ur comment didn't show up before. yes exactly!

packages/parser/src/index.tsx
bigmistqke: it is a bit awkward having to pass in T like that, especially because that also means u have to type the children like that as well from the component you want to do `childrenTokens`, but without it I can not get `childrenTokens` to return the proper type.

```ts
type Token = {id: 'token'}
const Parser = (props: { children: JSXElement | JSXElement[] }) => {
  const tokens = childrenTokens(() => props.children); 
  tokens[0].id  // type-error: token is of type unknown
}

const Parser = (props: { children: JSXElement | JSXElement[] | Token | Token[] }) => {
  const tokens = childrenTokens(() => props.children); 
  tokens[0].id  // no type-error
}
```

maybe a type-helper could help here?

```
type Children<TToken> = JSXElement | JSXElement[] | TToken | TToken[]
const Parser = (props: { children: Children<Token> }) => {
  const tokens = childrenTokens(() => props.children); 
  tokens[0].id  // no type-error
}
```

bigmistqke: O you mean probably that you would type every `childrenToken`?

like
```
const Parser = (props: { children: JSXElement | JSXElement[] }) => {
  const tokens = childrenTokens<Token>(() => props.children); 
  tokens[0].id  // no type-error
}
```
that would work, is a bit verbose maybe.

ye, that casting of the token is a bit tricky actually mm...

bigmistqke: maybe instead of typing childrenTokens and hoping that the dev enters the correct generic, we could `createJSXParser` accept a generic of `CustomToken` which is a union of all the possible types, and that's what is being returned.

```ts
type Props = {brightness: number}
type GreenToken = {id: 'green', props: Props}
type RedToken = {id: 'red', props: Props}

type CustomToken  = GreenToken | RedToken

const {createToken, childrenTokens} = createParser<CustomToken>('color-parser')

const Green = createToken<Props, GreenToken>(props => ({props, id: 'green'})
const Red = createToken<Props, RedToken>(props => ({props, id: 'red'})

const Colors = (props: {children: JSXElement | JSXElement[]) => {
   const tokens = childrenTokens(() => props.children)
   if(token.id === 'red')
       token // RedToken
}

```

bigmistqke: yes! i think this is the solution. we should then also be possible to type the returntype of the callback in `childrenTokens(callback)` so that it fits the generic of `createJSXParser`.

bigmistqke: oo, that also means that even if you don't provide a Token-generic in `createToken`, it will still typecheck if what you return from the callback fits any of the accepted types.

```
type Props = {brightness: number}
type GreenToken = {id: 'green', props: Props}
type RedToken = {id: 'red', props: Props}

type CustomToken  = GreenToken | RedToken

const {createToken, childrenTokens} = createJSXParser<CustomToken>('color-parser')

const Yellow = createToken(props => ({props, id: 'yellow'}) 
  // type-error: Type 'yellow' is not assignable to type 'green' | 'red'
```

packages/jsx-parser/src/index.ts
```diff
@@ -1,15 +1,12 @@
...
 
-export function createJSXParser(id: string = "solid-parser") {
+export function createJSXParser<TTokens = {}>(id: string = "solid-parser") {
   const $TOKEN = Symbol(id);
 
-  function createToken<
-    TProps extends { [key: string]: any },
-    TToken extends { [key: string]: any } & { id: string }
-  >(
+  function createToken<TProps extends { [key: string]: any }, TToken extends TTokens>(
```
bigmistqke, line 6: is there a way you can make `TToken` fit the union `TTokens` without `extends`?

``` ts
type GreenToken = {id: 'green'}

const {createToken, childrenTokens} = createJSXParser<GreenToken>('color-parser')

const Green = createToken<Props>(props => ({id: 'green', extra: 'key'}) 
  // no type-error, since {props, id: 'green', extra: 'key'} extends {id: 'green'}

```

packages/parser/src/index.tsx
bigmistqke: [can we do it without extends though?](https://github.com/solidjs-community/solid-primitives/pull/276/commits/e35fac886b7242885769b1a70af0e1b9b145c322#r1061866399)

packages/jsx-parser/src/index.ts
```diff
@@ -30,10 +30,7 @@ export function createJSXParser(id: string = "solid-parser") {
   function childrenTokens<T>(children: () => JSXElement | JSXElement[] | T | T[]) {
     return createMemo(() =>
       children
-        ? ([] as any[])
-            .concat(children())
-            .filter(child => child && $TOKEN in child)
-            .map((a: T) => a as T)
```
bigmistqke: I was thinking that we could hide the 'attaching properties to a function'-implementation if we wanted:

``` ts
.map((a: T) => ({...a}))
```

then the result of `childrenTokens` would be an array of objects instead of an array of `createToken`.
but I am not sure if it's even a good thing to hide the implementation.

bigmistqke:
Until now I have only been testing with static schemas, except for For-loops, in the CMS-example.
I was trying now to make `Show` work as well, but I am having some issues.

Show returns a signal, to resolve these I added
``` ts
const resolveChild = (child: any) => {
  while (true) {
    if (typeof child !== 'function') return child
    if ($TOKEN in child) return child
    child = child()
  }
}
```
And added a memo to the callback of `childrenTokens`, like `children` does too (otherwise it would instantiate all the children everytime children would change).

But now when implementing it in the ThreeParser, we get the following bug:

https://user-images.githubusercontent.com/10504064/210671304-f8c81c31-3044-4254-8542-3a4dbe1eddde.mp4

once a new element is shown, we lose reactivity to the props of the original element.

[link to the codesandbox](https://codesandbox.io/p/github/bigmistqke/solid-three-parser/master?file=%2Fsrc%2FCMSApp.tsx)

packages/jsx-parser/src/index.ts
bigmistqke: It would also make new references each time. So bad idea.

bigmistqke:
I just realized you can also access the token-properties immediately from any callbacks by simply accessing `this`

``` ts
createToken((props: Props) => {
  props,
  id: 'token',
  callback: function(){
     console.log(this.id);      // 'token'
     console.log(this.props);   // Props
  }
})
```

that's pretty clean. 
in 99.9999% that will be enough, so you don't have to type all of the callbacks.

bigmistqke:
played around with another parser: a poor man's flutter

https://user-images.githubusercontent.com/10504064/210903093-3ec21d65-8565-420b-82ab-dfe94def9172.mov

jsx to drive a canvas-layout, with konva as canvas-paint-library and yoga as flexbox-layout-engine.

another pattern I discovered was to use the callback to initialize some values:
``` ts
export const Flex = createToken<PropsFlex, TokenFlex>((props) => {
  const tokens = childrenTokens(() => props.children)
  const node = Yoga.Node.create()
  node.setFlexDirection(props.style.flexDirection)
  return {
    props,
    node,
    id: 'Flex',
  }
}
```

very handy.
I think the API is now pretty much finalized as far as I am concerned, I am pretty happy with it. 
the `createRoot` everywhere is a bit awkward, but besides that it's pretty lean and clear.

will start to write docs. 
I am a test-noob, so maybe if someone else could look into this, or could give me some assistance.

bigmistqke:
I have been looking deeper into why `createEffect`s stop running once another token is shown. 
I first thought it was because of losing reference to its root, but that was not the case.
I went a route as well where I put the parser in one root and distributed them with the tokens and did `runWithOwner(owner, () => ...the callback code...)`, but to no avail.

 The callback-function of the first Mesh is being disposed once another Mesh is shown, it runs when I do `onCleanup` inside. I don't really understand why.

I was checking for `onCleanup` in the component and it was being disposed too, so I realized the token was being called everytime. I added a cache-mechanism so that the token is only initialized once. This stopped the disposal of the component, but this did not help with the issue of losing reactivity (but is probably good we do it anyways).

found another pattern besides wrapping all `createEffect`s in a `createRoot` so reactivity can be remained: by wrapping the callback in a `createMemo`

```
export const Mesh = createToken<PropsMesh, TokenMesh>((props) => ({
  props,
  id: 'Mesh',
  callback: () => {
     createRoot(() = > createEffect(() => ...))
  }
}))
```
becomes
```
export const Mesh = createToken<PropsMesh, TokenMesh>((props) => ({
  props,
  id: 'Mesh',
  callback: createMemo(() => {
     createEffect(() => ...);
  })
}))
```

this does mean you can not pass parameters into the callback, but since we can access the token with `this` it is still a powerful pattern.

if we could just prevent it from disposing in the first place, that would be wonderful...

packages/jsx-parser/src/index.ts
```diff
@@ -26,8 +26,9 @@ export function createJSXParser<TTokens = {}>(id: string = "solid-parser") {
 
   function childrenTokens(fn: Accessor<JSXElement | JSXElement[]>): Accessor<TTokens[]> {
     const children = createMemo(fn);
-    const resolveChild = (child: any) => {
+    const resolveChild = (child: any): any => {
       while (true) {
+        if (Array.isArray(child)) return child.map(resolveChild).flat();
```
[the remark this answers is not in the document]
thetarnav: Resolving children to a flat array is commonly done by every component that needs to render children under it's scope.
Not sure what would be the usecase.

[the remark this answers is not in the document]
bigmistqke: u mean the usecase of `{flattenArrays: false}` right?

1 possible usecase would be in a schema-visualizer: p.ex in the [cms-example](https://solid-parser-kg731ps1d-bigmistqke.vercel.app/) the left-panel showcases the schema that was being used to create the cms. This schema-visualization is generated by the parser [code](https://github.com/bigmistqke/solid-parser/blob/853dda3ed5ebd1e2793a709e7d524bdbf1d2c350/src/CMSParser.tsx#L190). 

If I would add `{flattenArrays: false}` to `childrenTokens` the schema could, a little bit more accurately be described when a `For` is being used inside the schema:

```jsx
<Parser>
  <For each={new Array(4)}>
    {() => <Token/>}
  </For>
</Parser>
```

without `{flattenArrays: false}` would become visualized as

``` jsx
<Parser>
  <Token/>
  <Token/>
  <Token/>
  <Token/>
</Parser>
```

with it you would still not know if the mapping-function was a `For` or `Index` or ..., but at least you would know it is coming from a template that is being mapped, so you could visualize the schema like: 

``` jsx
<Parser>
  <Map 4 times> // or some other syntax
    {() => <Token/>}
  </Map>
</Parser>
```

bigmistqke:
yes, looks great, @thetarnav !
very nice!

bigmistqke:
@thetarnav 
on second thought, maybe we should return `TokenComponent<Token>` instead of `JSX.Element` in `createToken`?

```
const jsxToken = <SomeToken staticProp="text" dynamicProp={signal()}/>
jsxToken // is always gonna be typed JSX.Element, is a jsx-thing i suppose
jsx.props // type-error
(jsx as any as SomeToken).props // no type-error, but ye

const token = SomeToken({staticProp: "text", get dynamicProp(){return signal()}})
token // TokenComponent<SomeToken>
token.props // no type-error

const Component = () => <Parser>{token}</Parser> // no type-error here either
```

sometimes it's handy to use a token outside jsx, writing it like a callback might make sense there.

thetarnav:
yeah I don't mind, we could do that

PR #278, opened 2023-01-07
fix: keyup event on useKeyDownList

Wysnard:
keyup event is missing on useKeyDownList hook event signal. It sends keydown instead of keyup on key release.

### DIFF
--- a/.changeset/ten-toes-fly.md
+++ b/.changeset/ten-toes-fly.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/keyboard": patch
+---
+
+fix keyup event on useKeyDownList hook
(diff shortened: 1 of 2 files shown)


thetarnav:
The `events()` signal is supposed to hold the last `"keydown"` event. At least based on the initial idea and existing documentation, so it's not a quick patch.
I think the reasoning was that it's the `keydown` events that you'd want to call `preventDefault` on. But I'm not exactly sure. What do you want to use that event for? It was mainly exposed for the other primitives in the library to build upon.

Wysnard:
I see. Since the signal was named `events()` I was expecting any kind of events but you are right the comment in the example of the README specify that it is only the last `keydown` event. I was just using the primitive hook in one of my side project and wanted to know when the user release its key but I can do that another way. I close the merge request.

thetarnav:
Alright, feel free to open another PR if you find anything else worth fixing :)

Issue #269, opened 2022-12-08
[keyboard] shortcuts trigger default actions second time


Ehesp:
Hey, 

When I setup the following shortcut (on Mac):

```
createShortcut(['Meta', 'P'], () => {
    console.log('Pressed Meta + P');
});
```

Pressing this combo works fine, however if I fully let go, and press it again the native browser "Print" dialog will popup. I close it, press the combo again and it works fine, press again and the dialog opens (and so on...).

I've tried with `requireReset` on both true and false but the behaviour is the same.

This behaviour is the same on "Meta + S" too.

thetarnav:
This is interesting.
Preventing the default behavior for <kbd>Control</kbd> + <kbd>P</kbd> on Windows works as expected, but for <kbd>Meta</kbd> + <kbd>P</kbd> the native behavior is executed anyway. Not sure if events associated with <kbd>Meta</kbd> key cannot be prevented or what's the issue. Needs more investigating.

Ehesp:
So for a bit more context, my current (slimmed) logic looks like this:

```ts
const isMac = navigator.platform.toLowerCase().indexOf('mac') > -1;

window.addEventListener('keydown', shortcut);

function shortcut(event: KeyboardEvent): void {
    if (isMac && e.metaKey && e.key === 'p') {
      event.preventDefault();
      // do something
    }
}
```

I can mash Cmd + P as much as I want and the print dialog doesn't show (this is in use here: https://zapp.run/edit/dart) - it'll trigger the command palette.  I implemented the logic above to trigger the palette but yeah, it worked but then showed the print dialog the second time, so it should be possible to prevent it.

sotasan:
Any progress so far?

davedbase:
Not sure why this wasn't patched sooner, but doing old maintenance and applied the patch from the abruptly closed PR for Primitives 1.0 and ported to 2.0 as well. A release is being main for main branch right now.

Thanks for your help everyone and sorry for the daily on this.

PR #279, opened 2023-01-08
fix keyboard useKeyDownList bug caused by Meta key

aquaductape:
When Meta(CMD or ⌘) key is pressed and held and then press different key, for example "Meta+K", the keyup for "K" does not fire, so the `pressedKeys` list doesn't get cleaned up properly causing issues. 

One issue would be if you use `createShortcut(["META", "K"])` to fire on "Meta+K", it will only work the first time. It won't work again until window blurs (document page loses focus) which clears `pressedKeys`.

You can see another issue where on [demo site](https://solidjs-community.github.io/solid-primitives/keyboard/), if you press "Meta+K", on KeyPress element the "K" persists after not pressing keyboard.

Screen Shot 2023-01-08 at 3 16 22 AM

This PR fixes those issues

### DIFF
--- a/packages/keyboard/src/index.ts
+++ b/packages/keyboard/src/index.ts
@@ -62,7 +62,11 @@ export const useKeyDownList = /*#__PURE__*/ createSharedRoot<
   makeEventListener(window, "keyup", e => {
     if (typeof e.key !== "string") return;
     const key = e.key.toUpperCase();
-    setPressedKeys(prev => prev.filter(_key => _key !== key));
+    if (key === "META") {
+      setPressedKeys([]);
+    } else {
+      setPressedKeys(prev => prev.filter(_key => _key !== key));
+    }
   });
 
   makeEventListener(window, "blur", reset);


thetarnav:
Is this possibly related to this issue? #269

aquaductape:
> Is this possibly related to this issue? #269

Yup, same issue

aquaductape:
so i read other peoples issues, such as overriding Mac print, "Meta+P".
```js
createShortcut(["META", "P"], () => {
  addMessage("Print pressed");
});
```
This PR fixes that issue too, callback is fired and Mac default print doesn't fire

thetarnav:
When the preventDefault option is enabled every default action should be prevented. Alternatively, you would disable that option and decide for yourself in the callback.

aquaductape:
Then my PR isn't ready, cuz it shouldn't prevent that default action by default

aquaductape:
@thetarnav when you use `createShortcut` on `Control + P` on Windows, does it prevent default action?

thetarnav:
Yup, but not for Meta + P

aquaductape:
> Yup, but not for Meta + P

Oh so `createShortcut` does prevent default action by default, since listening <kbd>Ctrl</kbd> + <kbd>P</kbd> on `createShortcut` on Windows doesn't fire print dialog

Nauxscript:
Is there any new progress? I am also experiencing this issue when using Meta as a shortcut key.

Nauxscript:
I tried this fix locally and it seems to be working. What is the reason that this fix was closed without being merged?
I really need it 😂...

kylooh:
same here, meta+ "any" shortcut will not work after first time triggering.

PR #281, opened 2023-01-14
Implemented `createInfiniteScroll`

mohanavel15:


packages/pagination/README.md
```diff
@@ -99,6 +99,21 @@ return (
...
+```tsx
+export default function App() {
+  // fetcher: (page: number) => Promise<T[]>
+  const [pages, loaderDirective] = createInfiniteScroll(fetcher);
+  return (
+    <div>
+      <For each={pages()}>
+        {item => <h4>{item}</h4>}
+      </For>
+      <h1 use:loaderDirective>Loading...</h1>
```
atk, line 112: You should probably use `<Show when={!ended}>` for the loader. Also, use: already shows that it is a directive, which is why I suggested `asLoader` as an ideal name for the directive.

packages/pagination/src/index.ts
```diff
@@ -215,3 +218,72 @@ export const createPagination = (
...
+ * @method `setPages` allows to manually change the contents of page
+ * @method `end` is a boolean indicator for end of page
+ * @method `setEnd` allows to manually change the end
+ *
+ */
+export function createInfiniteScroll<T>(
+  fetcher: (page: number) => Promise<T[]>,
+): [
+    pages: Accessor<T[]>,
+    loaderDirective: Setter<Element | undefined>,
```
atk, line 252: Please alias the type into LoaderDirective instead of using that as name:

```ts
type LoaderDirective = (ref: Element) => void;

return {
  …,
  asLoader: LoaderDirective,
  …
};
```

atk (changes requested):

packages/pagination/src/index.ts
mohanavel15: i misread the message. i will fix it.

packages/pagination/src/index.ts
```diff
@@ -229,27 +229,28 @@ declare module "solid-js" {
...
  * ```typescript
- * const [pages, loaderDirective, { page, setPage, setPages, end, setEnd }] = createInfiniteScroll(fetcher);
+ * const [pages, asLoader, { page, setPage, setPages, end, setEnd }] = createInfiniteScroll(fetcher);
  * ```
  * @param fetcher (page: number) => Promise<T[]>
  * @return `pages()` is an accessor contains array of contents 
  * @property `pages.loading` is a boolean indicator for the loading state
  * @property `pages.error` contains any error encountered
- * @return `loaderDirective` is an directive used to set loader
+ * @return `asLoader` is an directive used to set loader
```
atk, line 242: Typo `a directive`.

atk (approved):

packages/pagination/README.md
```diff
@@ -99,6 +99,23 @@ return (
...
+export default function App() {
+  // fetcher: (page: number) => Promise<T[]>
+  const [pages, asLoader, { end }] = createInfiniteScroll(fetcher);
+  return (
+    <div>
+      <For each={pages()}>
+        {item => <h4>{item}</h4>}
+      </For>
+      <Show when={!end()}>
+        <h1 use:asLoader>Loading...</h1>
```
thetarnav, line 113: In the defined types the name of the directive is `loaderDirective`, and here `asLoader` is used, which ts won't like.
```ts
interface Directives {
  loaderDirective: boolean
}
```

packages/pagination/src/index.ts
```diff
@@ -215,3 +218,73 @@ export const createPagination = (
...
+ * @method `setPage` allows to manually change the page number
+ * @method `setPages` allows to manually change the contents of the page
+ * @method `end` is a boolean indicator for end of the page
+ * @method `setEnd` allows to manually change the end
+ */
+export function createInfiniteScroll<T>(
+  fetcher: (page: number) => Promise<T[]>,
+): [
+    pages: Accessor<T[]>,
+    asLoader: loaderDirective,
```
thetarnav, line 253: You could just use the `Directive` type from `utils` package.
No need to make it into a separate type.
```suggestion
    asLoader: Directive,
```

packages/pagination/src/index.ts
```diff
@@ -215,3 +218,73 @@ export const createPagination = (
...
+      setPage(p => p + 1);
+    }
+  })
+
+  createEffect(() => {
+    const content = contents();
+    if (!content) return;
+    if (content.length === 0) setEnd(true);
+    setPages(p => [...p, ...content!]);
+  })
```
thetarnav, line 281: Could this be done with derivations instead of synchronizing the state with effects?
If the effect is the only way, then `createComputed` will be better.

thetarnav (changes requested):
Looks interesting and nice to have :)

Is using directives the only way to use it though? They should never be the default, and definitely not the only API option.
Although I guess that this would work too?
```tsx
const [pages, loader, { end }] = createInfiniteScroll(fetcher)
// ...
<h1 ref={loader}>Loading...</h1>
```

packages/pagination/README.md
mohanavel15: forgot to rename 😅

packages/pagination/src/index.ts
```diff
@@ -215,3 +217,75 @@ export const createPagination = (
...
+    end: Accessor<boolean>;
+    setEnd: Setter<boolean>;
+  }
+] {
+  const [loader, setLoader] = createSignal<HTMLElement>();
+  const [pages, setPages] = createSignal<T[]>([]);
+  const [page, setPage] = createSignal(0);
+  const [end, setEnd] = createSignal(false);
+
+  const visible = createVisibilityObserver()(loader);
```
thetarnav, line 264: Not sure if using `createVisibilityObserver` is necessary here. We don't care about having a `visible` signal, we just want to fire a callback when the element is in view — probably better to use `makeIntersectionObserver` to not needlessly create another signal. Unless I'm missing something.

thetarnav (changes requested):
It would be nice to have a demo of it working under /dev or some unit tests. Right now I can only guess if this is working from reading the code.

packages/pagination/src/index.ts
```diff
@@ -215,3 +217,75 @@ export const createPagination = (
...
+      setPage(p => p + 1);
+    }
+  });
+
+  createComputed(() => {
+    const content = contents();
+    if (!content) return;
+    if (content.length === 0) setEnd(true);
+    setPages(p => [...p, ...content!]);
+  });
```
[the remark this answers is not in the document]
atk: I would love to leave that as an option, to either have a fetcher, which returns a promise or something that returns a `[resource: Resource, actions: ResourceActions]` Array, as fetcher argument, which then would be handled correctly either way, to combine a maximum of composability with a simple API.

packages/pagination/src/index.ts
mohanavel15: yes callback when element is in view. i just checked the package and used `createVisibilityObserver` because it was very simple.

thetarnav (approved):
Alright, I'm fine with adding this how it is now as an initial version.
The implementation could be improved as mentioned in threads here: https://github.com/solidjs-community/solid-primitives/pull/281#pullrequestreview-1272953492 (the added demo is great though)
But it works fine, and even in SSR, it will trigger a suspense boundary (although uncontrollably) so it's more than usable, providing a nice starting point for implementing infinite scroll like this.
@mohanavel15 please add your name to [the contributor's field](https://github.com/solidjs-community/solid-primitives/blob/8b6f6192639757db3812b71d5d6106970e98394b/packages/pagination/package.json#L6) and I'll merge this :)

thetarnav:
@atk the Deploy task keeps failing here. should it even run on PRs? Because if the deployment from PRs overwrites the one from `main`, then I think it's better to have only `main` deployed.

atk:
It shouldn't run on any other branch than main. I think I had a guard, but it doesn't seem to work. I'll have a look later.

PR #284, opened 2023-01-15
[pagination] Update computed pages when options signal changes.

martinpengellyphillips:
If the passed pagination options are a reactive signal, ensure that the pages array is recomputed when it changes. This prevents the pages array getting out of sync and the resulting pagination props missing pages.

### DIFF
--- a/.changeset/gentle-tools-wonder.md
+++ b/.changeset/gentle-tools-wonder.md
@@ -1,0 +1,9 @@
+---
+"@solid-primitives/pagination": patch
+---
+
+Update computed pages when options signal changes.
+
+If the passed pagination options are a reactive signal, ensure that the pages
+array is recomputed when it changes. This prevents the pages array getting out
+of sync and the resulting pagination props missing pages.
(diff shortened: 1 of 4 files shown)


atk:
This is something I was afraid to add, because then, we re-render all pages just because their count changed, which I wanted to avoid.

However, we can create a solution that reuses previous page props items by leveraging that a createMemo callback receives the previous value as an argument.

martinpengellyphillips:
Cool, I'll update this to reuse props by page index in the memo.

martinpengellyphillips:
Related: what are your thoughts on clamping the page signal value as well? Currently it is possible for page to be greater than pages after options update. Should we react to that (either a change in initialPage for controlled, or a detection of page > pages)?

atk:
> Should we react to that (either a change in initialPage for controlled, or a detection of page > pages)?

Actually, I haven't thought about that. If we select a page for the user after an options update, it will certainly be against the original (now impossible to fulfil) intentions of the user, so the best we can hope for is second best and move to the last page.

I conciously choose initialPage as a name - it's not controlled through that option, but solely used as an initial value. I think we should cap the value on both sides.

martinpengellyphillips:
Re: "I think we should cap the value on both sides." - I'll look at working up a pr for this and see how it feels. Are you thinking that `page` becomes derived or that we internally `setPage` when `pages` updated?

Meanwhile, for this pr I have updated to reuse pages across changes in the memo 😄

atk:
Closed with PR#314.

PR #285, opened 2023-01-15
[pagination] Prevent wrong pages slice

martinpengellyphillips:
When `maxPages` is greater than `pages`, ensure the min value of the computed start for slicing is 0 (rather than a negative number). A negative number starts indexing from the end of the array causing an incorrect slice to be made. 

E.g. `10 pages - 13 maxPages = pages[-3, 13] = pages[8,9,10]`

### DIFF
--- a/.changeset/calm-islands-serve.md
+++ b/.changeset/calm-islands-serve.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/pagination": patch
+---
+
+Prevent wrong pages slice by clamping start to 0.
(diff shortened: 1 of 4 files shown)


atk:
maxPages should in its usage not ever be more than pages. It would be better to derive it:

```ts
const maxPages = () => Math.min(opts().maxPages, opts().pages);
```

And use that throughout the calculations (including the slice of pages. So the start calculation would not get any more convoluted (I was pretty close to add a comment or extract variables to explain the meaning of both parts).

martinpengellyphillips:
Makes sense, I'll do that.

martinpengellyphillips:
@atk ready for another look.

atk:
As an aside, thanks for testing this more thoroughly. 👍

atk:
Closed with PR#314.

PR #289, opened 2023-01-23
feat: rootless/createRootlessEffect

bigmistqke:
An effect that runs with the encompassing owner instead of creating an owner each time the effect is run.
Was very useful when playing around with [createJSXParser](https://github.com/solidjs-community/solid-primitives/pull/276).


bigmistqke:
(it's a bit unclear to me what a changeset is, I can add it in case it is needed)

thetarnav:
changeset is mainly for maintainers so don't worry

your implementation seems to be exploiting a current bug with `runWithOwner` ([solid issue](https://github.com/solidjs/solid/issues/1495)) so I cannot really accept it, since it'll likely be patched.

bigmistqke:
🤔 mm could you elaborate? I don't understand how the issue relates to this PR.

thetarnav:
For the "createRootlessEffect" (which doesn't do anything with roots so I don't get the name) to "behave like normal effect" - be reactive, it has to track - assign itself as current global Listener when running the callback. You are wrapping the callback with `runWithOwner` so that the global Owner is reset to the parent one when running the callback, and because `runWithOwner`  has a bug where it doesn't change the current Listener your effect works this way when any accessed signal is subscribed by the effect, but any computation created inside is owned by effect's parent, instead of the effect.
When `runWithOwner` will be patched, wrapping the callback with it will cause it to not be reactive, as if your use `untrack`. Or `onMount` actually

thetarnav:
I don't get what is the main goal of this primitive actually. I get technically how it works, but now why.

bigmistqke:
> which doesn't do anything with roots so I don't get the name
I am a bit unclear about the difference between owners/roots tbh, so it could definitely be a misnomer.

> I don't get what is the main goal of this primitive actually. I get technically how it works, but not why.
It came in handy when playing around with `jsx-parser`, specifically to deal with [this issue](https://github.com/solidjs-community/solid-primitives/pull/276#issuecomment-1371555565) which I solved [here](https://github.com/solidjs-community/solid-primitives/pull/276#issuecomment-1373914172).

The tldr (lol went a bit overboard with sharing the process there): I was tracking children/tokens being added and removed inside an effect. Everytime the effect was re-run and a new child was being added, I would lose the reactivity to the already instantiated tokens, since those tokens were being disposed because each time an effect is re-run a new owner is being created. With `createRootlessEffect` the owner of the callback is the owner encompassing the effect instead, being the parent-token, so the effect's callback is only being disposed when the encompassing element is being disposed.

bigmistqke:
It allowed me to do something like
```tsx
function mapTokens<T>(tokens: Accessor<T[]>, callback: (token: T) => any) {
  const cache = new Map<T, () => void>()

  createRootlessEffect(() => {
    const current = new Set()
    tokens().forEach((token) => {
      if (!cache.has(token)) {
        createRoot((dispose) => {
          callback(token)
          cache.set(token, dispose)
        })
      }
      current.add(token)
    })

    cache.forEach((dispose, token) => {
      if (!current.has(token)) {
        cache.delete(token)
        dispose()
      }
    })
  })
}
```
so I could do
```tsx
const childrenEffect = (
  object: THREE.Object3D,
  tokens: Accessor<Token[]>
) => {
  mapTokens(tokens, (token) => {
    const element = token.callback(token.props as any)
    object.add(element)
    onCleanup(() => object.remove(element))
  })
}
```
to create and dispose the elements.

thetarnav:
in cases like this, you just use `createRoot` and manually control the disposal
this is actually what `mapArray` and `indexArray` does since it has to solve the same problem to not dispose existing items when the input changes.
And I think I need to actually add a `onDispose` primitive that would help here a bit

bigmistqke:
Yeah you are right, lol. Now that I look back on it, `createRootlessEffect` isn't doing anything in `mapTokens`, since everything is already wrapped in `createRoot` anyway.

> this is actually what mapArray and indexArray does

I expected mapArray to work like mapTokens, but couldn't get it to work, probably should look better into it.

Okay, will close this PR, thanks for the back-and-forward!

bigmistqke:
> I expected mapArray to work like mapTokens, but couldn't get it to work

[ok, I understand mapArray now 👍](https://playground.solidjs.com/anonymous/2b263430-18d2-4fd1-b6a9-85fb6d254db4)

PR #290, opened 2023-01-23
Add new `onDispose` primitive

thetarnav:
Adds new `onDispose` primitive to the `rootless` package.

It's a variant of `onCleanup` that only runs if the owner actually gets disposed of - not just rerun.
Similar to the returned function of `useEffect` from React.
The implementation gets into solids internals to do it, but normally to do it from within an effect, you'd need access to parent owner to which then you could attach cleanup with `runWithOwner` and `onCleanup`. This way you don't need to pass down the owner.

```ts
import { onDispose } from "@solid-primitives/rootless";

createRoot(dispose => {
  const [count, setCount] = createSignal(0);

  createEffect(() => {
    count();

    onCleanup(() => {
      console.log("cleanup");
    });

    onDispose(() => {
      console.log("disposed");
    });
  });

  setCount(1); // cleanup

  dispose(); // cleanup + dispose (twice)
});
```

@bigmistqke this is what I've mentioned in https://github.com/solidjs-community/solid-primitives/pull/289

### DIFF
--- a/.changeset/pretty-games-vanish.md
+++ b/.changeset/pretty-games-vanish.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/rootless": minor
+---
+
+Add new `onDispose` primitive.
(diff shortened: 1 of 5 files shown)


thetarnav:
I don't like that it uses solid internals to achieve this behavior.
I'm going to shelve it until I find a way to do it using public APIs.

PR #291, opened 2023-01-23
Add package: `option`

eudrinoagain:
This adds a new utility package for Optional values, Similar to [Rust's Option](https://doc.rust-lang.org/stable/std/option/enum.Option.html)
not sure if it belongs as a primitive, keep finding myself missing Rust's APIs when using Solid.
```tsx
const [user, setUser] = createStore({
  submissionValue: None<string>()
});
return (
  <>
    <p>Wrote: {user.submissionValue.unwrap_or("Didnt write yet.")}</p>
    <input
      type="text"
      onInput={e => setUser("submissionValue", makeOption((e.target as any).value || null))}
    />
  </>
);
```


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "solid-primitives",
-  "private": true,
+  "private": false,
```
thetarnav, line 3: Please leave this `true`
It means that the root "package" is not meant to be published to npm if somebody types `npm publish` by mistake 😄

packages/option/dev/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [user, setUser] = createStore({
+    role: Some(1),
+    submissionValue: None<string>()
+  });
+  return (
+    <>
+      <p>Wrote: {user.submissionValue.unwrap_or("Didnt write yet.")}</p>
+      <input
+        type="text"
+        onInput={e => setUser("submissionValue", makeOption((e.target as any).value || null))}
```
thetarnav, line 17: In solid `currentTarget` has the current ref assigned to the event so it also has the correct types.
```suggestion
        onInput={e => setUser("submissionValue", makeOption(e.currentTarget.value || null))}
```

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+   * @example
+   * ```ts
+   * const count = None();
+   *
+   * const onClick () => {
+   *   doSomething(count.unwrap_or_else(() => 5));
+   * };
+   * ```
+   */
+  unwrap_or_else(closure: () => T) {
```
thetarnav, line 92: Not sure how I feel about the snake case. It's obviously easier to read, but camelCase is the standard for js, especially for public APIs in libraries. So `unwrapOrElse` would be the way I think.

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Checks whether the option is the Some variant.
+   * @example
+   * ```ts
+   * const userSubmission = Some(5);
+   *
+   * if (userSubmission.isSome()) return alert(`The user has already submitted value of ${userInput}`);
+   * ```
+   */
+  isSome() {
```
thetarnav, line 41: Not sure if I get the point of having these methods attached to the object.
Wouldn't it be better to export just a normal function for checking that, it would also offer better ts narrowing support.
```ts
function isSome<T>(v: T | null | undefined): v is T {
  return v != null
}

declare x: number | null

if (isSome(x)) {
  x // T: number
}
```

Functions also are better tree-shakeable.

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+ * @example
+ * ```ts
+ * const option = Some(5);
+ *
+ * if (option.isSome()) {
+ *   doSomething(option.unwrap())
+ * }
+ * ```
+ */
+class Option<T> {
```
thetarnav, line 15: this class should probably be exported

packages/option/package.json
```diff
@@ -0,0 +1,74 @@
...
+      "import": "./dist/server.js",
+      "require": "./dist/server.cjs"
+    },
+    "import": "./dist/index.js",
+    "require": "./dist/index.cjs"
+  },
+  "scripts": {
+    "dev": "vite serve dev",
+    "page": "vite build dev",
+    "build": "jiti ../../scripts/build.ts --ssr",
```
thetarnav, line 67: The builds are the same for client and server so no need to build them separately
```suggestion
    "build": "jiti ../../scripts/build.ts",
```

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export const Some = <T extends Required<T>>(value: T) => new Option(value);
+
+/**
+ * Creates a new Option instance with the None variant.
+ * @return A None value of T
+ * @example
+ * ```tsx
+ * const user = createStore({
+ *   role: None
```
thetarnav, line 175: it could be called I guess
```suggestion
 *   role: None()
```

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+ * @example
+ * ```tsx
+ * const user = createStore({
+ *   role: Some(1)
+ * });
+ * <div>{user.role.unwrap() || "Role is null"}</div>
+ * ```
+ */
+
+export const Some = <T extends Required<T>>(value: T) => new Option(value);
```
thetarnav, line 167: Not sure if these should be starting with capital letters, it's not a class nor a component. So it's better to make it `some` or `makeSome` or `newSome`, or just export the Options class and not bother with these additional helpers.
But not sure, it may be part of the appeal.

packages/option/src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+   * });
+   *
+   * // very unsafe, possibly null!!!
+   * const onClick = () => {
+   *   doSomething(count.unwrap());
+   * };
+   *
+   * ```
+   */
+  unwrap() {
```
thetarnav, line 62: I kinda wish that the way to get the value would just to call it. Like a signal, not a property on it.
```ts
const count = Some(5);

// get value
doSomething(count())

// check it beforehand
// (isSome could narrow down the type of the getter to () => NonNullable<T>)
isSome(count) && doSomething(count())
```

packages/option/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+A set of utilities to simplify using Optional values, very much inspired by [Rust's Option](https://doc.rust-lang.org/stable/std/option/enum.Option.html)
+
+`createOption` - Creates a reactive Option instance.
+
+`makeOption` - Creates a non-reactive Option instance.
+
+`Some` - Similar to `makeOption` but with a required initial value.
+
+`None` - Similar to `makeOption` but with a default None value.
```
thetarnav, line 20: the table of contents should be clickable
```suggestion
- [`createOption`](#createOption) - Creates a reactive Option instance.
- [`makeOption`](#makeOption) - Creates a non-reactive Option instance.
- [`Some`](#Some) - Similar to `makeOption` but with a required initial value.
- [`None`](#None) - Similar to `makeOption` but with a default None value.
```

thetarnav (commented):
Okay, this is looking promising. I like the idea and exploration although I think it needs to be refined a bit to offer better dx.

thetarnav (commented):
I've also asked in discord for some feedback about the idea and this is what I got:
1. this package isn't bringing anything solid specific, one could just use `fp-ts` package and get the same benefit. (which is probably better standardized)
2. Option should never return null values, that's the whole point
3. mixing of camelCase and snake_case is no good
4. unwrap should throw an error if value is None
5. These should be the values
    ```ts
    Option<T> = Some<T> | None
    Some<T> = () => T
    None = () => null
    ```
    Instead of this
    ```ts
    Option<T | null> = () => T | null
    Some<T> = Option<T>
    None = Option<null>
    ```
    Or maybe this
    ```ts
    type Option<T> = Some<T> | None;
    const None: unique symbol = Symbol("None");
    type None = typeof None;
    type Some<T> = () => T;
    ```

eudrinoagain:
I'm gonna close this PR as i changed my mind about the idea, i think the primitive doesn't belong here for now.

PR #293, opened 2023-01-28
script-loader: Improve API, spread any props on to the script

thetarnav:
Add `onBeforeAppend` callback option. Rename onload and onerror to onLoad and onError. Do not return the remove script function - should be done with disposing the owner. React to src changes with createRenderEffect.

### DIFF
--- a/packages/script-loader/src/index.ts
+++ b/packages/script-loader/src/index.ts
@@ -1,44 +1,44 @@
 import { Accessor, createRenderEffect, onCleanup, splitProps, type ComponentProps } from "solid-js";
-import { spread, template } from "solid-js/web";
+import { spread } from "solid-js/web";
 
-export type ScriptProps = Omit<ComponentProps<"script">, 'src'> & {
+export type ScriptProps = Omit<ComponentProps<"script">, "src" | "textContent"> & {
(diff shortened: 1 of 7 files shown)


atk:
I don't like extra events, I'd very much prefer if we could just add normal props as we would for a component. See my latest branch.

thetarnav:
Ok, I like this, it's much cleaner. and removes the need for other events as well.
Should I merge yours into this branch? there are still other changes here

atk:
Yes, please.

atk (approved):

packages/script-loader/src/index.ts
```diff
@@ -1,43 +1,45 @@
...
 /**
  * Creates a convenient script loader utility
  *
- * @param string URL or source of the script to load.
- * @param type Type value to put in the script attribute.
- * @param function Callback to trigger onLoad.
- * @param function callback on error.
- * @returns
+ * @param props The props to spread to the script element.
+ * The `src` prop is required and will be used to set the `src` or `textContent` attribute.
```
atk, line 15: ... And can also be a signal's getter.

atk (approved):

PR #294, opened 2023-01-28
jsx-parser: set returnType createToken from JSX.Element to TokenComponent<Token>

bigmistqke:
[as discussed here](https://github.com/solidjs-community/solid-primitives/pull/276#issuecomment-1407043412)


thetarnav:
Thanks :)

PR #297, opened 2023-01-30
Update README.md of @solid-primitives/intersection-observer 

samuelstroschein:
The demo is outdated. Pointing to the dev directory for the latest API. Alternatively, can Stackblitz be used to point to the dev directory?

### DIFF
--- a/packages/intersection-observer/README.md
+++ b/packages/intersection-observer/README.md
@@ -211,7 +211,7 @@ function createViewportObserver(
 
 ## Demo
 
-You may view a working example here: https://stackblitz.com/edit/vitejs-vite-n2lwpq
+You may view a working example in the [/dev](https://github.com/solidjs-community/solid-primitives/tree/main/packages/intersection-observer/dev) directory.
 
 ## Changelog
 
--- a/packages/intersection-observer/README.md
+++ b/packages/intersection-observer/README.md
@@ -137,6 +137,9 @@ import { createVisibilityObserver } from "@solid-primitives/intersection-observe
 let el: HTMLDivElement | undefined;
 
 const useVisibilityObserver = createVisibilityObserver({ threshold: 0.8 });
+
+// ! extremely important to use an accessor for the element to not lose reactivity
+// ! typescript will give you a warning if you don't use an accessor
 const visible = useVisibilityObserver(() => el);
 
 <div ref={el}>{visible() ? "Visible" : "Hidden"}</div>;


thetarnav:
hmm that's a good idea

the demo is deployed here: https://solidjs-community.github.io/solid-primitives/intersection-observer/

samuelstroschein:
added a warning for usage. see the commit message

packages/intersection-observer/README.md
```diff
@@ -137,6 +137,9 @@ import { createVisibilityObserver } from "@solid-primitives/intersection-observe
 let el: HTMLDivElement | undefined;
 
 const useVisibilityObserver = createVisibilityObserver({ threshold: 0.8 });
+
+// ! extremely important to use an accessor for the element to not lose reactivity
+// ! typescript will give you a warning if you don't use an accessor
```
thetarnav, line 142: 1. there is no reactivity to be lost if `ref` is just a `let` variable. But it is important to wrap it in a thunk so that it could be evaluated lazily, inside createEffect, when the `ref` is already populated.
For example here you don't have to use the callback form:
```tsx
let el: HTMLDivElement | undefined;

const useVisibilityObserver = createVisibilityObserver({ threshold: 0.8 });

onMount(() => {
  const visible = useVisibilityObserver(el)
  
  createEffect(() => {
    console.log(visible())
  })
})
```
2. typescript won't warn you about it - I think. because the expected param type is `MaybeAccessor<Element | FalsyValue>`

samuelstroschein: The following code did not work without an accessor function `() => element`:

```ts
<For each={Object.keys(messages())}>
  {(id) => {
    let element: HTMLDivElement | undefined;
    const isVisible = useVisibilityObserver(() => element);

    return (
      <div ref={element}>
        <Show when={isVisible()}>
          <Messages messages={messages()[id]} />
        </Show>
      </div>
    );
  }}
</For>;

```


https://user-images.githubusercontent.com/35429197/215616376-610e1966-7174-46db-a486-41c932df8dcc.mp4

samuelstroschein: Your example uses useVisbilityObserver requires the `onMount` effect and, most likely, won't work otherwise. In short, reactivity is lost if you don't use an accessor function as element is and stays undefined.

samuelstroschein: Regardin 2: Yes. TypeScript won't warn because the type is `Falsy` which is true since a ref is undefined on rendering. I couldn't come up with a better type. Thus, I added my note to the docs

samuelstroschein: Looked at the type again. 

```ts
declare function createVisibilityObserver(options?: IntersectionObserverInit & {
    initialValue?: boolean;
}, setter?: MaybeAccessor<VisibilitySetter>)

type MaybeAccessor<T> = T | Accessor<T>
```

`MaybeAccessor` should probably be replaced with `Accessor` to avoid the unforeseen "bug" I encountered.

thetarnav: > Your example requires the `onMount` effect and, most likely, won't work otherwise. In short, reactivity is lost if you don't use an accessor function as an element is and stays undefined if no external reactive mechanism sets `useVisibilityObserver` like `onMount`.

you are right, but I just wanted to be more specific about whats going on. It's not as simple as "reactivity is lost" beasue it has to nothing to do with reactivity - only timing. So the commnet is fair, I just would prefer it to explain more about whats going on, rather then simply say "don't do this".

> `MaybeAccessor` should probably be replaced with `Accessor` to avoid the unforeseen "bug" I **encountered.**

As I showed, using the primitive without function param is fine in cases where the element is present immediately. So I cannot just change the type to `Accessor`.
BUT it might be a good idea to print a warning to the console is the passed `target` is `undefined` and not an `accessor`. And the type could be changed to this, to be more helpful:
```ts
Accessor<Element | FalsyValue> | Element
```
with this it won't be possible to pass in variable that could be `undefined` without wrapping it in a thunk.

PR #300, opened 2023-02-01
jsx-parser: Add the data to the `data` property

thetarnav:
@bigmistqke I think that the parser needs to be updated.
Haven't realized this previously but currently, the token data is being restructured and spread on the token with `Object.assign`. This is needlessly more expensive and prevents adding getters to the top-level object. Not only does getter, it just breaks the reference of the passed object.
First, this would be a weird issue to walk into if you have no idea how things get applied, and second, it isn't consistent with other solid APIs (context, props) where adding getters to passed object is common practice.

Also made it so that if you don't pass the function for creating the data object, the component props will be used as the object. Which is great if you only need one token in your api.

### DIFF
--- a/.changeset/great-moons-explain.md
+++ b/.changeset/great-moons-explain.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/utils": minor
+---
+
+Add Narrow type utility.
--- a/.changeset/light-wolves-remember.md
+++ b/.changeset/light-wolves-remember.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/jsx-parser": minor
+---
+
+Add the token object to the `data` property of the token element, instead of spreading it with `Object.assign`.
--- a/.changeset/light-wolves-remember.md
+++ b/.changeset/light-wolves-remember.md
@@ -3,3 +3,5 @@
 ---
 
 Add the token object to the `data` property of the token element, instead of spreading it with `Object.assign`.
+
+Separates available functions into own exports, `parser` is now required to be passed to the functions.
--- a/packages/jsx-parser/README.md
+++ b/packages/jsx-parser/README.md
@@ -52,6 +52,8 @@ Creates a token component associated with the corresponding jsx-parser.
 - `render` function that returns the fallback JSX Element to render _(if one isn't passed, nothing will get rendred)_
 
 ```tsx
+import { createToken } from "@solid-primitives/jsx-parser";
+
 const TokenExample = createToken(
   parser,
   // function that returns the data of the token - called when the token is resolved by `resolveTokens`
--- a/packages/jsx-parser/README.md
+++ b/packages/jsx-parser/README.md
@@ -6,29 +6,28 @@
 
 A primitive to extend the types of values JSX can return. These JSX-elements are named `tokens`.
 
-- [`createJSXParser`](#createJSXParser) — Provides the tools to create and identify `tokens`: `createToken`, `childrenTokens`, `isToken` and `id`.
-- [`createToken`](#createToken) — Instantiates a `token` associated with the corresponding jsx-parser.
-- [`childrenTokens`](#childrenTokens) — A function similar to Solid's `children()`, but that will only return valid `tokens` created by the corresponding jsx-parser's `createToken`
-- [`isToken`](#isToken) — A function to validate if an element is a `token` created by the corresponding jsx-parser's `createToken`
-- [`id`](#id) — The symbol unique to the corresponding jsx-parser
+- [`createJSXParser`](#createJSXParser) — Creates a JSX Parser that can be used to create tokenized components and parse JSX Elements for tokens.
+- [`createToken`](#createToken) — Creates a token component associated with the corresponding jsx-parser.
+- [`resolveTokens`](#resolveTokens) — A function similar to Solid's `children()`, but that will only return valid token elements created by the corresponding parser's `createToken`
+- [`isToken`](#isToken) — A function to validate if an element is a `token` created by the corresponding parser's `createToken`
 
 ## Installation
 
 ```bash
 npm install @solid-primitives/jsx-parser
 # or
-yarn add @solid-primitives/jsx-parser
-# or
 pnpm add @solid-primitives/jsx-parser
+# or
+yarn add @solid-primitives/jsx-parser
 ```
 
 ## `createJSXParser`
 
-Provides the tools to create and identify `tokens`.
+Creates a JSX Parser that can be used to create tokenized components and parse JSX Elements for tokens.
 
 ### How to use it
 
-`createJSXParser` takes an optional id as argument, and returns the following functions `createToken`, `childrenTokens`, `isToken` and the symbol associated with the jsx-parser `id`.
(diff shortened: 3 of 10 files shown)


bigmistqke:
I like it. 

Wdyt of returning the data-property in `resolveChildren` instead of the token-element? Then `childrenTokens` will only contain the token-data, and there is no need for users to `token.data...`. It feels a bit more trust-worthy to have an object with data in it then a function with data attached to it. Idk feels a bit sketchy.

On second thought, there might be a usecase to do something like 
``` tsx
const tokens = childrenTokens(() => props.children)
// do some stuff w the tokens ...
return <>{tokens}</> // render the components
```
which you couldn't do anymore if u return only the data-property.

thetarnav:
yeah I left it as `token.data` so that it is still renderable.
Not sure if it needs to be though, as I assume the token-element function is only supposed to render a fallback when not parsed. All the logic for how it's supposed to be rendered if parsed should theoretically be in the `tokenData` param.
```ts
createToken((props) => {
   return {
      props,
      render() {
         return <div>hello</div>
      }
   }
})

const tokens = childrenTokens(() => props.children)
const render = createMemo(() => tokens().map(t => t.render))

<>{render}</>
```

so not sure
I would like to have the resolved tokens as data only as this is the only functionality I personally need, but if the ability to render the fallback is still important, I don't mind the extra `.data`

bigmistqke:
Mm. `All the logic for how it's supposed to be rendered if parsed should theoretically be in the tokenData param.` 
I agree and we should encourage this behavior.

It makes `childrenTokens` diverge a bit from solid's `children`, since it wouldn't really act as a cache that you can then pass again into jsx. Another usecase besides rendering would be something like [this](https://playground.solidjs.com/anonymous/908b3da3-6bb4-4bac-ac81-a0d25b01565f), which I already came by in my own experimentations.

You could do something like [this](https://playground.solidjs.com/anonymous/5aa7b8db-753e-4b64-a36c-465438ffdabd) instead for caching the children, so maybe that's fine.

thetarnav:
> It makes `childrenTokens` diverge a bit from solid's `children`, since it wouldn't really act as a cache that you can then pass again into jsx.

well it's not a cache of resolved html elements, but the tokens

> You could do something like [this](https://playground.solidjs.com/anonymous/5aa7b8db-753e-4b64-a36c-465438ffdabd) instead for caching the children, so maybe that's fine.

this solution is not without flaws
https://playground.solidjs.com/anonymous/fcda1427-b7ef-454c-afd3-67dbce749931

So to proparely support what you are trying to do, we should leave the `data` property.

But if this is not the intended use of the parser, we can make the data property the resolved token.

I don't have a use for rendering the tokens as they are, but if one wanted to do that - it would be much harder to do if the data property was resolved. When the other way around you just have to add `.data`.

bigmistqke:
> well it's not a cache of resolved html elements, but the tokens
True, I guess it's about defining what is and isn't the token. TokenElement, TokenData.

Being able to access tokens and pass them on as children to another token, that can access them again, is a quality I would like to keep tbh. Less for rendering purposes, more for not restricting more advanced token-parsing. 

I encountered a couple of times some tricky situations where the token-callback of the same token would be called multiple times through some combination of `props.children` and `childrenTokens`. It can be a bit finnicky to work with. If you do more heavy things during the token-callback, say when wrapping some third party library, it becomes a consideration.

But I am also not too big of a fan of exposing the function like we are doing now. As a user I would expect to get what I returned from the callback.

We could make it optional? `childrenTokens(()=>props.children)` && `childrenTokens(()=>props.children, {elements; false})` returns TokenDatas, `childrenTokens(()=>props.children, {elements: true})` returns TokenElements?

thetarnav:
I would probably go for separate functions if I went this path.
`resolveTokens()` and `resolveData()` maybe
Options param makes it not tree-shakeable, and it's weird when it changes the return type.

To make it actually tree-shakeable the API would need to be changed a bit though. To something similar to Solid's Context API.

```tsx
import { createJSXParser, resolveTokens, resolveData, createToken } from "@solid-primitives/jsx-parser"

const parser = createJSXParser()

const Token = createToken(parser, () => ({ ... }))

const tokens = resolveTokens(parser, () => props.children)
createEffect(() => {
   tokens()  // => TokenElement<T>[]
})
// or
const tokens = resolveData(parser, () => props.children)
createEffect(() => {
   tokens()  // => T[]
})
```

But I'm still not truly convinced. It adds additional complexity for something that is simple if you know that the data is under `data` property.
`tokens().forEach(data => ...)` is not that different from `tokens().forEach(({ data }) => ...)`

We could make it clear from the readme, or even show this example or ship this as an additional function.
```ts
// (I still like the idea of functions being separate exports)
const tokens = resolveTokens(parser, () => props.children)
const data = createMemo(() => tokens().map(t => t.data))
createEffect(() => {
   data()  // => T[]
})
```

Having separate functions would also allow a pattern like this, where tokens can be understood by multiple parsers.
Like you can have `<Select>` component that accepts `<Option>` and `<Item>`, and `<List>` that accepts only `<Item>`
```ts
const MyToken = createToken([parserA, parserB], () => ({ ... }))
```
but that's just a loose idea - wouldn't add it in this PR

bigmistqke:
> But I'm still not truly convinced. It adds additional complexity for something that is simple if you know that the data is under data property. tokens().forEach(data => ...) is not that different from tokens().forEach(({ data }) => ...)

Yes, I think you are right. It's probably best to keep the surface as small as possible.

Regarding the separate functions, I like it. It's a bit more verbose, but I guess you could always wrap it if you prefer the short way of `createToken`, etc.

I wonder if it still would be possible to type-cast the returnvalues  of `childrenTokens` to `TokenElement`s from that specific parser? That's quite handy right now. `createJSXParser` takes as a generic the union of all `TokenTypes`, so you can get some type-information from `childrenTokens`.

thetarnav:
The types work the same way thanks to [this trick](https://github.com/solidjs-community/solid-primitives/blob/c1538561c905dfdeb74c98b41e243a8a441115e3/packages/jsx-parser/src/index.ts#L7).
`createToken`, `resolveElements`  and `isToken` just have one more type generic for the token data union.

bigmistqke:
Coool. That `TYPE`-trick is surprisingly effective. Really sweet trick!

[a couple of remarks](https://playground.solidjs.com/anonymous/47bffbae-beb1-4087-bcaf-f2fd55301abd)

thetarnav:
solid playground doesnt fetch types so `import { Narrow } from solid-primitives/utils` won't work
try this: https://playground.solidjs.com/anonymous/b9788e26-2148-41aa-8782-3c5a58b29de1

bigmistqke:
lol 🤣 so nooby

Okay really neat. Yes, I think this is the most flexible setup and the workflow is nice.

Before I was typing the generics in `createToken`, but this workflow invites me more to let the generics be inferred and type the props instead. It feels a bit more error-prone, but I think that's more a feeling then a reality. 

I could always wrap `createToken` and export it if I really wanted to type the `Props` and `TokenData` without having to pass parser and `AcceptedTokens` everwhere.

PR #318, opened 2023-02-13
Added a chained typesafe provider into the i18n package

Tanner-Scadden:
- Added playground into i18n package
- Added a `makeChainedI18nContext` that takes in dictionaries and starting locale. 
- Fully typesafe i18n dictionaries accessed by property chaining to a value. If the value is a string, you can pass in arguments that are then parsed into the string built on existing foundations. If the value is a function, types are inferred and applied for additional type safety. 
- Any method that adds locales or additional dictionaries were excluded from this context to ensure type safety. 
- Safety errors are thrown to help the developer get set up and understand if things are wrong. 
- Uses `store` for reactivity.

If this addition looks good, I will write tests and documentation for this addition.


packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+    }
+  };
+
+  const chainedI18n = createMemo(() => buildI18nChain(store.dictionaries[store.locale]));
+
+  const translate = () => {
+    return chainedI18n();
+  };
+
+  const context = createContext<[typeof translate, typeof utils] | null>(null);
```
atk, line 241: Maybe you could add a context default that allows a global i18n context, so the provider is optional?

Tanner-Scadden: Yes. I can add an option to have the context be passed in by default to make it a globally scoped one. Keeping it null by default allows the user to scope things to certain areas and avoid a global dictionary. 

I'll also remove the changeLocale method since it's not needed.

Tanner-Scadden: Added the option in, removed `changeLocale`, and fixed an issue with the build after I converted it to be a `.tsx` file.

packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+  );
+
+  return {
+    I18nProvider: (props: { children: JSXElement }) => (
+      <context.Provider value={[translate, utils]}>{props.children}</context.Provider>
+    ),
+    useI18n: () => {
+      const i18nContext = useContext(context);
+      if (!i18nContext) {
+        throw new TypeError("useI18n must be used within a I18nProvider");
```
thetarnav, line 250: why a `TypeError` instead of an `Error`?
Also the error message could be committed in the production build:
```suggestion
        throw new Error(process.env.DEV ? "useI18n must be used within a I18nProvider" : "");
```

Also, shouldn't the user choose how to handle the "context not found case"? Similarly to [this](https://github.com/solidjs-community/solid-primitives/blob/0620567359ec8bebdf3c99c9254960d4590c13e9/packages/i18n/src/index.tsx#L167). imo the primitives should do as little as possible, any additional functionality that could be extracted - should be extracted.
This style of handling missing context is very common so I would prefer to add a helper for it to [@solid-primitives/context](https://github.com/solidjs-community/solid-primitives/tree/main/packages/context#readme) and let the user compose it.

packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+  [K in keyof T]: T[K] extends Record<string, unknown>
+    ? I18nPath<T[K]>
+    : T[K] extends (...args: any) => string
+    ? (...args: Parameters<T[K]>) => string
+    : T[K] extends string
+    ? (args?: I18nFormatArgs) => string
+    : never;
+};
+
+const buildI18nChain = <T,>(obj: T): I18nPath<T> => {
```
thetarnav, line 182: what does the `,` after the generic does? I guess nothing but I saw it in a couple of places already so you never know 😅
```suggestion
const buildI18nChain = <T>(obj: T): I18nPath<T> => {
```

packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+      };
+    } else {
+      throw new TypeError("Unsupported data format on the keys");
+    }
+  }, {} as Partial<I18nPath<T>>);
+
+  return paths as I18nPath<T>;
+};
+
+export const makeChainedI18nContext = <T extends object, K extends keyof T>(props: {
```
thetarnav, line 212: Shouldn't there be a required interface for the translations (or directories) object? Instead of `object`. You are making assumptions that all functions will return `strings` and uses `I18nFormatArgs` for parameters so it would be useful to guide users to that.

thetarnav (commented):
Thanks for putting this together.
I wanted to tell you to add it as a separate primitive to the `i18n` package to test out the design but you've read my mind anyway 😄
You should add yourself to the `contributors` list in package.json and the new primitive to the primitives list.
The readme needs to be updated as well, but this can wait until after the design part is finished.
I've left some comments about the things I've noticed, but I still have to run it locally and test it.

packages/i18n/src/index.tsx
Tanner-Scadden: I would love to make this an improved interface. I'm not sure if my typescript skills are good enough to enforce this on the object that is being passed in, but I'll play with it some more. 

The current typescript will resolve any key that doesn't fit the type def into never if you try to use it and there is a type error that is thrown with some helper information.

packages/i18n/src/index.tsx
Tanner-Scadden: In TSX generics need a comma after them. I'm not sure why, but the linters aren't happy with it unless there is one.

packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+  dictionaries: T;
+  locale: keyof T;
+  setContext?: boolean;
+}) => {
+  const [store, setStore] = createStore<{ dictionaries: T; locale: keyof T }>(props);
```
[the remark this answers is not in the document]
Tanner-Scadden: Good point. Reverted it to only save the locale in a signal as there isn't a need to even save the dictionary to the store since it's never mutated.

packages/i18n/src/index.tsx
```diff
@@ -157,10 +158,98 @@ export const createI18nContext = (
...
+
+  const utils = {
+    locale(locale?: K): K {
```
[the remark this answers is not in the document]
Tanner-Scadden: Updated

packages/i18n/src/index.tsx
```diff
@@ -23,7 +23,7 @@ import { createStore } from "solid-js/store";
  * // => 'not found'
  * ```
  */
-const deepReadObject = <T = any>(
+const deepReadObject = <T extends any>(
```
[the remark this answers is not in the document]
Tanner-Scadden: Reverted with the file change back to .ts

packages/i18n/src/index.tsx
Tanner-Scadden: Removed this error handle and updated my example to show wrapping it yourself if you want to handle it easier.

Tanner-Scadden:
Added tests and updated readmes!

packages/i18n/src/index.tsx
thetarnav: Ah I know now
It's because of the `tsx` extension - without it `<T>` is interpreted as JSX

Tanner-Scadden:
When I was adding server tests, I had some issues with the translations to work if the locale is changed. I had to update the API back to using a store for the current dictionary. Doing so I think also provided the end user with a better API. 

Instead of 
```ts
t().hello({ name: name() })
```
It is now just:
```ts
t.hello({ name: name() })
```

packages/i18n/src/index.tsx
Tanner-Scadden: That makes sense!

packages/i18n/src/index.tsx
Tanner-Scadden: Updated the typing on this and it will now have issues if there is an end key path does not return a string or a function that does not return a string.

packages/i18n/src/index.ts
```diff
@@ -157,10 +158,135 @@ export const createI18nContext = (
...
+export const makeChainedI18nContext = <
+  T extends Dictionaries<I18nObject>,
+  K extends keyof T
+>(props: {
+  dictionaries: T & GuaranteeIdenticalSignatures<T>;
+  locale: keyof T;
+  setContext?: boolean;
+}) => {
+  const [locale, _setLocale] = createSignal<K>(props.locale as K);
+  const [translate, _setTranslate] = createStore(buildI18nChain(props.dictionaries[props.locale]));
```
thetarnav, line 262: Translations don't have to be a store either.
This is what meant in the previous comment.
All the properties of the dictionary objects will be mostly unique. Even if two languages will share the same word for something - there is no point in optimizing for that. You pretty much always want to rerun everything that is listening to the translations once the locale changes.
But since `t.hello` is nicer than `t().hello` you can do something like this:
https://playground.solidjs.com/anonymous/d7ff4242-94f3-43e5-85bc-f99dff3e6519

packages/i18n/src/index.ts
```diff
@@ -157,10 +158,135 @@ export const createI18nContext = (
...
+    I18nProvider: (props: { children: JSXElement }): JSXElement => {
+      return createComponent(chainedI18nContext.Provider, {
+        value: [translate, utils],
+        get children() {
+          return props.children;
+        }
+      });
+    },
+    useI18nContext: () => useContext(chainedI18nContext)
+  };
```
thetarnav, line 291: A primitive for i18n should worry only about i18n. How do you decide to share returned signals with the rest of the app? that's a separate concern. It could be done in many ways. What if you prefer to have one giant context provider for one app instead of multiple smaller ones? What if you want to share the translations by just exporting them from a file?
What do you think about separating the current primitive into two?
- one that creates the translations object and manages the locales. e.g. `createChainedI18n`
- and one that could wrap it into a convenient context if the user wants it? e.g. `createI18nContext`
Although I'm not sure if the second one for context is even necessary. If it could be abstracted and added to the /context package to be used with any state it would be much better than creating one for each.

packages/i18n/src/index.ts
```diff
@@ -157,10 +158,135 @@ export const createI18nContext = (
...
+        ...acc,
+        [key]: (args: I18nFormatOptions) => {
+          return value.replace(/{{(.*?)}}/g, (_, key) => deepReadObject(args, key, ""));
+        }
+      };
+    } else {
+      throw new TypeError(
+        `Unsupported data format on the keys. Values must resolve to a string or a function that returns a string. Key name: "${
+          key as string
+        }"`
```
thetarnav, line 241: This could be behind a DEV check
```suggestion
        process.env.DEV ? `Unsupported data format on the keys. Values must resolve to a string or a function that returns a string. Key name: "${
          key as string
        }"` : ""
```

thetarnav (changes requested):
this is looking better and better 👍 
I have last few comments and we can proceed to merge this in.

packages/i18n/src/index.ts
Tanner-Scadden: Okay. Maybe it's better to make this just a utility for i18n that people can use to build out chained dictionaries then they decide how to ingest it! Would make it a lot less opinionated and a smaller bundle.

packages/i18n/src/index.ts
Tanner-Scadden: I can move it behind a DEV flag. This is a dead end that will cause it to crash if you tried to call it and would be typed as never.

packages/i18n/src/index.ts
Tanner-Scadden: I'll play around with this. Very interesting, never seen this before. Will be fun to learn

packages/i18n/src/index.ts
thetarnav: not saying to remove the error. But the message could be an empty string in production and a helpful message for the developer during development. No point shipping it to the end-user.

packages/i18n/src/index.ts
Tanner-Scadden: After playing with this more, I believe that i18n needs context. I split it up, but without providing a friendly way to interact with the translations, a user will need to create their own wrapper every time to use the data and handle locales.

packages/i18n/src/index.ts
Tanner-Scadden: Instead of doing it this way, I made the t function a proxy that returns the dictionaries with the current locale path.

Tanner-Scadden:
If the latest changes look good, I will add the new primitive into the readme as well. Tests were added to make sure it all works correctly.

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,167 @@
...
+ * @returns {string} - The fully injected template
+ *
+ * @example
+ * ```js
+ * const txt = template('Hello {{ name }}', { name: 'Tom' });
+ * // => 'Hello Tom'
+ * ```
+ */
+const template = (str: string, params: Record<string, string>, reg: RegExp = /{{(.*?)}}/g): any =>
+  str.replace(reg, (_, key) => deepReadObject(params, key, ""));
```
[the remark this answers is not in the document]
atk: Instead of `.*?`, you could do `[^{}]+`.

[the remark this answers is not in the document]
Tanner-Scadden: Switched it out and that one works great!

thetarnav:
Okay, the last piece would be updating readme and adding some jsdoc. @Tanner-Scadden you want to do this, or should I?

Tanner-Scadden:
> Okay, the last piece would be updating readme and adding some jsdoc. @Tanner-Scadden you want to do this, or should I?

I'll get it updated!

Tanner-Scadden:
I18n docs are updated, updated readme to include new primitives, and added jsdocs. I18n is mentioned in the main readme twice, is that intentional?

thetarnav:
> I18n is mentioned in the main readme twice, is that intentional?

The main readme is updated automatically in a gh action so don't worry about it. Better to not update in manually in PRs to avoid merge conflicts.

thetarnav (approved):

PR #319, opened 2023-02-14
feat: new autofocus primitive

jer3m01:
Solid does not natively support the `autofocus` attribute. This PR adds a new package with 2 new primitives to support it.
`@solid-primitives/autofocus`:
- `use:autofocus` a directive to use on HTML components.
- `createAutofocus` a ref based autofocus for custom components.

See [autofocus/README](https://github.com/jer3m01/solid-primitives/blob/feat/autofocus/packages/autofocus/README.md) for more details and examples.

The `use:autofocus` directive uses the native `autofocus` attribute to decide on focusing. This provides a seamless integration with existing code<sup>1</sup>, but might come as repetitive/confusing (`use:autofocus autofocus`). Let me know what you think! 

1: Add `use:autofocus` and, provided the intrinsic props are passed down, no extra logic is required.

### DIFF
--- a/packages/autofocus/README.md
+++ b/packages/autofocus/README.md
@@ -36,7 +36,7 @@ import { autofocus } from "@solid-primitives/autofocus";
 </button>;
 ```
 
-The `autofocus` directive uses the native `autofocus` attribute to determine if it should focus th element or not.
+The `autofocus` directive uses the native `autofocus` attribute to determine it should focus the element or not.
 Using this directive without `autofocus={true}` (or the shorthand `autofocus`) will not perform anything.
 
 ### createAutofocus
--- a/packages/autofocus/README.md
+++ b/packages/autofocus/README.md
@@ -28,7 +28,7 @@ pnpm add @solid-primitives/autofocus
 
 ### use:autofocus
 
-```ts
+```tsx
 import { autofocus } from "@solid-primitives/autofocus";
 
 <button use:autofocus autofocus={true}>
@@ -41,7 +41,7 @@ Using this directive without `autofocus={true}` (or the shorthand `autofocus`) w
 
 ### createAutofocus
 
-```ts
+```tsx
 import { createAutofocus } from "@solid-primitives/autofocus";
 
 // Using ref
--- a/packages/autofocus/README.md
+++ b/packages/autofocus/README.md
@@ -11,6 +11,8 @@
 
 Primitives for autofocusing HTML elements.
 
+The native autofocus attribute only works on page load, which makes it incompatible with SolidJS. These primitives run on render, allowing autofocus on initial render as well as dynamically added components.
+
 - `autofocus` - Directive to autofocus an element on render.
 - `createAutofocus` - Reactive primitive to autofocus an element on render.
 
@@ -31,9 +33,12 @@ pnpm add @solid-primitives/autofocus
 ```tsx
 import { autofocus } from "@solid-primitives/autofocus";
 
-<button use:autofocus autofocus={true}>
+<button use:autofocus autofocus>
   Autofocused
 </button>;
+
+// autofocus directive can be disabled if `false` is pased as options
+<button use:autofocus={false}>Not Autofocused</button>;
 ```
 
 The `autofocus` directive uses the native `autofocus` attribute to determine it should focus the element or not.
--- a/packages/autofocus/README.md
+++ b/packages/autofocus/README.md
@@ -26,7 +26,7 @@ pnpm add @solid-primitives/autofocus
 
 ## How to use it
 
-### use:autofocus directive
+### use:autofocus
 
 ```ts
 import { autofocus } from "@solid-primitives/autofocus";
@@ -39,7 +39,7 @@ import { autofocus } from "@solid-primitives/autofocus";
 The `autofocus` directive uses the native `autofocus` attribute to determine if it should focus th element or not.
 Using this directive without `autofocus={true}` (or the shorthand `autofocus`) will not perform anything.
 
-### createAutofocus;
+### createAutofocus
 
 ```ts
 import { createAutofocus } from "@solid-primitives/autofocus";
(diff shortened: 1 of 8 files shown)


packages/autofocus/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param autofocus - Should this directive be enabled. defaults to false.
+ *
+ * @example
+ * ```ts
+ * import { autofocus } from "@solid-primitives/autofocus";
+ *
+ * <button use:autofocus autofocus={true}>Autofocused</button>;
+ * ```
+ */
+export const autofocus = (element: HTMLInputElement, autofocus: () => boolean) => {
```
thetarnav, line 24: The type of `element` param should probably be `Element`. (in the readme you used a button for the example so.. )
You can use the `Directive` type from utils for it: https://github.com/solidjs-community/solid-primitives/blob/main/packages/utils/src/types.ts#L19
```suggestion
export const autofocus = (element: Element, autofocus?: () => boolean) => {
```
Also if you make the `autofocus` parameter optional, it could be used nicely with `ref` as well:
```tsx
<button autofocus ref={el => {
  autofocus(el)
}}/>
```
But I guess that `createAutofocus` is for that

packages/autofocus/README.md
```diff
@@ -0,0 +1,68 @@
...
+## How to use it
+
+### use:autofocus
+
+```tsx
+import { autofocus } from "@solid-primitives/autofocus";
+
+<button use:autofocus autofocus={true}>
+  Autofocused
+</button>;
```
thetarnav, line 36: `={true}` is implicit, so point in adding if I think (same as you don't do `use:autofocus={true}`)
```suggestion
<button use:autofocus autofocus>
  Autofocused
</button>;
```

packages/autofocus/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
```
thetarnav, line 37: Worth mentioning
```suggestion

// autofocus directive can be disabled if `false` is pased as options
<button use:autofocus={false}>
  Not autofocused
</button>
```
```

File: packages/autofocus/README.md [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+</p>
+
+# @solid-primitives/autofocus
+
+[![turborepo](https://img.shields.io/badge/built%20with-turborepo-cc00ff.svg?style=for-the-badge&logo=turborepo)](https://turborepo.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/autofocus?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/autofocus)
+[![version](https://img.shields.io/npm/v/@solid-primitives/autofocus?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/autofocus)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
+
+Primitives for autofocusing HTML elements.
```
thetarnav, line 12: I think the package needs some more explanation. It's not obvious from the readme why using the `autofocus` attribute might not be enough.

packages/autofocus/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ * ```ts
+ * let ref!: HTMLButtonElement;
+ *
+ * createAutofocus(() => ref); // Or using a signal accessor.
+ *
+ * <button ref={ref}>Autofocused</button>;
+ * ```
+ */
+export const createAutofocus = (
+  ref: Accessor<HTMLElement | undefined>,
```
thetarnav, line 54: You can use `FalsyValue` from utils to allow for this pattern as well:
```ts
createAutofocus(() => enabled() && el())
```
```suggestion
  ref: Accessor<HTMLElement | FalsyValue>,
```

packages/autofocus/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  autofocus: boolean = true
+) => {
+  createEffect(() => {
+    if (autofocus)
+      setTimeout(() => {
+        ref()?.focus();
+      });
+  });
```
thetarnav, line 62: setTimeout inside effect will cause the `ref` to not be tracked.
```suggestion
  createEffect(() => {
    const el = ref()
    el && setTimeout(() => el.focus());
  });
```

packages/autofocus/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ * let ref!: HTMLButtonElement;
+ *
+ * createAutofocus(() => ref); // Or using a signal accessor.
+ *
+ * <button ref={ref}>Autofocused</button>;
+ * ```
+ */
+export const createAutofocus = (
+  ref: Accessor<HTMLElement | undefined>,
+  autofocus: boolean = true
```
thetarnav, line 55: I wouldn't add a param that completely disables the primitive, no point using it at all if you don't want to have it. I know that you cannot do that as easily with directives, but with primitives it's not neccessry:
```ts
if (isEnabled) createAutofocus(ref)
```

thetarnav (changes requested):
Thanks for bringing it here, it looks like a nice utility to have.
I've pointed a couple of things to change though. They should be quick to resolve and I'll merge this in :)

packages/autofocus/src/index.ts
jer3m01: Changed it to `HTMLElement`.
I can't use `Directive` as it accepts an `Element` and I require an `HTMLElement` to `.focus()`.

packages/autofocus/src/index.ts
```diff
@@ -0,0 +1,62 @@
+import { createEffect, onMount, JSX, Accessor } from "solid-js";
+import { FalsyValue } from "../../utils";
```
thetarnav, line 2: utils package have to be imported and added to package.json file
```suggestion
import { FalsyValue } from "@solid-primitives/utils";
```

thetarnav (approved):
👍 🎉 🎉

PR #320, opened 2023-02-14
Allow for passing `null` to `createRoot` in `createSubRoot`.

Exelord:
From version Solid v1.6.11 createRoot now correctly handles `null` values as detachedOwner in createRoot. It should be allowed to createSubRoots with `null` owner in order to not inherit the context.


thetarnav:
right, thanks for adding this (to solid as well)

PR #321, opened 2023-02-15
createForm proposal

Tanner-Scadden:
- exports a createForm primitive 
- Uses lodash.set to handle nested sets
- Uses zod to do schema validations 
- Returns a handleSubmit function to put onto a form
- returns data gathered in the form. Converts number values to numbers for ease of use. 
- If there is an error from the validation, returns the zod errors & form data for debugging.

### DIFF
--- a/packages/form/CHANGELOG.md
+++ b/packages/form/CHANGELOG.md
@@ -1,0 +1,5 @@
+# @solid-primitives/form
+
+0.0.100
+
+First commit of the form primitive.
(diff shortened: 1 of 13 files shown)


Tanner-Scadden:
@thetarnav Let me know what else needs to be done for this proposal!

davedbase:
Hey! This has grown a bit old and with the new Solid2 release we have new primitives to account for. I'm going to close this in favour of a new PR that has been pushed for 2.0 release.

PR #325, opened 2023-02-17
Update solid-js to 1.7.0

thetarnav:
This is to make sure that we can transition to 1.7 smoothly


thetarnav:
Done in 43ac489f6c1155b0b808eb26859e0e679a123ffd

Issue #310, opened 2023-02-08
Hydration issue


TomPichaud:
Following https://github.com/withastro/astro/issues/6140

Hello,

I have a simple Solid component with a `usePrefersDark`. I expect the component to render with the default value and then hydrate with the correct value. It's only working properly with the `client:only` Astro property.

**Bug**: inlined value not updated on hydration

If I create my own signal and event listener it's working properly.

I'm willing to help to fix it but I would require some guidance. Thank you for your help.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-tzjxcu?file=src/pages/index.astro

thetarnav:
Hey, thanks for the issue.
I've heard that the media package is having some issues with hydration so I'm not supprised.
It has to be related to how `createMediaQuery` is implemented: https://github.com/solidjs-community/solid-primitives/blob/main/packages/media/src/index.ts#L47
I don't have much experience with hydration in solid so I might miss some issues there.
My guess is that the initial value of the signal gets ignored during hydration and effects will only take it's value on subsequent updates - but since the media query doesn't change - it is out of sync.
So the fix might be as simple as adding `createEffect` that will re-set the value of the signal. But if you set the signal to the value it had before - effects won't retrigger... So maybe you'd need to play with the `equals` option of signals to force update when the effect runs but only update if the signal is different afterwards...
This is interesting. wonder if there is a way to run some logic only during hydration, but keep it simple during normal CSR. Becasue there is no point in re-setting the signal value on CSR...

So that's about fixing it. But then testing. In out repo we have /dev directory with a simple vite project so it won't help much for hydration. The unit tests are either in SSR or CSR.
Maybe we need to figure out a way to better test hydration.

So probably the best way to test it would be to pnpm link the package and run it the the example astro project you've prepared.

So yeah, if you are down to help I would appreciate it :)
and ask away if you have any questions

TomPichaud:
I feel like the issue runs deeper into solid. Take the following code:

```jsx
const prefersDark= usePrefersDark()
createEffect(() => {
	console.log(prefersDark())
})
return (
	<>
		<pre>prefersDark : {prefersDark() ? 'true' : 'false'}</pre>
		<pre textContent={`should be: ${prefersDark().toString()}`} />
	</>
)
```

The first one is bugged and showing the pre-render value until a new change is triggered; but the second one is synced correctly. How can you explain it ? _The log is also correct._

thetarnav:
[...]

PR #326, opened 2023-02-19
utils: new createHydrateSignal - fix:hydration issues

TomPichaud:
Following https://github.com/solidjs-community/solid-primitives/issues/310

New `createHydrateSignal` with test on `utils:package`

Fixes hydration issue and improve shared root usage on : 

- `createMediaQuery` & `usePrefersDark`
- `createActiveElement` & `createFocusSignal`
- `createConnectivitySignal` & `useConnectivitySignal`
- `createPageVisibility` & `usePageVisibility`
- `createRemSize` & `useRemSize`

Fixes SSR function return order on `createWebShare`

Thank you @thetarnav for helping on my first contribution, I've tried my best.

### DIFF
--- a/.changeset/twenty-glasses-act.md
+++ b/.changeset/twenty-glasses-act.md
@@ -1,0 +1,11 @@
+---
+"@solid-primitives/active-element": patch
+"@solid-primitives/connectivity": patch
+"@solid-primitives/media": patch
+"@solid-primitives/page-visibility": patch
+"@solid-primitives/share": patch
+"@solid-primitives/styles": patch
+"@solid-primitives/utils": minor
+---
+
+Add `createHydrateSignal` primitive to utils and fix hydration issues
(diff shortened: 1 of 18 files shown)


thetarnav (approved):
Ok, this is great, in fact, I have nothing to point out. You could add yourself as a contributor to the package.json files of packages you've updated.
The logic of picking between using the shared root and not definitely feels repeated, but I think this is something that could be addressed in another PR.
If you have nothing more to add to this, I'm down to merge it 👍

TomPichaud:
Thank you, it's done, you can merge it!

Concerning the `createSharedRoot` we could do something like this:

```ts
export function createSharedRoot<T>(factory: (dispose?: VoidFunction) => T): () => T {
   if(sharedConfig.context){
      console.warn("Shared root is disable on island architecture.")
      return factory
   }
...
```

What's your thought ?

thetarnav:
something like that.
But this should be something that you enable in the options param - I would leave the default behavior as is for now.
And the factory function should receive a dispose function, so it should be wrapped in solid's createRoot.
Also, I wouldn't console.warn about it. if you enable it in the options - that is expected behavior.

PR #327, opened 2023-02-19
New `flux-store` package

Zalexios:
SolidJS Primitive to simplify the creation of global stores and the ability to access and mutate them.

###### stores/counter-store.ts

```ts
// `counter-store.ts`
import { createStoreFactory } from "@solid-primitives/store";
const [CounterProvider, { useStore: useCounterStore }] = createStoreFactory(
  {
    value: 5
  },
  (state, setState) => ({
    count: () => state.value,
    increment: () => setState(val => ({ value: val.value + 1 })),
    reset: () => setState({ value: 0 })
  })
);
export { CounterProvider, useCounterStore };
 ```
###### App.tsx
 ```tsx
// `App.tsx`
import { CounterProvider } from "./stores/counter-store.ts";

// Wrap the app in the store's Provider
<CounterProvider>
  <App />
</CounterProvider>;
 ```
###### pages/Example.tsx
 ```tsx
// `Example.tsx`
import { useCounterStore } from "../stores/counter-store.ts";

const [counterState, { count, increment, reset }] = useCounterStore();
count(); // => 5
increment();
count(); // => 6
reset(); // => 0
```

## Demo

View Demo:
https://vu5z5u-3000.preview.csb.app/

Open Demo Editor:
https://codesandbox.io/p/sandbox/solid-primitives-store-web-demo-vu5z5u

[README](https://github.com/solidjs-community/solid-primitives/blob/458fda3011ece42a3bc11b53cced859c9d0fc592/packages/store/README.md)


Zalexios:
@thetarnav the build workflow failed and it looked related to the changeset related commit so I removed the file it created. I think everything should be ready now. Please let me know if I'm missing anything or if there are additional tasks I need to complete. Thank you.

thetarnav (commented):
Thank you for this very interesting addition.
I think this could be a nice pattern for creating shared stores, but it also seems like it could be separated into individual primitives to provide more composable (so adaptable) pieces.
It reminds me of a recent PR extending the i18n package: https://github.com/solidjs-community/solid-primitives/pull/318
It had the same issue where two separate concerns were joined together, making the whole primitive a bit more opinionated. I'm referring to how 1. managing state and actions, and 2. way to share that state with the application are placed under the same primitive.
What I proposed there also applies very well here: https://github.com/solidjs-community/solid-primitives/pull/318#discussion_r1108358810
The [`createContextProvider`](https://github.com/solidjs-community/solid-primitives/tree/main/packages/context#createcontextprovider) primitive already takes care of providing a easy interface to share state using context in a typesafe way. [It could also be reused](https://github.com/solidjs-community/solid-primitives/blob/main/packages/i18n/src/chained-i18n.ts#L182) if it makes sense to include it in the package.

Another thing that concerns me is the name. `createStoreFactory` doesn't reflect what the primitive is doing - it doesn't create a factory function for making reusable stores. So or the primitive should be renamed, or it should consider the reusablity.
And I think that considering reusability would be really worth it. You could imagine it as classes or state machines with specific purpose that can be used in various, multiple places at once.
It could be as easy as wrapping the returned value in thunk just to defer and reuse it. It should also provide a way to override the initial (which should rather be `fallback`) state.

```tsx
const createCounterState = createStoreFactory(
  // actions factory
  (state, setState) => ({
    // having accessors here breakes the idea of actions a bit...
    // probably should be rethinked
    // the distinction is useful to have actions untracked and accessors tracked
    count: () => state.value,
    increment: () =>
      setState(
        // produce ban be composed
        produce((proxy) => {
          proxy.value++;
        })
      ),
    reset: () => setState("value", 0),
  }),
  // fallback state - if provided, makes the initial state optional
  { value: 5 }
);

// normal context api could be used, but there could be a wrapper for context as well
const CounterCtx = createContext<StateType<typeof createCounterState>>();

function Counter(props: { initialValue?: number }) {
  const [state, { increment }] = createCounterState({ value: initialValue });

  return (
    <CounterCtx value={state}>
      <button onClick={increment}>{state.count}</button>
    </CounterCtx>
  );
}
```

Next would be the actions. If they are actions - the dedicated way the state should be updated - it would be nice if they were `batch`ed and `untrack`ed by default. Someone proposed adding a `createAction` primitive on discord just for that, so it could be added here I think. It suits the theme of the package - simplifying the common patterns of state management.

I don't like bundling `produce` and `unwrap` in with everything. They were designed as composable and tree-shakeable primitives in core, and this is breaking that design. If you are already returning `setState` then both of them can already be used so it doens't seem to bring any value. The primitives should focus only on one thing and do as little as possible.
If having a standalone `produce` function that could be used without the setter seems valuable to you, you can make a `createProduce` primitive that will wrap it 🤷‍♂️

Zalexios:
> Thank you for this very interesting addition. I think this could be a nice pattern for creating shared stores, but it also seems like it could be separated into individual primitives to provide more composable (so adaptable) pieces. It reminds me of a recent PR extending the i18n package: #318 It had the same issue where two separate concerns were joined together, making the whole primitive a bit more opinionated. I'm referring to how 1. managing state and actions, and 2. way to share that state with the application are placed under the same primitive. What I proposed there also applies very well here: [#318 (comment)](https://github.com/solidjs-community/solid-primitives/pull/318#discussion_r1108358810) The [`createContextProvider`](https://github.com/solidjs-community/solid-primitives/tree/main/packages/context#createcontextprovider) primitive already takes care of providing a easy interface to share state using context in a typesafe way. [It could also be reused](https://github.com/solidjs-community/solid-primitives/blob/main/packages/i18n/src/chained-i18n.ts#L182) if it makes sense to include it in the package.
> 
> Another thing that concerns me is the name. `createStoreFactory` doesn't reflect what the primitive is doing - it doesn't create a factory function for making reusable stores. So or the primitive should be renamed, or it should consider the reusablity. And I think that considering reusability would be really worth it. You could imagine it as classes or state machines with specific purpose that can be used in various, multiple places at once. It could be as easy as wrapping the returned value in thunk just to defer and reuse it. It should also provide a way to override the initial (which should rather be `fallback`) state.
> 
> ```tsx
> const createCounterState = createStoreFactory(
>   // actions factory
>   (state, setState) => ({
>     // having accessors here breakes the idea of actions a bit...
>     // probably should be rethinked
>     // the distinction is useful to have actions untracked and accessors tracked
>     count: () => state.value,
>     increment: () =>
>       setState(
>         // produce ban be composed
>         produce((proxy) => {
>           proxy.value++;
>         })
>       ),
>     reset: () => setState("value", 0),
>   }),
>   // fallback state - if provided, makes the initial state optional
>   { value: 5 }
> );
> 
> // normal context api could be used, but there could be a wrapper for context as well
> const CounterCtx = createContext<StateType<typeof createCounterState>>();
> 
> function Counter(props: { initialValue?: number }) {
>   const [state, { increment }] = createCounterState({ value: initialValue });
> 
>   return (
>     <CounterCtx value={state}>
>       <button onClick={increment}>{state.count}</button>
>     </CounterCtx>
>   );
> }
> ```
> 
> Next would be the actions. If they are actions - the dedicated way the state should be updated - it would be nice if they were `batch`ed and `untrack`ed by default. Someone proposed adding a `createAction` primitive on discord just for that, so it could be added here I think. It suits the theme of the package - simplifying the common patterns of state management.
> 
> I don't like bundling `produce` and `unwrap` in with everything. They were designed as composable and tree-shakeable primitives in core, and this is breaking that design. If you are already returning `setState` then both of them can already be used so it doens't seem to bring any value. The primitives should focus only on one thing and do as little as possible. If having a standalone `produce` function that could be used without the setter seems valuable to you, you can make a `createProduce` primitive that will wrap it 🤷‍♂️
> 
> I think the takeaway from my too long comment would be to try separating the patterns used in current `createStoreFactory` into individual, composable primitives. The more the better. Then users could evaluate which ones make sense to them and use them.


Thanks for all the feedback. Reactive shared stores are exactly what I was going for. I wanted to replicate the simplicity and global accessibility of `useState` in Nuxt. I had looked at the `context` package and while it provides a simplified experience creating the `Provider` and `useContext`, it doesn't allow use of a `Store` unless you manually construct it and pass it as the defaults argument, and it incorporates the use of `Provider` to set the initial values whereas I'd like to eliminate the `Provider` altogether (other than the necessity to wrap the App with it).

I see your point that providing `produce` and `unwrapped` is quite unnecessary so I will remove those.

You gave me an idea, a better name would be:
`createSharedStore`

I could further slim the returned values by returning the `SharedStoreProvider`, `useStore` aka useContext(Context) and either return `setState` or just return `store`.

As for pulling out the `createActions`, they're typically included in the `createContext` function but I don't think that's a necessity. Perhaps it's just a separate function that accepts a `Store` or better yet `setState` (or both) to address your comment about the accessors in the actions and the second parameter would be the actions. This would allow it to be used on its own.

I need to think about the simplest way to incorporate a fallback or default value override. I'll think about that while making these changes. Thanks again for the advice.

Zalexios:
@thetarnav 

I refined createActions and limited it to mutations, only passing `setState` which worked but ended up creating a situation where a more complex solution might be necessary. Considering the goal is to create a shared store, any requirement for `getters` breaks the flow and would have required manually creating them. This led me to expand the second parameter into an object containing getters, actions, both or neither. The `state` is provided when creating `getters` and `setState` is provided when creating `actions`. The results enabled type inference for any access/mutation methods provided and so long as something is provided it will not be undefined nor potentially undefined.

It reminded me of how **Pinia** creates stores with its `defineStore` function that accepts the `state` and optional [getters](https://pinia.vuejs.org/core-concepts/getters.html) and [actions](https://pinia.vuejs.org/core-concepts/actions.html).

These changes make `createSharedStore` feel like a lightweight state management library.

Right now it returns the `Provider`, `useStore` and then I also include `setState` from the `Store` object as it's required if the user wants to use `produce` and/or `reconcile`.


```


export interface SharedStoreMethods<T extends object, G extends object | undefined, A extends object | undefined> {
  getters?: (state: T) => G;
  actions?: (setState: SetStoreFunction<T>) => A;
}

/**
 * An interface representing the returned shared state functionality. 
 * It consists of the state, any predefined methods to access and mutate the state and the setState function.
 *
 * @interface SharedStore
 * @template <T: state type of the store, G: store access methods, A: store mutation methods>
 * @property {() => [state: T, methods: { getters: G; actions: A; }]} useStore - A function to retrieve the current state and access/mutation functions.
 * @property {SetStoreFunction<T>} setState - A function to manually mutate the current state of the store.
 */
export interface SharedStore<T extends object, G extends object | undefined, A extends object | undefined> {
  useStore: () => [state: T, methods: {
    getters: G;
    actions: A;
  }];
  setState: SetStoreFunction<T>;
}

/**
 * A tuple for a Shared Store
 *
 * @typedef SharedStoreReturn
 * @template <T: state type of the store, G: store access methods, A: store mutation methods>
 * @property {FlowComponent<Partial<T>>} SharedStoreProvider - A Flow component that provides partial state to the store
 * @property {SharedStore<T, G, A>} SharedStore - The state, any predefined methods to access and mutate the state and the setState function.
 */
export type SharedStoreReturn<T extends object, G extends object | undefined, A extends object | undefined> = [
  SharedStoreProvider: FlowComponent<Partial<T>>,
  SharedStore: SharedStore<T, G, A>
];


export function createSharedStore<T extends object>(initialState: T): SharedStoreReturn<T, undefined, undefined>;
export function createSharedStore<T extends object, G extends object | undefined, A extends object | undefined>(
  initialState: T,
  createMethods: SharedStoreMethods<T, G, A>
): SharedStoreReturn<T, undefined extends G ? undefined : G, undefined extends A ? undefined : A>;
export function createSharedStore<T extends object, G extends object | undefined, A extends object | undefined>(
  initialState: T,
  createMethods?: SharedStoreMethods<T, G | undefined, A | undefined>
): SharedStoreReturn<T, G | undefined, A | undefined> {
  const store = createStore<T>(initialState);
  const [state, setState] = store;
  const getters = createMethods?.getters?.(state);
  const actions = createMethods?.actions?.(setState);
  const methods = {
    getters,
    actions,
  };
  type Getters = typeof getters;
  type Actions = typeof actions;
  type Methods = typeof methods;

  const SharedStoreContext = createContext<[state: T, methods: Methods]>([state, methods]);
  const useStore = () => useContext(SharedStoreContext);
  const SharedStoreProvider: FlowComponent<Partial<T>> = props =>
    SharedStoreContext.Provider({
      ...props,
      value: useStore(),
      children: props.children
    });

  return [
    SharedStoreProvider,
    {
      useStore,
      setState
    }
  ] as SharedStoreReturn<T, Getters, Actions>;
}

```

thetarnav:
> These changes make createSharedStore feel like a lightweight state management library.

Thats the point 😉
Not only lightweight, but composable.

> As for pulling out the `createActions`, they're typically included in the `createContext` function but I don't think that's a necessity. Perhaps it's just a separate function that accepts a `Store` or better yet `setState` (or both) to address your comment about the accessors in the actions and the second parameter would be the actions. This would allow it to be used on its own.

The actions should be batched and untracked though. This is what I have in mind:
https://playground.solidjs.com/anonymous/45202299-73a0-4746-bd28-848a04f5ddf2

Which can then be nicely turned into a primitive for creating reusable states:
https://playground.solidjs.com/anonymous/72e8ab57-32bb-4d82-8dda-647c415b79d5

Context is a separate concern here and I wouldn't include it in the base primitives. They shouldn't prescribe how the store is supposed to be shared.
If you need a global store shared with context, you can easily use this `createSharedStore` with `createContextProvider` from the context package to create that expirience.
https://stackblitz.com/edit/github-c6tkmr?file=src%2FApp.tsx

Zalexios:
@thetarnav  Thank you, this is fantastic feedback. I have two quick questions:

1. I'm under the impression it's best practice to use the `setState` previous state argument if your mutation depends on value(s) from the `state`. You reminded me of this when recommending I change the way actions are created. I plan to remove `state` from `actions`, what do you think?

So instead of:
```tsx
const createCounter = createStoreFactory(
  { value: 0 },
  {
    getters: (state) => ({}),
    actions: (state, setState) => ({
      increment(by: number = 1) {
        if (state.value > 49) return;
        setState("value", (p) => p + by);
      },
    }),
  }
);
```

We could, for example, use something like:
```tsx
const createCounter = createStoreFactory(
  { value: 0 },
  {
    getters: (state) => ({}),
    actions: (setState) => ({
      increment2(by: number = 1) {
        setState(prev => {
          if (prev.value > 49) return prev;
          return {
            ...prev,
            value: prev.value + by,
          };
        })
      },
      increment3(by: number = 1) {
        setState("value", prev => prev > 49 ? prev : prev + by);
      },
    }),
  }
);
```
2. I noticed the native `SetStoreFunction` already wraps the mutation function passed to it with `batch`. Does that make the `batch` inside `createAction` redundant?

Would these basically be the same considering they'll contain `setState` mutations? Could `batch` be removed as in `createAction2`?
```tsx
export function createAction<T extends AnyFunction>(
  fn: T
): (...args: Parameters<T>) => ReturnType<T> {
  return (...args) => batch(() => untrack(() => fn(...args)));
}

export function createAction2<T extends AnyFunction>(
  fn: T
): (...args: Parameters<T>) => ReturnType<T> {
  return (...args) => untrack(() => fn(...args));
}
```

thetarnav:
1. it's mainly about ergonomics here.
    If you are calculating the next value based on the previous, it is more clear. And probably slightly more performant.
    But I've seen more complex "setters" where it's much easier to do an early return before going to a setter. E.g. if you want to utilize the path syntax to get to a nested property, but want to check first if the top object is even defined.
    Also do actions always have to be used for setting new state? At least the term `action` seems a bit broader then that.
    I'm down to reverse the order of arguments though. If it were `(setState, state)` you make the default clear without removing the usability.
2. yes the store setter already handles that, but whats stopping you from using two in the same action?
    I know, "just use `batch` in this case", but if we were to make `createAction` and `createActions` a thing, they need to be consistent. So removing `batch` removes the point of `createAction` and `createActions` as standalone primitives, and having them makes it a bit easier to implement your own version of the store. Like I could see myself using that I'm my stores where I'm not even using `createStore` which makes the implicit batch very useful.

In solid 2.0 this whole API can change a lot if store setters were to have `produce` API by default, instead of the path syntax, which might happen.

Zalexios:
> 1. it's mainly about ergonomics here.
>    If you are calculating the next value based on the previous, it is more clear. And probably slightly more performant.
>    But I've seen more complex "setters" where it's much easier to do an early return before going to a setter. E.g. if you want to utilize the path syntax to get to a nested property, but want to check first if the top object is even defined.
>    Also do actions always have to be used for setting new state? At least the term `action` seems a bit broader then that.
>    I'm down to reverse the order of arguments though. If it were `(setState, state)` you make the default clear without removing the usability.
> 2. yes the store setter already handles that, but whats stopping you from using two in the same action?
>    I know, "just use `batch` in this case", but if we were to make `createAction` and `createActions` a thing, they need to be consistent. So removing `batch` removes the point of `createAction` and `createActions` as standalone primitives, and having them makes it a bit easier to implement your own version of the store. Like I could see myself using that I'm my stores where I'm not even using `createStore` which makes the implicit batch very useful.

This all sounds good. I incorporated `(setState, state)` into `createMethods.actions` and created the `createAction` function with `batch` and `untrack`.

I also added the `createStoreFactory` function from your example. It accepts the fallback state and any getters/actions then returns a function to instantiate the store with an optional overridden fallback state for the resulting instance.

The demo was updated using a simple store that can contain a person's age. I use the factory to create 3 instances for 3 sample people, overriding the fallback state with different ages for each person.

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
+import type { JSX } from "solid-js";
+import { batch, untrack } from "solid-js";
+import type { SetStoreFunction } from "solid-js/store";
+import { createStore } from "solid-js/store";
+
+type AnyFunction = (...args: any[]) => any;
+type Action<A extends AnyFunction> = (...args: Parameters<A>) => ReturnType<A>;
+type CreateAction = <A extends AnyFunction>(fn: A) => Action<A>;
+type CreateActions = <A extends object | undefined>(fns?: A) => A | undefined;
```
thetarnav, line 9: This should accept only objects with functions.
If the value is `undefined` or includes something else besides functions, the user should take care of that.
The types in my playground should already be good.
```ts
type AnyFunctionsRecord = { readonly [K in string]: AnyFunction };
type Actions<T extends AnyFunctionsRecord> = { [K in keyof T]: Action<T[K]> };
function createActions<T extends AnyFunctionsRecord>(functions: T): Actions<T>
```

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export const createAction: CreateAction =
+  fn =>
+  (...args) =>
+    batch(() => untrack(() => fn(...args)));
+export const createActions: CreateActions = actions => {
+  if (!actions) return undefined;
```
thetarnav, line 16: Don't call `createActions` if the object is undefined

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  let result = { ...actions };
+  // for (const [name, fn] of Object.entries(result) as [keyof typeof result, AnyFunction][]) {
+  for (const [name, fn] of Object.entries(result)) {
+    if (typeof fn !== "function") continue;
+    result = {
+      ...result,
+      [name]: createAction(fn)
+    };
```
thetarnav, line 25: No point in destructuring the `results` object here. It only makes the operation slower.
```suggestion
    result[name] = createAction(fn)
```

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+  const [state, setState] = createStore<T>(initialState);
+  const getters = createMethods?.getters?.(state);
+  const actions = createActions(createMethods?.actions?.(setState, state));
+
+  return {
+    state,
+    getters,
+    actions,
+    setState
+  };
```
thetarnav, line 140:
```suggestion
  return {
    state,
    getters,
    actions
  };
```
I thinking if the setter is even supposed to be shared by default. It won't push people to read/write segregation if you have the setter always available. all the writes should be ideally covered by actions.
If you really want to have access to a setter, you can make it into an action:
```ts
actions: (setState, state) => ({
  set: setState,
}),
```

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+  initialState: T,
+  createMethods: {
+    actions: (setState: SetStoreFunction<T>, state: T) => A;
+  }
+): SharedStore<T, undefined, A>;
+export function createSharedStore<T extends object>(
+  initialState: T,
+  createMethods?: {}
+): SharedStore<T>;
+export function createSharedStore<
```
thetarnav, line 120: I know that I kept the `createSharedStore` in my examples, but it still doesn't convey the function.
Maybe `createFluxStore`? At least it hints at the use of actions.
I don't really have much better ideas, unfortunately.

packages/store/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+ *   count: () => state.value,
+ *  }),
+ *  actions: (setState, state) => ({
+ *   increment: () => setState(val => ({ ...val, value: val.value + 1 })),
+ *   reset: () => setState("value", 0),
+ *  })
+ * });
+ *
+ * // `Example.tsx`
+ * import { counterStore } from './counter-store.ts';
```
thetarnav, line 84: Separation into multiple files is unnecessary information to add to the documentation in my opinion. It's not like this is the required way to use the primitive.

packages/store/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @solid-primitives/store
+
+## 0.0.100
+
+- Initial commit and **PR**.
```
thetarnav, line 5:
```suggestion
- Initial commit and [**PR**](https://github.com/solidjs-community/solid-primitives/pull/327).
```
Might as well link here 😉

packages/store/dev/components/CounterButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+export default CounterButton;
+
+export const IncreaseButton: ParentComponent<CounterButtonProps> = props =>
+  CounterButton({
+    ...props,
+    onClick: (state, actions) => {
+      actions.resetCount(state.value + 1);
+    },
+    text: "+"
+  });
```
thetarnav, line 37: I know this is only a demo, but I wouldn't add code like this, just because someone might copy it and cause some unwanted side-effects.
In solid there is little difference between `<Component/>` and `Component()`, but one very important detail is that the function call should be untracked to avoid tracking signals read in the component body.
```suggestion
  <CounterButton
    {...props}
    onClick={(state, actions) => {
      actions.resetCount(state.value + 1);
    }}
    text="+"
  />
```

thetarnav (commented):
Alright, a couple of last details and we can proceed to merge it in.
The readme/tests need an update as well. I don't mind taking care of it though.

packages/store/src/index.ts
Zalexios: What about a naming convention building on the existing function name? Since we're using `createStore` to create a store library, what do you think about `createStoreLibrary`? 

Here are those two names with a few last minute ideas:
-   ***createFluxStore***
-   ***createStoreLibrary***
-   createState
-   createStateStore
-   createStateLibrary
-   createStoreDefinition
-   createComposableStore

What do you think?

thetarnav: We take a normal solid store, organize reads to state and getters, and write to actions.
My other idea would be `createImplStore` as the user includes the implementation of state management during creation.

packages/store/src/index.ts
Zalexios: Oh wow I was really over thinking that. Even after seeing your example I thought you meant to incorporate `set` as part of the store creation 😂

packages/store/dev/components/CounterButton.tsx
Zalexios: I had wondered what the difference was between what I had and using `createComponent` but I didn't look it up until just now. Thanks for fixing this.

packages/store/src/index.ts
Zalexios: I addressed your concerns but with a slightly different solution to allow store creation with getters and/or actions while also maintaining types.

```ts
type AnyFunction = (...args: any[]) => any;

type Action = <A extends AnyFunction>(fn: A) => (...args: Parameters<A>) => ReturnType<A>;
type Actions = Record<string, Action>;

export const createAction: Action =
  fn =>
  (...args) =>
    batch(() => untrack(() => fn(...args)));
export const createActions = <A extends Actions>(actions: A) => {
  let result: Actions = { ...actions };
  for (const [name, fn] of Object.entries(result)) {
    result[name] = createAction(fn);
  }

  return result as A;
};
```

Zalexios:
@thetarnav Each of the items has been updated and committed except for the tests but I'll need to get to that tomorrow. If you have time and look it over, let me know if you find or think of anything I should change. Thanks for the help.

Zalexios:
@thetarnav it took a little longer to complete because I identified a few scenarios where type inference wasn't working. Those issues are resolved and I updated tests & demo.

Zalexios:
I pulled the changes and found a few issues with the types. These are what I addressed in [this comment](https://github.com/solidjs-community/solid-primitives/pull/327#issuecomment-1451173258).

1. Cannot use a return type for actions.

Error example A:
```ts
export interface CounterState {
  value: number;
}
export interface CounterGetters {
  isZero: () => boolean;
  count: () => number;
}
export interface CounterActions {
  updateCount: (newValue: number) => void;
  resetCount: (overrideValue?: number) => void;
}
const sharedStore = createFluxStore(
  { value: 0, } as CounterState,
  {
    actions: (setState, state): CounterActions => ({
      updateCount: (newValue) => setState("value", newValue),
      resetCount: (overrideValue) => setState("value", overrideValue ?? 0),
    })
  }
);
```

Error example B:
```ts
const sharedStore = createFluxStore<CounterState, CounterActions>(
  { value: 0, },
  {
    actions: (setState, state) => ({
          updateCount: (newValue) => setState("value", newValue),
          resetCount: (overrideValue) => setState("value", overrideValue ?? 0),
    })
  }
);
```

2. Cannot create a readonly store without actions.

Error example:
```ts
const sharedStore = createFluxStore(
  { value: 0, },
  {
    getters: (state) => ({
      count: () => state.value,
      isZero: () => state.value === 0,
    }),
  }
);
```

3. It is missing an overload to allow specifying the type for the state.

Error example:
```ts
const sharedStore = createFluxStore<CounterState>(...
```

4. Similarly, specifying the type for state and getters fails because there is no overload with state and getters (along with #2 above).

Error example:
```ts
const sharedStore = createFluxStore<CounterState, CounterGetters>(...
```


5. Each of the items above occur with the factory too.

thetarnav:
It needs to be a `type` not an `interface`. To use an `interface` it needs to be transformed with something like this before using it as a return type.
```ts
type Objectify<T> = { [K in keyof T]: T[K] }
```

Maybe they could be improved by being less restrictive on the input, but would be transformed to only include functions.

Also there is no point in creating a store that is missing actions because you cannot write to it. Adding them is the reason why you use this API.

thetarnav:
Alright, I'm gonna merge this in its current state. Thanks, @Zalexios for bringing this here.
This is an interesting pattern to have available and there are probably ways to improve it after we get a chance to use it a bit.
If you have any further ideas feel free to open more issues/PRs :)

PR #332, opened 2023-02-22
storage: fix hydration issue

TomPichaud:
Fixes hydration issue on `createStorageSignal `

### DIFF
--- a/packages/storage/src/storage.ts
+++ b/packages/storage/src/storage.ts
@@ -1,5 +1,7 @@
 import { Accessor, createEffect, createSignal, onCleanup, onMount, Setter } from "solid-js";
 
+import { createHydrateSignal } from "@solid-primitives/utils";
+
 import type {
   AnyStorageProps,
   AsyncStorage,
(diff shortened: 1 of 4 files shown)


thetarnav (approved):

packages/storage/src/storage.ts
```diff
@@ -473,7 +475,11 @@ export function createStorageSignal<T, O = {}>(
       }
       return value;
     }, null);
-  const [accessor, setter] = createSignal<T | null>(read() ?? (initialValue as T), props as any);
+  const [accessor, setter] = createHydrateSignal<T | null>(
+    initialValue as T,
+    () => read() ?? (initialValue as T),
```
atk, line 480: This way, initialValue will also be returned if the signal's value is `null`. This is not the intended behavior.

atk (changes requested):

packages/storage/src/storage.ts
TomPichaud: I disagree, it was implemented this way.

When storage is empty (or SSR): `read()` return `null` therefore it should set the storage.key to the `initialValue` (or return it during SSR).

Then, if you call the `setter` to set the value to `null` it will remove the storage.key

atk: Thanks for the clarification. Sorry, I had only my mobile to check on and there must have been an error with StackBlitz.

atk (approved):

PR #333, opened 2023-02-23
Merge and extend `resolveData` and `resolveTokens`

thetarnav:
I'm coming around to the idea of allowing the user to choose if he wants to include JSX Elements in the resolved array or not.
I've hit this case where I wanted to render something different depending on if the resolved element is an HTMLElement or a token. But the current primitives didn't allow for that flexibility, or you resolve only tokens or do it yourself.

```tsx
// without jsx elements (just the token data)
const tokens = resolveTokens(parser, () => props.children);

createEffect(() => {
  tokens().forEach(token => {
    // token is the data returned by the tokenData function
    console.log(token);
  });
});

// with jsx elements
const els = resolveTokens(parser, () => props.children, true);

createEffect(() => {
  els().forEach(el => {
    // isToken has a practical usecase now
    if (!isToken(parser, el)) {
      // el is a normal JSX Element
      return;
    }
    // token is a function that returns the JSX Element fallback
    // token.data is the data returned by the tokenData function
    console.log(token.data);
  });
});

// the elements can be used in JSX
return <>{els()}</>;
```

@bigmistqke


bigmistqke:
Mm 🤔 don't think I completely get it.

So `resolveTokens(..., true)` will also return elements which aren't tokens?
Why not just `const maybeTokens = children(() => props.children)` and then `isToken(maybeToken)`?

bigmistqke:
It feels like two options bundled in 1 too: 
1. not filtering non-token values
2. returning JSX instead of data

Maybe it should be `{jsx: boolean, returnNonTokensTooOrSomeShorterName: boolean}` instead of `true`?

thetarnav:
> So `resolveTokens(..., true)` will also return elements which aren't tokens?
> Why not just `const maybeTokens = children(() => props.children)` and then `isToken(maybeToken)`?

Currently `resolveTokens` and `resolveData` does pretty much the same thing, they search for tokens in the structure produced by jsx and return them.

If you did `const maybeTokens = children(() => props.children)` all of the tokens would get rendered, and they will fire console warnings. `isToken()` won't be able to help here.
You'd need to manually resolve children, and use `isToken()` to identify token elements there.

So I want to consider this use-case in the next design:
The default behavior of `resolveTokens` will be doing what `resolveData` did - search the jsx structure for tokens and return an array of the data objects. This is what you need most of the time, and it pushes the users to use tokens mainly for passing data, and not for rendering.
But we could pass `true` as a third param (could be `{ addElements: true }` options object as well) to handle both resolved jsx elements and token elements. This is what you use when you consider passing other jsx elements to children for valid use and want to handle rendering.

It is true that I'm bundling two options into one.
But by doing that there are only two combinations. Easy mode - just give me the data; and advanced mode - do the hard work for me, but let me handle the rest.

If I made it into two options, it would produce 4 combinations.
And you can't use `isToken()` to distinguish tokens from other jsx elements if you choose to resolve the data.
But it may be what you want 🤷‍♂️
The options could be:
- `filterJSXElements` for filtering non-token values (default: true)
- `resolveData` for getting the data object instead of token elements (default: true)

bigmistqke:
> If you did const maybeTokens = children(() => props.children) all of the tokens would get rendered, and they will fire console warnings. isToken() won't be able to help here.

True, lol. I should really know this.

>  And you can't use isToken() to distinguish tokens from other jsx elements if you choose to resolve the data.

This is a very valid point. `{ filterJSXElements: false, resolveData: true }` would not provide anything of value, so that's not an option we should be providing, I agree 👍

I think where I am standing right now:
1. either keep `resolveData` and `resolveTokens` as two separate functions. Add a third parameter to `resolveTokens`  for including non-token JSX-Elements. `resolveData` will be used 99% of the time, if you want to go a level deeper there is `resolveTokens`.
1. the same as 1. but remove `resolveData` completely. I think I am rn where you were last PR: `({data})` is really not too bad of a DX and there is value in keeping transparency in the API and limiting the API-surface.

Maybe would be a good to have a third set of eyes go over it, if these concepts are readable from an outsider pov.

PR #341, opened 2023-03-06
new filesystem primitive (WIP)

atk:
This is the first draft of a filesystem primitive. The idea is to make virtual and actual filesystems more accessible. Current to-dos include

* fix bug for nested subdirectories
* implement async access
* add adapters for Web APIs for file system access / webdb
* test tauri adapter


samuelstroschein:
Interested in this primitive, in part, due to this discussion https://github.com/orgs/inlang/discussions/373. 

The API doesn't seem to follow `node:fs`. For example, `writeFile` is called `mkfile`. For simplicity of plugging a node implementation/familiarity, have you thought about obeying to node's fs API?

atk:
I'm still exploring different possible APIs. My current draft was meant as a simplification of it and also tied well into the reactive system.

Since I'm using a subset of this API internally already, it should be simple to explore different ideas.

thetarnav:
Seems like `packages/map/test/index.test.ts` got deleted by accident?

atk:
* Fixed the commit that was mangled by Windows / OneDrive (please pull, had to force-push)
* Currently deciding whether to keep the object API, because it is almost impossible for async
* Need to test the async filesystem APIs
* Want to add Web FIleSystem & Web FileSystem Access API (needs secure env).

samuelstroschein:
@atk why did you choose `[]` for the API and not `readFile`, `readDir` etc?

CleanShot 2023-03-15 at 09 14 53@2x

packages/filesystem/README.md
```diff
@@ -0,0 +1,123 @@
...
+  readFile: (path: string) => Promise<string>;
+  rename?: (previous: string, next: string) => Promise<void>;
+  rm: (path: string) => Promise<void>;
+  writeFile: (path: string, data: string) => Promise<void>;
+};
+```
+
+If no rename method is given, mkdir/writeFile/rm are used to achieve the same effect. An actual rename call might be more performant, though.
+
+The `createFileSystem` call then creates a reactive surface above each of these APIs so that the return values of all reading calls are signals for synchronous and resources for asynchronous filesystem adapters; writing methods for asynchronous APIs will return the same promise for convenience reasons. These getters returned from reading methods are bound to Solid's reactivity so that they will automatically cause effects using them outside of untrack() to re-run on updates. Asynchronous getters will initially return undefined, but will update to the correct value once evaluated.
```
samuelstroschein, line 66: A few remarks: 

1. Are changes in "external" filesystems like node propagated into the reactive system? E.g. someone edits a file with node:fs/promises. Is that change reflected in the reactive fs?
2. Why is a sync API supported? Sync calls will block the main thread. Running fs stuff in a worker is complicated. Did someone _require_ a sync API? If not, remove it to reduce the surface area of this primitive?

atk: Thanks for your answer.

1. No, we don't receive those changes yet - and even if I did, node:fs/promises is not the only source of possible changes to a file system. I intend to add a `registerWatcher`-method at a later point that allows automatic re-evalutation.
2. The sync API is mainly for virtualized file systems like @typescript/vfs, which is what I need it for; cramming them forcefully into an async interface is bad for the ergonomics - I'd be forced to not rely on Suspense for other operations - but you don't need to use it if you don't want it, this primitive is meant to be fully tree-shakable.

thetarnav (commented):
Ok this is some great piece of software :)
The PR is probably good to be merged how it is now - no need to hold this pointlessly.
But there are two concerns I'm thinking about.

The first one is what I've already mentioned on discord:
Becasue the `createFilesystem` API doesn't provide a way to read a file without creating a signal, you are either gonna:
- use the reactive interface and have worst performance and possible memory leaks by overcreating signals
- revert to using native fs API you are targeting - inconsistent DX
- use the object returned my `make___` adapter - risking causing untracked changes to fs by using that object to do writes

I believe that switching to the API proposed below where the `fs.readFile()` could in itself be reactive instead of returning a signal will solve this issue as well.

The second concern is: isn't this a bit too bloated? (and will grow larger in the future as you have more planned features)
I mean it is fine for a package to accumulate features like this, but it's more of a concern from the perspective of this being a part of solid-primitives. It feels more like a standalone library than a single reusable primitive.
All of the exports are a part of one system instead of separate primitives that could be used on it's own.
image
I feel like this deserves a reconsideration - what should or shouldn't be a part of this library. Just like previously we wanted to include a `/ui` package with headless components but realized that we are making a library inside of a library and made a separate repo for it.
And I don't mean that all of this should be removed - there are probably pieces that fit well to the "composable primitive" category and they should be kept and maintained here (Just like @solid-primitives/transition-group and solid-transition-group providing respectively lower and higher-level API), but I feel like we might be streaching that a bit here.

Well anyway, those are just my thoughts. I think it's fine to merge this now and reconsider the design later again.

packages/filesystem/src/reactive.ts
```diff
@@ -0,0 +1,242 @@
...
+      readdirMap.get(getParentDir(path))?.[1](items => [...items, path] as [string, ...string[]]);
+      fs.readdir(path)();
+    },
+    readFile: path => {
+      if (!readFileMap.has(path)) {
+        readFileMap.set(path, createSignal(""));
+      }
+      const [data, setData] = readFileMap.get(path)!;
+      setData(adapter.readFile(path));
+      return data;
```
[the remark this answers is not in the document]
atk: I had originally intended such an API, but lost the thought along the way. You're right, the DX is definitely better for that one.

packages/filesystem/src/reactive.ts
```diff
@@ -0,0 +1,242 @@
...
+          .get(getParentDir(path))?.[1]
+          .mutate(items =>
+            items.includes(name) ? items : ([...items, name] as [string, ...string[]]),
+          );
+      }),
+    rename: async (previous, next) => {
+      const previousType = await adapter.getType(previous);
+      const nextType = await adapter.getType(next);
+      if (previousType === "dir" && nextType === "file") {
+        throw new Error(`Cannot overwrite file "${next}" with directory "${previous}"`);
```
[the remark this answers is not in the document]
atk: I don't think hiding errors from the user is a good idea in that case. On the contrary, I believe useful error messages are a worthwhile feature even in production, because this is something that may fail there as well and you need to be aware to recover gracefully.

packages/filesystem/README.md
```diff
@@ -0,0 +1,132 @@
...
+// will be called again if the contents of "/" change due to
+// writing a new file or deleting an existing file or directory
+createEffect(() => console.log("/", fs.readdir("/")));
+// reading files
+// this signal (or resource for async adapters) will update if
+// the file is written by the same fs
+createEffect(() => console.log("/src/index.ts", fs.readFile("/src/index.ts")));
+
+// isomorphic file system reader with lazy evaluation
+const Item = (props: { path: string, fs: SyncFileSystem | AsyncFileSystem }) => (
```
thetarnav, line 97: The component code is wrapped with `()` instead of `{}`

thetarnav (approved):
Alright, my previous issue got resolved partially - the API feels a lot better. Although there is still an issue with creating signals/resources when reading without a listener.
Also, I still believe that this should be moved to its own library, it's just too big to be considered the same as the rest of the packages. What are your thoughts on this @atk? What's the point in keeping it here, instead of making it a separate library that could grow freely without the restriction of being just a primitive?
The `/fetch` package is in a similar situation tbh, it's less of a reusable primitive that solves a specific problem and more of an opinionated API extending solid resources with solution common needs. But those solutions cannot be used on their own - you are locked into the createFetch interface.

atk:
> Although there is still an issue with creating signals/resources when reading without a listener.

I had to weight the performance advantage of not creating a listener against the penalty of checking for a listener on every call. Since you can still use the adapters outside of the reactive part, I thought that was the better choice.

> Also, I still believe that this should be moved to its own library, it's just too big to be considered the same as the rest of the packages. What are your thoughts on this @atk?

Our graphql package is bigger than this filesystem package. I don't think the size is an issue.

> What's the point in keeping it here, instead of making it a separate library that could grow freely without the restriction of being just a primitive?

The signal/resource creation? That's a minuscule part of the whole library - and extracting it doesn't reduce the overall size significantly (or at all unless it is re-used). For the time being, I wanted something that just worked without too many external dependencies.

>  The `/fetch` package is in a similar situation tbh, it's less of a reusable primitive that solves a specific problem and more of an opinionated API extending solid resources with solution common needs. But those solutions cannot be used on their own - you are locked into the createFetch interface.

Also, I don't see many parts of this package that could be reused for a different cause. File system access is a very specific topic. I guess that the fetch package could work as a set of createResource-Extensions that you manually put together instead of the automatism that I created for the package for the sake of better DX. But that is a different topic.

thetarnav:
> Our graphql package is bigger than this filesystem package.

Is this true? the qraphql package is a single file with 200 LOC with half of it being commented when the filesystem one is split into 10 dense files. I mean I agree that size in itself doesn't mean anything, but it hints at the complexity. graphql also exports multiple standalone primitives where each could be used on its own, not a single API with the rest of the exports serving only as adapters.

> I guess that the fetch package could work as a set of createResource-Extensions that you manually put together instead of the automatism that I created for the package for the sake of better DX. But that is a different topic.

This might be it, the topic that I'm after 😄 Not sure if something like this could be done for the filesystem package, but it is the design choice that I'm starting to question recently. I think we should be favoring low-level and composable approaches to provided APIs, so that each could be used in any project, without being locked with the rest of the package. And more opinionated, all-inclusive APIs could be created on top of these primitives in userland or in separate libraries focusing on that.
If you have to do something manually to fill the gaps when using our primitives - fine. But when you cannot use it because its scope is too big and your use-case is not considered - that's an issue.
Just like solid provides createResource and Suspense and solid-query fill in the gaps with a developer-friendly API.
Or we provide the low-level transition-group primitives and motionone/solid and solid-transition-group can be built further abstractions on top of them.
In the same way, I would love to see the filesystem package exporting multiple primitives solving specific fs related tasks, and a different library using them to create an all-inclusive and complete abstraction for the filesystem in solid.

But yeah, I don't know how applicable this is to the filesystem. I just don't want to have a situation where someone may not want to use some of our packages because they are too big (conceptually) or too opinionated, but on the other hand not opinionated enough for those who look for a complete solution.

atk:
The only thing that I might see extracted from this into a useful separate, reusable logic is the tree-walking stuff that is re-used a few times in different flavors, but I wanted to avoid hasty abstractions.

samuelstroschein:
FYI About this package being too large: We will likely provide a reactive filesystem as part of inlang that can be used anywhere. It might not be worth the maintenance effort for Solid to provide its own (relatively complete) filesystem implementation.

But, super early. Can't provide a timeline yet.

atk:
@samuelstroschein:

Since the reactivity needs to be tailored to the framework (or you use solid.js internally and create an adapter for other frameworks), this is hardly an advantage for us and your package won't be much smaller, lest it removes features, which due to tree-shaking cannot yield any savings.

I'm not yet 100% sure if I will make this a separate package with support for different frameworks at some point, but for the time being, I'm happy with having it here.

Lastly, if you want to use parts of my code in your project, you hereby have my explicit permission to do so.

samuelstroschein:
@atk 

> Lastly, if you want to use parts of my code in your project, you hereby have my explicit permission to do so. 

Thank you! Much appreciated. 

> I'm not yet 100% sure if I will make this a separate package with support for different frameworks at some point, but for the time being, I'm happy with having it here. 

Yes makes sense. I just dropped the note that if you use the filesystem primitive for the new docs only, exposing the filesystem to external users might not be worth the maintenance effort.

samuelstroschein:
(In case you are curious) Some first API drafts for an fs implementation we need popped up https://github.com/inlang/inlang/pull/455/files#diff-e270c392bc0bb35f13e1aa01b6a228749ac79cac3e44ecccb7f48919679b39cdR36-R54

PR #347, opened 2023-03-11
New script, update packages' package.json primitive list

aquaductape:
Added new script `update-packages-package-json`, to update packages' package.json primitive list. Currently there's a lot of packages that have outdated `primitive.list` in package.json file, such as having primitives that have been removed/renamed or added new primitives. This will also help maintain updated primitives table in README file on github and on live site since they depend on `primitive.list` in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "test": "pnpm run test:client && pnpm run test:ssr",
     "pages": "turbo run page && jiti ./scripts/collectPages.ts",
     "new-package": "jiti ./scripts/new-package.ts",
+    "update-packages-package-json": "jiti ./scripts/update-packages-package-json.ts && prettier -w packages/**/package.json",
     "update-readme": "jiti ./scripts/update-readme.ts",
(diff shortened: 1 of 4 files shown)


thetarnav:
I know that it's tempting to just automate it, but there is a reason for managing the list manually.
Not every export is worth adding to the primitive's list and displaying in the table.
Just like signal-builders have the list collapsed, the other packages also have exports that are "ignored" intentionally.
If some list is outdated as you said, it should be fixed manually.

But I don't mind having this script as a way to get additional information that can be used on the website. The site could have an additional section where all exports are listed - why not.

atk:
> I know that it's tempting to just automate it, but there is a reason for managing the list manually.

We could have an optional comment at the top of the file which primitives are supposed to be put into the list.

aquaductape:
Closing PR, as `package.json`s' primitive list shouldn't be modified for the site needs. This script has been removed, but is modified and used internally in [scripts/update-site/build-package-data.ts](https://github.com/solidjs-community/solid-primitives/blob/678528919e997d6a1228361976cfb8b8cb5f39ff/scripts/update-site/build-package-data.ts) to get most updated primitive list for the `update-site` script

PR #350, opened 2023-03-13
feat: deep effect function

iosamuel:
```typescript
createEffect(
    createDeep(
        () => store,
        () => { /* function to execute */ }
    )
);
```

Effect Function to execute effect when deep properties on a store change.

### DIFF
--- a/packages/deep/package.json
+++ b/packages/deep/package.json
@@ -1,0 +1,71 @@
+{
+  "name": "@solid-primitives/deep",
+  "version": "0.0.100",
+  "description": "Create a Deep effect for Stores.",
+  "author": "Samuel Burbano <me@iosamuel.dev>",
+  "contributors": [],
+  "license": "MIT",
+  "homepage": "https://github.com/solidjs-community/solid-primitives/tree/main/packages/deep#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/solidjs-community/solid-primitives.git"
+  },
+  "bugs": {
+    "url": "https://github.com/solidjs-community/solid-primitives/issues"
+  },
+  "primitive": {
+    "name": "deep",
+    "stage": 0,
+    "list": [
+      "createDeep"
+    ],
+    "category": "Reactivity"
+  },
+  "keywords": [
+    "solid",
+    "primitives"
+  ],
+  "private": false,
+  "sideEffects": false,
+  "files": [
+    "dist"
+  ],
+  "type": "module",
+  "main": "./dist/server.cjs",
+  "module": "./dist/server.js",
+  "browser": {
+    "./dist/server.cjs": "./dist/index.cjs",
+    "./dist/server.js": "./dist/index.js"
+  },
+  "types": "./dist/index.d.ts",
+  "exports": {
+    "worker": {
+      "import": "./dist/server.js",
+      "require": "./dist/server.cjs"
+    },
+    "browser": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    },
+    "deno": {
+      "import": "./dist/server.js",
+      "require": "./dist/server.cjs"
+    },
+    "node": {
+      "import": "./dist/server.js",
+      "require": "./dist/server.cjs"
+    },
+    "import": "./dist/index.js",
+    "require": "./dist/index.cjs"
+  },
+  "scripts": {
+    "dev": "vite serve dev",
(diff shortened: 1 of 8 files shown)


packages/deep/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+ *         () => { ... }
+ *     )
+ * );
+ * ```
+ */
+export function createDeep<S, Next extends Prev, Prev = Next>(
+  dep: Accessor<S>,
+  fn: OnEffectFunction<S, undefined | NoInfer<Prev>, Next>,
+): EffectFunction<undefined | NoInfer<Next>, NoInfer<Next>>;
+export function createDeep<S, Next extends Prev, Prev = Next>(
```
thetarnav, line 28: What do you think about implementing it in a composable way?
I mean so that it could be used together with `on` (rather than instead of it), [`defer`](https://github.com/solidjs-community/solid-primitives/blob/main/packages/utils/src/index.ts#L162), or with any computation without a wrapper for stating explicit dependencies. I mean it's just a way of deep-tracking the store. Better to do as little as possible and let the user compose what he wants.

So `deep` or `trackDeep` may be a better name. `create___` would indicate that you are instantiating something reactive.

```ts
createEffect(() => {
  trackDeep(store)
  return untrack(() => { /*...*/ })
})

createEffect(on(
  [() => trackDeep(store), anotherSource],
  ([store, anotherValue]) => { /*...*/ },
  { defer: true }
))

import { defer } from "@solid-primitives/utils"
createEffect(defer(
  () => trackDeep(store),
  (store) => { /*...*/ },
))
```

iosamuel:
@thetarnav I updated the lockfile, I think that was the problem with the CI process.

packages/deep/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@solid-primitives/deep",
```
thetarnav, line 2: Wondering if the package shouldn't just be called `@solid-primitives/watch` / `@solid-primitives/on` / `@solid-primitives/track` and provide utilities for tracking / limiting sources.
Such as `defer` from utils or `on` from solid (there is an idea to move it to solid-primitives)

But it's fine for now - that's something to figure out in the future.

packages/deep/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      deepTraverse(value[i]);
+    }
+  } else if (typeof value === "object" && value !== null) {
+    for (const key in value) {
+      deepTraverse(value[key]);
+    }
+  }
+  return value;
```
thetarnav, line 62: Maybe there could be two primitives.
`trackDeep` and `trackStore`
One for accessing each object/array property as you do above.
And the other make specifically for stores.
You might look into [`isWrappable`](https://github.com/solidjs/solid/blob/77373cc455a7c6dc8275297a08256afa5594a344/packages/solid/store/src/store.ts#L69) or `unwrap`/`wrap` to see for stores are wrapped. This way you could avoid traversing values that are not stores.

thetarnav (commented):
Ok, this is great.
I would be leaning toward a more composable design, but it's looking good otherwise.

packages/deep/src/index.ts
iosamuel: Loved the idea, much simpler and powerful implementation.

iosamuel:
@thetarnav applied changes.

packages/deep/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+function deepTraverse<T>(value: Store<T>): T {
+  if (!isWrappable(value)) {
+    return value;
+  }
+
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      deepTraverse(value[i]);
+    }
+  } else if (isWrappable(value)) {
```
thetarnav, line 45: Is it necessary to check if the value is wrappable twice?
Also `isWrappable` also does `Array.isArray` check, so it could be removed either from here or from `isWrappable`.

iosamuel: Oops, yeah, sorry, It should be good now!

iosamuel:
@thetarnav I updated the README as well to include an example of a more composable way of doing things. Should give a good idea of how to use it now. Thanks for all the back and forth on this! 🔥

thetarnav (approved):

PR #352, opened 2023-03-14
Improve and rename `jsx-tokenizer`

thetarnav:
- Package got renamed from `jsx-parser` to `jsx-tokenizer`.
- Improved jsdoc and readme with better descriptions and examples.
- `createJSXParser` renamed to `createTokenizer`.
- `isToken` and `resolveTokens` can now accept an array of tokenizers to match.
- `createToken` can be used without passing a tokenizer. This will create a token with the component function as the tokenizer.

```tsx
type TabProps<T> = { value: T };

function Tabs<T>(props: {
  children: (Tab: Component<TabProps<T>>) => JSX.Element;
  active: T;
  onSelect: (item: T) => void;
}) {
  const Tab = createToken((props: TabProps<T>) => props.value);
  // resolveTokens will look for tokens created by Tab component
  const tokens = resolveTokens(Tab, () => props.children(Tab));

  return (
    <ul>
      <For each={tokens()}>
        {token => (
          <li
            onClick={() => props.onSelect(token.data)}
            classList={{ active: token.data === props.active }}
          >
            {token.data}
          </li>
        )}
      </For>
    </ul>
  );
}

// usage
<Tabs active="tab1" onSelect={tab => console.log(tab)}>
  {Tab => (
    <>
      <Tab value="tab1" />
      <Tab value="tab2" />
    </>
  )}
</Tabs>;
```

[README](https://github.com/solidjs-community/solid-primitives/tree/jsx-tokenizer/packages/jsx-tokenizer#readme)


bigmistqke:
Oo very nice additions. 

The name-change is great. It was bothering me too.

I really like the `createToken` without `tokenizer`, that's really smart. Handy if you are just having one token, but also with multiple, now that `resolveTokens` and `isToken` can take arrays. 

Sort of allows to reverse the process all together: instead of having 1 `tokenizer` out of which you make tokens, you could make individual tokens and then bundle them together for `resolveTokens` and `isToken`. And then you could mix-and-match to make custom `tokenizers` too if needed. Cool stuff.

thetarnav:
Thanks, I like it a lot too.
It's interesting how your custom-renderer usecases mainly benefit from having a single tokenizer and many tokens created from it. And when used as a composition/control-flow pattern like `<Routes>` and `<Route>` you mostly only want to have a single resolver-token pair.

PR #358, opened 2023-03-17
i18n: add custom readFn function

framp:
Hello!

This PR adds support to the i18n module for specifying a custom `readFn` function which is used in place of `deepReadObject`. The default behaviour when reading the store is to consider keys with dots as wanting to traverse nested objects in the store.

This breaks a series of use-cases: 
 - "some.namespace.key.intro_text": "Welcome!"
 - "I like to have the keys being the english version of a string, there is a negligible perf hit but the code is more readable": "Mi piace che le chiavi siano la versione inglese di una stringa, c'è un calo di prestazioni ma il codice è più leggibile"
 
An anecdotal experience:
As a user who quickly skimmed through the examples, I was not expecting this behaviour and I imagined it would just try to `dict[key]` and I was puzzled when I didn't see my translations appearing
 
I think adding a custom function allows us to keep retro-compatibilty and give the users the chance to provide an alternative implementation for accessing the store.

EDIT: Thanks for the reviews and apologies, I will be very busy until June but I'll try to come back to it asap! 🙌

### DIFF
--- a/.changeset/good-trees-wait.md
+++ b/.changeset/good-trees-wait.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/i18n": patch
+---
+
+Add an option to provide a custom function for resolving translation values.
--- a/packages/i18n/src/chained-i18n.ts
+++ b/packages/i18n/src/chained-i18n.ts
@@ -12,10 +12,10 @@ type I18nFormatArgs = Record<string, string | number>;
 
 export type I18nPath<T extends I18nObject> = {
   [K in keyof T]: T[K] extends I18nObject
-    ? I18nPath<T[K]>
-    : T[K] extends (options: infer OptionsArgs) => string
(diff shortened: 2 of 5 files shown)


thetarnav (commented):
> I think adding a custom function allows us to keep retro-compatibilty and give the users the chance to provide an alternative implementation for accessing the store.

I love that. It would be great if we could push the design of i18n primitives into more composable and by extend - customizable.

thetarnav:
The `i18n` package was reworked here: https://github.com/solidjs-community/solid-primitives/pull/510
Now keys with dot delimiters are valid. So i’m gonna close this, but feel free to propose more changes to the new i18n version.

framp:
Oh, damn, my bad

I think I missed your original notification and miss the review🤦 sorry!

And congrats on the rework, glad to hear this was addressed

PR #363, opened 2023-03-25
[site]: Generate data not source code

thetarnav:
A bigger PR mainly targetting the way the site is currently getting generated - by creating `.tsx` files in `/routes` which then get turned into `.html` files during the actual build.
This changes the generation step to create only `.json` files that are being used as api-endpoints of sorts with which you fetch the data required to render the page.

haha still deleted more then added 🚀

image


aquaductape:
Beautiful

aquaductape:
For https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-35ac96e21c02642fbc0b8e07e7a59c676a741bd8c39a8b61fcf1f0423fdcd40c, Needs to create `_generated/packages` directories inside `site/src/` if they don't exist

aquaductape:
For https://github.com/solidjs-community/solid-primitives/pull/363/commits/65d19b5b0f27aedf8860ff36edd8002bc5f1988b,

Package route needs a script tag to add styling when that route is visited via network instead of client-side.

Without it, the styling declared inside the component, even without the onMount, fires during Hydration, leaving a couple milliseconds of not seeing that style.
```js
const Page: Component = () => {

  if (!isServer) {
    document.documentElement.classList.add("primitives-page-main");
    onCleanup(() => {
      document.documentElement.classList.remove("primitives-page-main");
    });
  }
 // ...
 ```

**Comment by aquaductape** [comment]
Why remove the `make` information for tooltip https://github.com/solidjs-community/solid-primitives/pull/363/commits/a328b365250b815b1101a90dbc8b131dd7293e39#diff-6af8524ba3ef37ac5e00cad409a06b4adc895f6ef252e31cea5a6e0af74e059aL56-L66.

This benefits new users, not understanding the `make` prefix, there are plenty of `README`s where they don't include description of `make` vs `create`.

**Comment by thetarnav** [comment]
Thanks for the review.
You might want to use the pr review feature if you are pointing to specific parts of the code 😅

> For https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-35ac96e21c02642fbc0b8e07e7a59c676a741bd8c39a8b61fcf1f0423fdcd40c, Needs to create `_generated/packages` directories inside `site/src/` if they don't exist

I think I'm doing it here: https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-35ac96e21c02642fbc0b8e07e7a59c676a741bd8c39a8b61fcf1f0423fdcd40cR152

> Package route needs a script tag to add styling when that route is visited via network instead of client-side.
> 
> Without it, the styling declared inside the component, even without the onMount, fires during Hydration, leaving a couple milliseconds of not seeing that style.

Alright, got it. I was wondering why the class was added in two places before, but then I saw that the 404 page doing it like this, so I've just copied it:
https://github.com/solidjs-community/solid-primitives/blob/65d19b5b0f27aedf8860ff36edd8002bc5f1988b/site/src/routes/%5B...404%5D.tsx#LL8C19-L8C19

I think I'll just abstract it into a separate primitive.


> Why remove the `make` information for tooltip [a328b36#diff-6af8524ba3ef37ac5e00cad409a06b4adc895f6ef252e31cea5a6e0af74e059aL56-L66](https://github.com/solidjs-community/solid-primitives/commit/a328b365250b815b1101a90dbc8b131dd7293e39#diff-6af8524ba3ef37ac5e00cad409a06b4adc895f6ef252e31cea5a6e0af74e059aL56-L66).
> 
> This benefits new users, not understanding the `make` prefix, there are plenty of `README`s where they don't include description of `make` vs `create`.

That's fair. I wanted to simplify it, as there only point of this naming choice is to separate reactive primitives (`create___` & `use___`), from your usual, vanilla-js ones (`make___`, `get___`, etc.). And this is the distinction worth emphasizing.
But I agree that after the change, the tooltip became a little less helpful — it could tell a bit more about the meaning behind that specific prefix:
- `get___` - pure function calculating the result based on the params (implicit and explicit ones).
- `make___` - function instantiating a subject from the name (can have side-effects)
- `create___` - a function creating a new instance of a reactive subject (can hook into existing reactivity graph for scheduling effects or disposal)
- `use___` - same as `create___` but it can reuse or connect to an existing reactive subject through context or other means (placement in the reactive graph matters more than with `create___`)

File: site/src/routes/package/[name]/components/primitive-name-tooltip.tsx [hunk 1]
```diff
@@ -1,16 +1,54 @@
...
 import { Content } from "tippy.js";
 import { BASE } from "~/constants";
 import { BundlesizeItem } from "~/types";
 
-function createTooltipContent(
-  el: HTMLElement,
-  data: BundlesizeItem,
-  type: ReturnType<typeof getTypeOfPrimitive>,
-): Content {
+export type PrimitiveType = "create" | "use" | "make" | "get" | "component" | "utility";
```
Comment (aquaductape) [inline site/src/routes/package/[name]/components/primitive-name-tooltip.tsx:9]: Love it!!

> That's fair. I wanted to simplify it, as there only point of this naming choice is to separate reactive primitives (`create___` & `use___`), from your usual, vanilla-js ones (`make___`, `get___`, etc.). And this is the distinction worth emphasizing.
> But I agree that after the change, the tooltip became a little less helpful — it could tell a bit more about the meaning behind that specific prefix, or link to the more detailed explanation:
> - `get___` - pure function calculating the result based on the params (implicit and explicit ones).
> - `make___` - function instantiating a subject from the name (can have side-effects)
> - `create___` - a function creating a new instance of a reactive subject (can hook into existing reactivity graph for scheduling effects or disposal)
> - `use___` - same as `create___` but it can reuse or connect to an existing reactive subject through context or other means (placement in the reactive graph matters more than with `create___`)
> - \+ a component

It would be cool to see these longer explanations, with least amount of text since tooltips should be small and simple IMO, but I'm good with the current changes 👍

site/src/routes/package/[name]/components/primitive-name-tooltip.tsx
Comment (aquaductape) [inline site/src/routes/package/[name]/components/primitive-name-tooltip.tsx:1]: Code tag that has primitive name textcontent also exists by itself, so this query, `"h2 code, a code"`, wouldn't be able to find it. That's why code element that contains primitive name should have data attribute to specify that this code element is a primitive code element, since other code elements contain other content such as code blocks. So that's why I added this to modify primitive code element to include attribute https://github.com/solidjs-community/solid-primitives/pull/363/commits/8ffc84f848bc011cd8bddd5608a9c4b086a7867f#diff-35ac96e21c02642fbc0b8e07e7a59c676a741bd8c39a8b61fcf1f0423fdcd40cR117-R123

site/src/routes/package/[name]/components/primitive-name-tooltip.tsx
```diff
@@ -219,7 +219,7 @@ export function createPrimitiveNameTooltips(props: {
   primitives: BundlesizeItem[];
 }): void {
   onMount(() => {
-    const els = props.target.querySelectorAll<HTMLElement>("h2 code, a code");
+    const els = props.target.querySelectorAll<HTMLElement>("[data-code-primitive-name]");
```
Comment (aquaductape) [inline site/src/routes/package/[name]/components/primitive-name-tooltip.tsx]: Code tag that has primitive name textcontent also exists by itself, so this query, `"h2 code, a code"`, wouldn't be able to find it. That's why code element that contains primitive name should have data attribute to specify that this code element is a primitive code element, since other code elements contain other content such as code blocks. So that's why I added this to modify primitive code element to include attribute https://github.com/solidjs-community/solid-primitives/pull/363/commits/8ffc84f848bc011cd8bddd5608a9c4b086a7867f#diff-35ac96e21c02642fbc0b8e07e7a59c676a741bd8c39a8b61fcf1f0423fdcd40cR117-R123

Comment (aquaductape) [reply site/src/routes/package/[name]/components/primitive-name-tooltip.tsx]: Edit: updated the regex here https://github.com/solidjs-community/solid-primitives/pull/363/commits/ab8b58b09bf6370c2be6d2833eaeb42c03729d91

scripts/get-modules-data.ts
```diff
@@ -0,0 +1,76 @@
...
+        description: pkg.description ?? "",
+        tags: pkg.keywords ?? [],
+        category: pkg.primitive.category ?? "Misc",
+        stage: pkg.primitive.stage ?? 0,
+        primitives: pkg.primitive.list ?? [],
+        localDependencies,
+        peerDependencies,
+      }),
+    };
+  });
```
aquaductape, line 73: Tags such as "primitive" and "solid", are not needed, adds noise to search function results in UI.

Should be updated with 
```js
    const excludeKeywords = ["primitive", "solid", pkg.name];
    const tags = (pkg.keywords ?? []).filter(keyword => !excludeKeywords.includes(keyword));

    return {
      data: await mapFn({
        name,
        version: pkg.version ?? "0.0.0",
        description: pkg.description ?? "",
        tags,
        category: pkg.primitive.category ?? "Misc",
        stage: pkg.primitive.stage ?? 0,
        primitives: pkg.primitive.list ?? [],
        localDependencies,
        peerDependencies,
      }),
    };
```

scripts/get-modules-data.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    return {
+      data: await mapFn({
+        name,
+        version: pkg.version ?? "0.0.0",
+        description: pkg.description ?? "",
+        tags: pkg.keywords ?? [],
+        category: pkg.primitive.category ?? "Misc",
+        stage: pkg.primitive.stage ?? 0,
+        primitives: pkg.primitive.list ?? [],
```
aquaductape, line 68: You're getting primitive list from package.json file, it contains missing  and removed primitives. Shouldn't we get primitives from source like here https://github.com/solidjs-community/solid-primitives/blob/main/scripts/update-site/get-package-data.ts#L82-L144

site/src/routes/package/[name]/components/primitive-name-tooltip.tsx
Comment (thetarnav) [reply site/src/routes/package/[name]/components/primitive-name-tooltip.tsx]: imo better to just target all `code` blocks. if it matches it will create a tooltip.
but it's fine let's just roll with this

scripts/get-modules-data.ts
thetarnav: The list in package.json is the source of truth of what are the main primitives for given package. Only those should be displayed in the primitives table.
If a list contains missing primitives, it could just be fixed. And if it misses others, it may be on purpose.

site/src/routes/package/[name]/components/primitive-name-tooltip.tsx
Comment (aquaductape) [reply site/src/routes/package/[name]/components/primitive-name-tooltip.tsx]: I'd rather make the route page a tiny bit larger by adding those attributes, then add more runtime on page load, i'm definitely overestimating but getting textContent on those potential multiple large code blocks could be bad

Comment (aquaductape) [reply site/src/routes/package/[name]/components/primitive-name-tooltip.tsx]: Actually I double down, yeah the execution time of textContent on a code element that contains a code block will contain nested spans and converting those to a string is bad

scripts/get-modules-data.ts
aquaductape: > The list in package.json is the source of truth of what are the main primitives for given package. Only those should be displayed in the primitives table just like it works on the main readme.

I agree with primitives table just showing main primitives.

> But I was thinking about the possibility of adding the rest of the exports to a additionalExports field, so that they could be used by tooltips, or displayed in some way in the package page in the future.

I would like additionalExports field to be added, because I found so many helpful primitives that the README Table didn't provide, when I was building/testing the Search function. It would be nice for users to be able to discover non-main primitives. If it's too much noise the Search function, I could put it behind a toggle.

mohanavel15:
Few files have commented code, are they supposed to be like that?

- https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-4ac618cf5e0086efa16d623c0a4a4b55ce995495489d2d17068cdef17c18a650R85-R98
- https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-2b21b670ccccfc97998c06a6c722a5666826ee4cba03890e20959afbea5580e7R6-R20
- https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-f164b785903ea9841af8de3d7d2305a9402ba12eee130b750f6dac3cb20f8331R12-R42
- https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-54f9a30d7a0448a7f1e1758b5710eb4523feef68dc573d67db5e73e744becdf6R33-R35
- https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-c2ea4b5863c4324561039c25f2370734e803d621dc447b7e4952151137467f03R67-R71

aquaductape:
@mohanavel15, I guess it's just a bad habit of me and @thetarnav, commenting out because we think we need it because we're quickly prototyping stuff then forgetting about it about later. I'll removed all of them except this one https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-4ac618cf5e0086efa16d623c0a4a4b55ce995495489d2d17068cdef17c18a650R85-R98, I'll let @thetarnav decide that one.

site/src/components/Primitives/SizeBadge.tsx
```diff
@@ -1,43 +1,39 @@
...
+  exportName?: string;
+  peerDependencies: string[];
+}) => {
+  const query = encodeURIComponent(`@solid-primitives/${packageName}`);
+  const config = encodeURIComponent(
+    `{"esbuild":${JSON.stringify({ external: peerDependencies })}}`,
+  );
+  return `https://bundlejs.com/?q=${query}${
+    exportName ? `&treeshake=${encodeURIComponent(`[{${exportName}}]`)}` : ""
+  }&config=${config}`;
```
aquaductape, line 21: This runs unnecessary on client side, since the href has been been set on static build.

Maybe wrap `NoHydration` on `<SizeBadge>` here https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-c2ea4b5863c4324561039c25f2370734e803d621dc447b7e4952151137467f03R124-R129 ?

```js
<Table.TD>
  {pkg.primitives.map(primitive => (
    <SizeBadgeWrapper primitiveName={`${pkg.name}_${primitive.name}`}>
      <NoHydration>
        <SizeBadge
          value={primitive.gzip}
          packageName={pkg.name}
          exportName={primitive.name}
          peerDependencies={pkg.peerDependencies}
        />
      </NoHydration>
    </SizeBadgeWrapper>
  ))}
</Table.TD>
```

thetarnav: Theoretically the whole table could be wrapped with no hydration, right?
Maybe not all but right now that content doesn't change.
I'll try experimenting with that.

thetarnav:
> @mohanavel15, I guess it's just a bad habit of me and @thetarnav, commenting out because we use it as reference or we think we need it later because we're quickly prototyping stuff then forgetting about it about later. I'll removed all of them except this one https://github.com/solidjs-community/solid-primitives/pull/363/files#diff-4ac618cf5e0086efa16d623c0a4a4b55ce995495489d2d17068cdef17c18a650R85-R98, because I actually need that code block for checking bundle size of every single primitive, but since we're planning to use `additionalExports` field from primitives package.json, there's no need, but I'll let @thetarnav decide that one.

Yeah, I have no idea what that one did, but the code seemed to work without it so I just left it for reference.
And about additionalExports, which maybe I wasn't clear about - I want them to be generated by statically analyzing the entry files, or running them to get exports, or any other automatic way. I just want to separate the `primitives` key - manually updated, used in the root table, toc, bundlesize overview, etc. and `additionalExports` - automatically generated, used as additional keywords for search, tooltips, maybe displayed in the package pages later.
I don't think additional exports should have their bundle size checked anyway so that commented code could probably be deleted.

site/src/components/Primitives/SizeBadge.tsx
aquaductape: Not the whole table, because `<SizeBadgeWrapper>` has interactivity https://github.com/solidjs-community/solid-primitives/blob/generate-pages/site/src/components/Primitives/SizeBadge.tsx#L47-L56.

thetarnav: `<NoHydration>` and `<Hydration>` together?

aquaductape (commented):
Resolves https://github.com/solidjs-community/solid-primitives/pull/363#issuecomment-1483950952

aquaductape:
Finally https://github.com/solidjs-community/solid-primitives/pull/363/commits/a76e3f08ecda630c904d27621a568a304c8276b7, it was the dang overlay elements that were causing those 300+ missing hydration key warnings even though they only mount via interaction.

Screenshot 2023-03-31 at 1 53 00 AM

aquaductape:
Stage Modal crashes, I think due to recent `1.7.0` change on removing auxilary closing tags. https://github.com/solidjs-community/solid-primitives/blob/a76e3f08ecda630c904d27621a568a304c8276b7/site/src/components/Stage/Stage.tsx#L10

site/src/components/Primitives/SizeBadge.tsx
aquaductape: Like nesting `<Hydation>` inside `<NoHydration>`? 
```js
<NoHydration>
  <Table.Table>
    // ...
    <tbody>
      {Object.entries(categories).map(([category, packages]) => (
        <>
         // ...
          {packages.map(pkg => (
              // ...
              <Hydration>
                <Table.TD>
                  <StageBadge value={pkg.stage} />
                </Table.TD>
              </Hydration>
          )}
        </>
      ))}
    </tbody>
  </Table.Table>
</NoHydration>

```

Doesn't work I tried it, even though it does add `data-hk` to `<Table.TD>` which is just a `td` element, but not on `<StageBadge>`'s `button` element.

aquaductape:
> opening Stage content crashes, I think due to recent solidjs `1.7.0` update on removing auxilary closing tags.
> 
> https://github.com/solidjs-community/solid-primitives/blob/a76e3f08ecda630c904d27621a568a304c8276b7/site/src/components/Stage/Stage.tsx#L10

updating `pnpm-lock.yaml` with ryan's dom-expressions to latest version `0.36.4` fixes this issue

thetarnav:
Deploy preview: https://app.netlify.com/sites/solid-primitives/deploys/642c80e9b61e9b3ed07bf178

thetarnav:
@aquaductape I've removed `dotenv` and just used the `VITE_SITE_URL` env. Also the `.env` file is now public (secrets can be kept in `.env.local` if needed) so no need for setting envs on netlify.

image

aquaductape:
> Also the `.env` file is now public (secrets can be kept in `.env.local` if needed) so no need for setting envs on netlify.

Since I'm used to Next.js I thought that we should name commited `.env` file as `.env.public`. But vite doesn't support this out of box and their documentation suggests that `.env` name is public by default, while `.env.local`. 

So I support the commited env file name of `.env.local` 👍

PR #366, opened 2023-03-27
new resource package

atk:
The current fetch package is sufficiently modular, but the modules are not reusable for other things than fetch (or with some extra work, other request libraries) requests. The semi-automatic composition is also increasing the bundle size without helping too much to reduce boilerplate.

So what if instead of

```ts
const [data] = createFetch(url, [makeAbortable()]);
```

we would write

```ts
const [signal, abort] = makeAbortable();
const [data] = createResource(url, url => fetch(url, { signal: signal() }).then(r => r.json()));
onCleanup(abort);
```

and still get a smaller overall bundle size? (1.77kb minzipped with fetch vs. 0.7kb minzipped with resource).

The solution lies in 4 simple primitives to augment createResource, agnostic of where what resource is obtained:

- `makeAbortable` - adds a convenient way to use AbortController to stop ongoing during new requests
- `makeAggregated` - aggregates the results
- `makeCache` - caches the results, handles persistence and invalidation
- `makeRetry` - in case of errors, try again after a delay for a configurable number of times


thetarnav (commented):
just looked at the readme so far but this is great. the primitives are small and easy to understand.
and the examples are golden, it's something we should be doing in more packages to make sure that all the usecases are possible and considered.

packages/resource/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+  options: RetryOptions = {},
+): ResourceFetcher<S, T, R> {
+  const delay = options.delay ?? 5000;
+  let retries = options.retries || 3;
+  const retrying: ResourceFetcher<S, T, R> = async (k: S, info: ResourceFetcherInfo<T, R>) => {
+    try {
+      return await fetcher(k, info);
+    } catch (error: unknown) {
+      if (retries-- > 0) {
+        delay && await new Promise(resolve => setTimeout(() => resolve, delay));
```
thetarnav, line 192:
```suggestion
        delay && await new Promise(resolve => setTimeout(resolve, delay));
```
Here, `resolve` will never be called, right?

packages/resource/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+ * ```ts
+ * const fetcher = makeRetrying((url) => fetch(url), { retries: 5 });
+ * ```
+ * Receives the fetcher and an optional options object and returns a wrapped fetcher that retries on error after a delay multiple times.
+ *
+ * The optional options object contains the following optional properties:
+ * - `delay` - number of Milliseconds to wait before retrying; default is 5s
+ * - `retries` - number of times a request should be repeated before giving up throwing the last error; default is 3 times
+ */
+export function makeRetrying<S, T, R = unknown>(
```
thetarnav, line 181: I really don't like the name 🤣
But `makeRetries` or `makeRetry` aren't much better...

packages/resource/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+      : Array.isArray(previous || current)
+      ? [...toArray(previous), ...toArray(current)]
+      : typeof (previous || current) === "object"
+      ? Object.assign({}, previous || {}, current || {})
+      : typeof previous === "string" || typeof current === "string"
+      ? (previous?.toString() || "") + (current || "")
+      : previous
+      ? [previous, current]
+      : [current];
+  }, initialValue);
```
thetarnav, line 235: I feel like this could be also exported as a separate helper.
Like `aggregate` or `getAggregated` just as a vanilla js pure function.

packages/resource/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+ * Depending on the content of the initialValue or the first response, this will aggregate the incoming responses:
+ * - null will not overwrite undefined
+ * - if the previous value is an Array, incoming values will be appended
+ * - if any of the values are Objects, the current one will be shallow-merged into the previous one
+ * - if the previous value is a string, more string data will be appended
+ * - otherwise the incoming data will be put into an array
+ *
+ * Objects and Arrays are re-created on each operation, but the values will be left untouched, so `<For>` should work fine.
+ */
+export function makeAggregated<R, I extends R | R[]>(res: Resource<R>, initialValue?: I) {
```
thetarnav, line 221: 1. this shouldn't be called `makeAggregated` but `createAggregated` if it creates a memo. The point of the `make` name is to inform that the primitive is not reactive, when this clearly is.
2. this seems like something that would be added to separate package on it's own. I don't care about doing this now, but it feels very generic.
3. It doesn't have to require `Resource<R>`, but just `Accessor<R>`, right?

packages/resource/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+ * ```
+ * Wraps the fetcher to use a cache. Returns the wrapped fetcher, an invalidate callback that requires the source to invalidate the request and a signal accessor with the last automatically invalidated request.
+ *
+ * Can be customized with the following optional options:
+ * - `cache` - allows to use a local cache instead of the global one
+ * - `expires` - allows to define a custom timeout; either accepts a number or a function that receives an object with source and data of the request and returns a number in Milliseconds
+ * - `storage` - a storage like localStorage to persist the cache over reloads
+ * - `storageKey` - the key which is used to store the cache in the storage
+ */
+export function makeCache<T, S, R>(
```
thetarnav, line 107: I don't like that the cache makes so many assumptions.
1. What if the params / result are not, and shouldn't be serializable?
2. What If I want to clear all the cache? not one by one by a single key
3. What if I don't care about expiry data, shouldn't this be just another primitive that hooks into `invalidate` or `invalidateAll`?

Probably doesn't matter now as it's focusing on createResource and fetching, but in the future it could be generalized and extracted to a separate package.

packages/resource/README.md
```diff
@@ -0,0 +1,222 @@
...
+import { createEventListener } from "@solid-primitives/event-listener";
+
+const [data, { refetch }] = createResource(() => fetch("url"));
+createEventListener(document, "visibilitychange", () => document.hidden || refetch());
+```
+
+You could also augment this code with the [`scheduled`](https://github.com/solidjs-community/solid-primitives/tree/main/packages/scheduled) package to throttle the refetch calls not to happen again sooner than after 5s:
+
+```ts
+import { createEventSignal } from "@solid-primitives/event-listener";
```
thetarnav, line 176:
```suggestion
import { createEventListener } from "@solid-primitives/event-listener";
```

thetarnav (commented):
some more 🙃

Feel free to merge this once you resolve or just consider the comments. The package is good and I'm just thinking of ways to further improve it.

packages/resource/src/index.ts
atk: 1. Then you should replace them with a serializable key, as stated in the documentation. That being said, I'm able to serialize more than JSON.stringify can handle.
2. That's actually a use case I should cover with invalidate.all(), let me add that later.
3. I originally had that separated in the fetch primitive, but the logic to allow for this separation was larger than the signal.

packages/resource/src/index.ts
atk: It was the least worst name I could find.

packages/resource/src/index.ts
atk: That being said: how would you persist the cache at all then?

packages/resource/README.md
```diff
@@ -11,10 +11,10 @@
...
 A collection of composable primitives to augment [`createResource`](https://www.solidjs.com/docs/latest/api#createresource)
 
-- [`makeAbortable`](#makeabortable) - wraps the fetcher to be abortable and auto-abort on re-fetch or timeout
-- [`makeAggregated`](#makeaggregated) - wraps the resource to aggregate data instead of overwriting it
-- [`makeCache`](#makecache) - wraps the fetcher to cache the responses for a certain amount of time
-- [`makeRetrying`](#makeretrying) - wraps the fetcher to retry requests after a delay
+[`createAggregated`](#createaggregated) - wraps the resource to aggregate data instead of overwriting it
+[`makeAbortable`](#makeabortable) - wraps the fetcher to be abortable and auto-abort on re-fetch or timeout
+[`makeCache`](#makecache) - wraps the fetcher to cache the responses for a certain amount of time
+[`makeRetrying`](#makeretrying) - wraps the fetcher to retry requests after a delay
```
thetarnav: if you don't put `-` in front, or add extra empty line between they get rendered like this:

image

PR #367, opened 2023-03-28
filesystem: improve fail case

atk:
As previously announced, I wanted to improve the fail case; instead of returning a defunct file system adapter, returning null enables the user to detect if they need another case; this also allows to ||-chain file system adapters.


packages/resource/test/index.test.ts
thetarnav, line 1: I think this got included by mistake

atk: Good catch :)

PR #369, opened 2023-03-29
new package: Lifecycle

thetarnav:
This adds a new `lifecycle` package that extends the lifecycle primitives that solid provides with some new ones.

### `createIsMounted`

```tsx
const isMounted = createIsMounted();
const windowWidth = createMemo(() => (isMounted() ? ref.offsetWidth : 0));

let ref: HTMLElement;
<div ref={ref}>{windowWidth()}</div>;
```

### `isHydrated`

```tsx
// This component will only render its children on the client
export const ClientOnly: FlowComponent = props => {
  const children = createMemo(() => isHydrated() && props.children);
  return <>{children()}</>;
};

// Usage
<ClientOnly>
  <ComponentThatBreaksOnServer />
</ClientOnly>;
```

### `onConnect`

```tsx
<div
  ref={el => {
    // often false, but will be true during hydration
    el.isConnected;

    onMount(() => {
      // often true, but will be false if the executed component is not actually getting rendered
      el.isConnected;
    });

    onConnect(el, () => {
      // always true
      el.isConnected;
    });
  }}
/>
```

[README](https://github.com/solidjs-community/solid-primitives/tree/lifecycle/packages/lifecycle#readme)

### DIFF
--- a/packages/lifecycle/README.md
+++ b/packages/lifecycle/README.md
@@ -13,7 +13,7 @@ Package providing extra layer of lifecycle primitives for Solid.
 
 - [`createIsMounted`](#createIsMounted) - Returns a boolean signal indicating whether the component is mounted or not.
 - [`isHydrated`](#isHydrated) - A signal with the same behavior as [`isHydrating`](#isHydrating) but this one focused only on client-side updates.
-- [`onConnect`](#onConnect) - Calls the given callback when the target element is connected to the DOM.
+- [`onElementConnect`](#onElementConnect) - Calls the given callback when the target element is connected to the DOM.
 
 ## Installation
 
@@ -90,7 +90,7 @@ export const ClientOnly: FlowComponent = props => {
 </ClientOnly>;
 ```
 
-## `onConnect`
+## `onElementConnect`
 
 `onMount` is a common lifecycle hook that is used to perform side-effects when the component is mounted.
 However, it is not certain that the elements are actually connected to the DOM when the mount callback is called.
@@ -99,7 +99,7 @@ However, it is not certain that the elements are actually connected to the DOM w
 >
 > And if this is something intentional, you probably already have a way to hook into the actual DOM rendering.
 >
-> If you are not sure, you can use `onConnect` instead of `onMount` to make sure that you are caling your callback when the elements are connected to the DOM.
+> If you are not sure, you can use `onElementConnect` instead of `onMount` to make sure that you are caling your callback when the elements are connected to the DOM.
 
 ```tsx
 <div
@@ -112,7 +112,7 @@ However, it is not certain that the elements are actually connected to the DOM w
       el.isConnected;
     });
 
-    onConnect(el, () => {
+    onElementConnect(el, () => {
       // always true
       el.isConnected;
     });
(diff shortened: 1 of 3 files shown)


packages/lifecycle/README.md
```diff
@@ -0,0 +1,125 @@
...
+[![turborepo](https://img.shields.io/badge/built%20with-turborepo-cc00ff.svg?style=for-the-badge&logo=turborepo)](https://turborepo.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/lifecycle?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/lifecycle)
+[![version](https://img.shields.io/npm/v/@solid-primitives/lifecycle?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/lifecycle)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
+
+Package providing extra layer of lifecycle primitives for Solid.
+
+- [`createIsMounted`](#createIsMounted) - Returns a boolean signal indicating whether the component is mounted or not.
+- [`isHydrated`](#isHydrated) - A signal with the same behavior as [`isHydrating`](#isHydrating) but this one focused only on client-side updates.
+- [`onConnect`](#onConnect) - Calls the given callback when the target element is connected to the DOM.
```
atk, line 16: This name might lead to some confusion with the connectivity package. `onReady` maybe?

atk (approved):

packages/lifecycle/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+ * ```tsx
+ * let ref: HTMLElement
+ * const isMounted = createIsMounted();
+ * const windowWidth = createMemo(() => isMounted() ? ref.offsetWidth : 0)
+ * <div ref={ref}>{windowWidth()}</div>
+ * ```
+ * @see https://github.com/solidjs-community/solid-primitives/tree/main/packages/lifecycle#createIsMounted
+ */
+export function createIsMounted(): Accessor<boolean> {
+  if (isServer) return () => false;
```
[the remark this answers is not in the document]
thetarnav: it'll get tree shaken as `isServer` is a const

packages/lifecycle/README.md
thetarnav: I'm not sure
connect is related to the element being connected to the DOM, if you know about the `el.isConnected` field, you'll understand this one.
if they were in the same package - sure, but they are separated for a reason - component/reactivity lifecycle has nothing to do with internet connection

thetarnav: so it's probably the connectivity package that is wrongly named. for "connect" to have meaning it needs to have context

thetarnav: I guess I could name it `onElementConnect` to give that missing context here 🤔

atk: Maybe `onConnectNode`?

PR #370, opened 2023-03-29
New package: Static Store

thetarnav:
I've decided to pull the `createStaticStore` primitive out of `utils` and make it into a separate package: `static-store`.
The reasons are:
- that this primitive has been used a lot internally - there are 7 other packages that depend on it, and I've been using it outside of solid-primitives too.
- I've added a `createDerivedStaticStore` primitive that will help with more derived cases where previously `createComputed` had to be used - in an attempt to reduce the use of computers as much as possible. This increases the total amount of static-store primitives to 3 - including `createHydratableStaticStore`, and that's a size of a small family.

Removing the primitives from utils means a breaking change - hence the major version.

[README](https://github.com/solidjs-community/solid-primitives/tree/static-store/packages/static-store#readme)

Wondering if it shouldn't just be called `static` - `crateStatic`, `createDerivedStatic`, etc.


packages/static-store/README.md
```diff
@@ -0,0 +1,112 @@
...
+npm install @solid-primitives/static-store
+# or
+yarn add @solid-primitives/static-store
+# or
+pnpm add @solid-primitives/static-store
+```
+
+## `createStaticStore`
+
+A shallowly wrapped reactive store object. It behaves similarly to the createStore, but with limited features to keep it simple. Designed to be used for reactive objects with static keys, but dynamic values, like reactive Event State, location, etc.
```
atk, line 29: I find this description slightly confusing. Maybe rephrase it like:

"Creates read-only object that is solely reactive on its first level and for the enumerable properties specified in the initial value, and a setter. It behaves similarly to createStore, but with limited features to keep it simple and performant. Designed to be used for reactive objects with static keys and dynamic values, like reactive Event State, location, etc."

packages/static-store/dev/index.tsx
atk, line 1: This is not even demoing the primitives.

atk (changes requested):

packages/static-store/README.md
thetarnav: hmm I love that, thanks 👍

packages/static-store/dev/index.tsx
thetarnav: Yeah I know I didn't have the energy to come up with a demo, do I just disabled the `page` script 🙃
Maybe I'll add one later.

atk (approved):

PR #381, opened 2023-04-04
test script loading

atk:
It's known to be impossible to test actual script loading with vitest and jsdom, so, naturally, I tried to work around that issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@mdx-js/mdx": "^2.3.0",
     "@solidjs/testing-library": "^0.7.0",
     "@types/fs-extra": "^11.0.1",
+    "@types/jsdom": "^21.1.1",
     "@types/node": "^18.15.11",
     "@typescript-eslint/eslint-plugin": "^5.57.0",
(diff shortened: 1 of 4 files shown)


packages/script-loader/package.json
```diff
@@ -42,6 +42,10 @@
     "test": "vitest -c ../../configs/vitest.config.ts",
     "test:ssr": "pnpm run test --mode ssr"
   },
+  "devDependencies": {
+    "@types/jsdom": "^21.1.1",
```
thetarnav, line 46: Those should probably be added to the root package.json do they are available in each package
`jsdom` is already added this way

atk: I know, but I get a TS error, since it cannot find the dependencies in the local package.json.

thetarnav: I mean it has to be just the types, they are not added to root package.json. Might as well at it there. It should be available just as `vitest` it

atk: Let me try tomorrow.

PR #384, opened 2023-04-07
pass the keyboard event to the callback in createShortcut

vanillacode314:
Use Case: I want to only preventDefault if document.activeElement is not an HTMLInputElement


thetarnav:
Sorry for the delay. I think it's reasonable to get the event that triggered the callback. I wasn't exposing it originally because it's available if you use `useKeyDownList`, but there is no harm in making it a bit easier to get.

PR #385, opened 2023-04-07
use passive event listener for createScrollPosition

vanillacode314:
reasoning: we don't preventDefault and the user cannot preventDefault


thetarnav:
Makes sense, thanks!

PR #394, opened 2023-04-12
site: Performance optimization, fix #393

### LINKED ISSUE #393: Requests flood on client after page load on primitives site
### Describe the bug

This bug is regarding the primitives site at https://primitives.solidjs.community/
I noticed after loading the site that there was a flood of fetch requests on initial load,
i dug a bit in the code and found that the Component `<VersionBadge>` is using the helper `createPackageVersion` to get the package version, which in itself uses `fetchPackageVersion` on the client if the version is not cached, code:
```tsx
function createPackageVersion(source: Accessor<string>) {
  const memo = createMemo(() => {
    const name = source();

    return untrack(() => {
      const cached = versionCache.get(name);

      const [version, setVersion] = createSignal(
        // use cached version if available (if not hydrating to avoid mismatches)
        cached ||
          (!isServer && !sharedConfig.context && getCachedPackageListItemData(name)?.version) ||
          "0.0.0",
      );

      // fetch the version on the client so it won't be out of sync with SSG build
      if (!isServer && !cached) {
        fetchPackageVersion(name).then(v => {
          setVersion(v);
          versionCache.set(name, v);
        });
      }
      return version;
    });
  });

  return () => memo()();
}

```
I wondered if there was an optimization we can do to maybe batch fetch the versions instead of calling fetchPackageVersion for each package (will get even more costly when the number grows),
or just use the value returned from the server,
i'd like to state that this is my first "bug report" and not sure about the reasoning of this behaviour, would be great to correct me

### Minimal Reproduction Link

https://primitives.solidjs.community/

eudrinoagain:
This PR fixes the issue #393 regarding initial load times & unneccesary fetch requests on the client,

Changes:
1. i've changed the VersionBadge component to accept a `version` prop, which we can use instead of calling `createPackageVersion` which introduced the request flood.
2. i've changed the `route.tsx` (homepage) to pass the `pkg.version` as a prop to `VersionBadge`, which is already available without extra requests.

i'm not used to the codebase so feel free to suggest changes.

### DIFF
--- a/site/src/components/Primitives/VersionBadge.tsx
+++ b/site/src/components/Primitives/VersionBadge.tsx
@@ -50,9 +50,7 @@ function createPackageVersion(source: Accessor<string>) {
   return () => memo()();
 }
 
-export const VersionBadge: ParentComponent<{ name: string }> = props => {
-  const version = createPackageVersion(() => props.name);
-
+export const VersionBadge: ParentComponent<{ name: string; version: string }> = props => {
   return (
     <a
       class="flex h-[28px] min-w-[90px] items-baseline justify-center rounded-md border-2 border-[#cae0ff] bg-[#cae0ff40] font-sans transition-colors hover:border-[#80a7de] hover:bg-[#cae0ff66] dark:border-[#5577a7] dark:bg-[#6eaaff14] dark:hover:border-[#8ba8d3] dark:hover:bg-[#6eaaff33]"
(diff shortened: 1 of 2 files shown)


atk (approved):
Looks good to me.

thetarnav (approved):
Looks good to me as well
@eudrinoagain did you have a look at the `VersionBadgePill` component as well? It's used in the `/package/:name` route and potentially could also receive the version through props. wdyt?
https://github.com/solidjs-community/solid-primitives/blob/db0b84408cf0d97289470f02fc81fa34ddd6481b/site/src/components/Primitives/VersionBadge.tsx#L68

thetarnav:
here's a deploy preview: https://6436fa1273cfe41435a7b174--solid-primitives.netlify.app/
the request flood is indeed gone :)

PR #397, opened 2023-04-13
site: Removes unneccessary fetch from (package).tsx, related #393

eudrinoagain:
This PR fixes a related issue to #393,
but for the `<VersionBadgePill>` component,
as it also can recieve its `version` resource from the package route,


thetarnav:
Thanks again
Since `createPackageVersion` is not being used anymore, why not remove it?
Here is a deploy preview: https://643826fb33e783166c6b0604--solid-primitives.netlify.app
Seems to be working fine

eudrinoagain:
@thetarnav
I've removed both `createPackageVersion` and `fetchPackageVersion` as they are not used anymore

thetarnav:
thanks!

PR #398, opened 2023-04-13
Fix MutationObserver doesn't disconnect when there're no pending notifications

QiroNT:
Since this is a one-line change, I wasn't bothered to post a issue for it.

> **MutationObserver: takeRecords() method**
> The [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) method takeRecords() returns a list of all matching DOM changes that have been detected but not yet processed by the observer's callback function, leaving the mutation queue empty.
> -- <https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/takeRecords>

Therefore `takeRecords()` returns an empty array when there're no pending notifications, canceling the `disconnect()`call.

### DIFF
--- a/.changeset/quick-plums-punch.md
+++ b/.changeset/quick-plums-punch.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/mutation-observer": patch
+---
+
+Fix MutationObserver doesn't disconnect when there're no pending notifications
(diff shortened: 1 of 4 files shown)


thetarnav:
From looking at the code, I can't really tell what was the point of that check anyway.
Thanks for the fixing that and adding more tests 🙏

PR #403, opened 2023-04-15
resource: add createDeepSignal

atk:


packages/resource/README.md
```diff
@@ -65,6 +66,24 @@ const pages = createAggregated(currentPage, []);
 
 Objects and Arrays are re-created on each operation, but the values will be left untouched, so `<For>` should work fine.
 
+### createDeepSignal
+
+This is an optimized implementation of what is required by the [resource storage option](https://www.solidjs.com/docs/latest/api#createresource:~:text=Resources%20can%20be%20set%20with%20custom%20defined%20storage):
```
thetarnav, line 71: This doesn't really explain much of what `createDeepSignal` actually does. Besides the fact that it's meant to be used with the `storage` option.
The `storage` option lets you use something else besides a simple signal for storing the response, it doesn't assume anything about what it should implement besides the signal interface.
So it's not really an "optimized implementation", because there is nothing to compare against.
I think it should instead say that it's creating a store and will reconcile new responses so that the changes are fine-grain, and references will be preserved so using `<For>` is possible.

packages/resource/README.md
```diff
@@ -65,6 +66,24 @@ const pages = createAggregated(currentPage, []);
...
+
+```ts
+// this adds fine-grained reactivity to the contents of data():
+const [data, { refetch }] = createResource(fetcher, { storage: createDeepSignal });
+```
+
+> ⚠ if your resource is a deep signal, you can no longer rely on reactive changes to the base signal. If you want to combine this with [`createAggregated`](#createaggregated), you will need to wrap the resource to either call [`deepTrack`](https://solid-primitives.netlify.app/package/deep#deeptrack) or read one of the reactive parts that you know for certain will change every time:
```
thetarnav, line 78:
```suggestion
> **Warning** if your resource is a deep signal, you can no longer rely on reactive changes to the base signal. If you want to combine this with [`createAggregated`](#createaggregated), you will need to wrap the resource to either call [`deepTrack`](https://solid-primitives.netlify.app/package/deep#deeptrack) or read one of the reactive parts that you know for certain will change every time:
```
Github mas a special styling for these: https://github.com/orgs/community/discussions/16925
Our website will too

packages/resource/src/index.ts
```diff
@@ -232,11 +235,23 @@ export function createAggregated<R, I extends R | R[]>(res: Accessor<R>, initial
...
       ? (previous?.toString() || "") + (current || "")
       : previous
       ? [previous, current]
       : [current];
   }, initialValue);
 }
+
+export function createDeepSignal<T>(): Signal<T | undefined>;
+export function createDeepSignal<T>(value: T): Signal<T>;
+export function createDeepSignal<T>(v?: T): Signal<T> {
```
thetarnav, line 249: Some JSDoc? 🙃

thetarnav (approved):

packages/resource/README.md
atk: Actually, it is size-optimized compared to the implementation in the solid docs (185bytes vs. 212bytes). But that point is probably moot.

PR #413, opened 2023-04-21
Add marker package for search highlighting

thetarnav:
It replaces the `Mark.js` package used for search highlighting with a custom solid primitive for highlighting.
~~I'll make it into an actual package because it seems to be working well, this is just a draft.~~ (done)
I removed the highlight checkboxes too as they don't have much sense to me, but it could be reverted if actually needed.

Adds new `marker` package to UI Patterns category with `createMarker` primitive for doing this kind of highlighting.

```tsx
import { createMarker } from "@solid-primitives/marker";

const highlight = createMarker(matchedText => <mark>{matchedText()}</mark>);

<p>
  {highlight("Hello world!", /\w+/g)} {/* <mark>Hello</mark> <mark>world</mark>! */}
</p>;
```

[README](https://github.com/solidjs-community/solid-primitives/tree/site-highlights/packages/marker#readme)


packages/marker/README.md
```diff
@@ -0,0 +1,81 @@
...
+});
+```
+
+## Caching
+
+The marker callback is cached between calls.
+
+This way returned elements are reused as much as possible.
+
+But every cache needs a limit. By default the cache size is 100. You can change it by passing the `cacheSize` option to `createMarker`.
```
atk, line 67: Missing comma: "By default the cache size" -> "By default, the cache size"

packages/marker/dev/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+] as const;
+
+const App: Component = () => {
+  const [search, setSearch] = createSignal("");
+
+  const highlight = createMarker(text => <mark>{text()}</mark>);
+  const regex = createMemo(() => new RegExp(search(), "gi"));
```
atk, line 15: How does this handle inputs that result in illegal regex, like "+" or "]"?

Since the example is also a form of documentation, we should at least add a comment and/or an error boundary.

site/src/components/Search/Search.tsx
atk, line 1: :heart: that we are dogfooding.

atk (approved):
There are only minor issues, so it's your choice to fix them before you merge.

packages/marker/dev/index.tsx
thetarnav: could you explain a bit more? I would love to improve this example (in the our site too), but I don't exactly know how.
Should I just wrap it in a `try catch`? I wanted to search for some quality examples online, but every article is just showing that you can use `new Regex` to create a dynamic regex, not how it's supposed to be done

thetarnav: alright, I've added `makeSearchRegex` to improve that part a little bit

atk: Good idea.

PR #414, opened 2023-04-22
Proposal: presence-signal

EthanStandel:
I need to write tests for this package, so I'm keeping it as a draft for now. But I'd love to hear any initial feedback or validate if there is any interest in this addition.

This is a proposal to merge my [`solid-presence-signal`](https://www.npmjs.com/package/solid-presence-signal) package from it's own repo into the @solid-primitives ecosystem. This package is a conversion of a lightweight React animation package called [use-presence](https://www.npmjs.com/package/use-presence). I'd also be up for a name change like `create-presence` or some other alternative if there was a preference.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,23864 @@
+{
+  "name": "solid-primitives",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "solid-primitives",
+      "license": "MIT",
+      "devDependencies": {
+        "@changesets/cli": "^2.26.1",
+        "@mdx-js/mdx": "^2.3.0",
+        "@solidjs/testing-library": "^0.7.0",
+        "@types/fs-extra": "^11.0.1",
+        "@types/jsdom": "^21.1.1",
+        "@types/node": "^18.15.11",
+        "@typescript-eslint/eslint-plugin": "^5.58.0",
+        "@typescript-eslint/parser": "^5.58.0",
+        "concurrently": "^8.0.1",
+        "cross-env": "^7.0.3",
+        "esbuild": "^0.17.16",
+        "eslint": "^8.38.0",
+        "eslint-plugin-eslint-comments": "^3.2.0",
+        "eslint-plugin-no-only-tests": "^3.1.0",
+        "eslint-plugin-solid": "^0.12.1",
+        "fs-extra": "^11.1.1",
+        "gzip-size": "^7.0.0",
+        "jiti": "^1.18.2",
+        "jsdom": "^21.1.1",
+        "json-to-markdown-table": "^1.0.0",
+        "prettier": "^2.8.7",
+        "prettier-plugin-tailwindcss": "^0.2.7",
+        "rehype-autolink-headings": "^6.1.1",
+        "rehype-highlight": "^6.0.0",
+        "rehype-slug": "^5.1.0",
+        "remark-gfm": "^3.0.1",
+        "solid-devtools": "^0.26.0",
+        "solid-js": "1.7.3",
+        "tsup": "^6.7.0",
+        "tsup-preset-solid": "^0.1.8",
+        "turbo": "^1.9.1",
+        "type-fest": "^3.8.0",
+        "typescript": "~5.0.4",
+        "unocss": "^0.50.8",
+        "vite": "^4.2.1",
+        "vite-plugin-solid": "^2.7.0",
+        "vitest": "^0.29.8"
+      },
+      "engines": {
+        "node": ">=18.0.0"
+      }
+    },
+    "node_modules/@ampproject/remapping": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
+      "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
+      "dev": true,
+      "dependencies": {
+        "@jridgewell/gen-mapping": "^0.3.0",
+        "@jridgewell/trace-mapping": "^0.3.9"
+      },
+      "engines": {
+        "node": ">=6.0.0"
+      }
+    },
+    "node_modules/@antfu/install-pkg": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/@antfu/install-pkg/-/install-pkg-0.1.1.tgz",
+      "integrity": "sha512-LyB/8+bSfa0DFGC06zpCEfs89/XoWZwws5ygEa5D+Xsm3OfI+aXQ86VgVG7Acyef+rSZ5HE7J8rrxzrQeM3PjQ==",
+      "dev": true,
+      "dependencies": {
+        "execa": "^5.1.1",
+        "find-up": "^5.0.0"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/antfu"
+      }
+    },
+    "node_modules/@antfu/install-pkg/node_modules/find-up": {
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/find-up/-/find-up-5.0.0.tgz",
+      "integrity": "sha512-78/PXT1wlLLDgTzDs7sjq9hzz0vXD+zn+7wypEe4fXQxCmdmqfGsEPQxmiCSQI3ajFV91bVSsvNtrJRiW6nGng==",
+      "dev": true,
+      "dependencies": {
+        "locate-path": "^6.0.0",
+        "path-exists": "^4.0.0"
+      },
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@antfu/install-pkg/node_modules/locate-path": {
+      "version": "6.0.0",
+      "resolved": "https://registry.npmjs.org/locate-path/-/locate-path-6.0.0.tgz",
+      "integrity": "sha512-iPZK6eYjbxRu3uB4/WZ3EsEIMJFMqAoopl3R+zuq0UjcAm/MO6KCweDgPfP3elTztoKP3KtnVHxTn2NHBSDVUw==",
+      "dev": true,
+      "dependencies": {
+        "p-locate": "^5.0.0"
+      },
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@antfu/install-pkg/node_modules/p-limit": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-3.1.0.tgz",
+      "integrity": "sha512-TYOanM3wGwNGsZN2cVTYPArw454xnXj5qmWF1bEoAc4+cU/ol7GVh7odevjp1FNHduHc3KZMcFduxU5Xc6uJRQ==",
+      "dev": true,
+      "dependencies": {
+        "yocto-queue": "^0.1.0"
+      },
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@antfu/install-pkg/node_modules/p-locate": {
+      "version": "5.0.0",
+      "resolved": "https://registry.npmjs.org/p-locate/-/p-locate-5.0.0.tgz",
+      "integrity": "sha512-LaNjtRWUBY++zB5nE/NwcaoMylSPk+S+ZHNB1TzdbMJMny6dynpAGt7X/tl/QYq3TIeE6nxHppbo2LGymrG5Pw==",
+      "dev": true,
+      "dependencies": {
+        "p-limit": "^3.0.2"
+      },
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@antfu/utils": {
+      "version": "0.7.2",
+      "resolved": "https://registry.npmjs.org/@antfu/utils/-/utils-0.7.2.tgz",
+      "integrity": "sha512-vy9fM3pIxZmX07dL+VX1aZe7ynZ+YyB0jY+jE6r3hOK6GNY2t6W8rzpFC4tgpbXUYABkFQwgJq2XYXlxbXAI0g==",
+      "dev": true,
+      "funding": {
+        "url": "https://github.com/sponsors/antfu"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.21.4.tgz",
+      "integrity": "sha512-LYvhNKfwWSPpocw8GI7gpK2nq3HSDuEPC/uSYaALSJu9xjsalaaYFOq0Pwt5KmVqwEbZlDu81aLXwBOmD/Fv9g==",
+      "dev": true,
+      "dependencies": {
+        "@babel/highlight": "^7.18.6"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/compat-data": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.21.4.tgz",
+      "integrity": "sha512-/DYyDpeCfaVinT40FPGdkkb+lYSKvsVuMjDAG7jPOWWiM1ibOaB9CXJAlc4d1QpP/U2q2P9jbrSlClKSErd55g==",
+      "dev": true,
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/core": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.21.4.tgz",
+      "integrity": "sha512-qt/YV149Jman/6AfmlxJ04LMIu8bMoyl3RB91yTFrxQmgbrSvQMy7cI8Q62FHx1t8wJ8B5fu0UDoLwHAhUo1QA==",
+      "dev": true,
+      "dependencies": {
+        "@ampproject/remapping": "^2.2.0",
+        "@babel/code-frame": "^7.21.4",
+        "@babel/generator": "^7.21.4",
+        "@babel/helper-compilation-targets": "^7.21.4",
+        "@babel/helper-module-transforms": "^7.21.2",
+        "@babel/helpers": "^7.21.0",
+        "@babel/parser": "^7.21.4",
+        "@babel/template": "^7.20.7",
+        "@babel/traverse": "^7.21.4",
+        "@babel/types": "^7.21.4",
+        "convert-source-map": "^1.7.0",
+        "debug": "^4.1.0",
+        "gensync": "^1.0.0-beta.2",
+        "json5": "^2.2.2",
+        "semver": "^6.3.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/babel"
+      }
+    },
+    "node_modules/@babel/core/node_modules/semver": {
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
+      "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw==",
+      "dev": true,
+      "bin": {
+        "semver": "bin/semver.js"
+      }
+    },
+    "node_modules/@babel/generator": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.21.4.tgz",
+      "integrity": "sha512-NieM3pVIYW2SwGzKoqfPrQsf4xGs9M9AIG3ThppsSRmO+m7eQhmI6amajKMUeIO37wFfsvnvcxQFx6x6iqxDnA==",
+      "dev": true,
+      "dependencies": {
+        "@babel/types": "^7.21.4",
+        "@jridgewell/gen-mapping": "^0.3.2",
+        "@jridgewell/trace-mapping": "^0.3.17",
+        "jsesc": "^2.5.1"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-annotate-as-pure": {
+      "version": "7.18.6",
+      "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.18.6.tgz",
+      "integrity": "sha512-duORpUiYrEpzKIop6iNbjnwKLAKnJ47csTyRACyEmWj0QdUrm5aqNJGHSSEQSUAvNW0ojX0dOmK9dZduvkfeXA==",
+      "dev": true,
+      "dependencies": {
+        "@babel/types": "^7.18.6"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-builder-binary-assignment-operator-visitor": {
+      "version": "7.18.9",
+      "resolved": "https://registry.npmjs.org/@babel/helper-builder-binary-assignment-operator-visitor/-/helper-builder-binary-assignment-operator-visitor-7.18.9.tgz",
+      "integrity": "sha512-yFQ0YCHoIqarl8BCRwBL8ulYUaZpz3bNsA7oFepAzee+8/+ImtADXNOmO5vJvsPff3qi+hvpkY/NYBTrBQgdNw==",
+      "dev": true,
+      "optional": true,
+      "dependencies": {
+        "@babel/helper-explode-assignable-expression": "^7.18.6",
+        "@babel/types": "^7.18.9"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-compilation-targets": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/helper-compilation-targets/-/helper-compilation-targets-7.21.4.tgz",
+      "integrity": "sha512-Fa0tTuOXZ1iL8IeDFUWCzjZcn+sJGd9RZdH9esYVjEejGmzf+FFYQpMi/kZUk2kPy/q1H3/GPw7np8qar/stfg==",
+      "dev": true,
+      "dependencies": {
+        "@babel/compat-data": "^7.21.4",
+        "@babel/helper-validator-option": "^7.21.0",
+        "browserslist": "^4.21.3",
+        "lru-cache": "^5.1.1",
+        "semver": "^6.3.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      },
+      "peerDependencies": {
+        "@babel/core": "^7.0.0"
+      }
+    },
+    "node_modules/@babel/helper-compilation-targets/node_modules/semver": {
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
+      "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw==",
+      "dev": true,
+      "bin": {
+        "semver": "bin/semver.js"
+      }
+    },
+    "node_modules/@babel/helper-create-class-features-plugin": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/helper-create-class-features-plugin/-/helper-create-class-features-plugin-7.21.4.tgz",
+      "integrity": "sha512-46QrX2CQlaFRF4TkwfTt6nJD7IHq8539cCL7SDpqWSDeJKY1xylKKY5F/33mJhLZ3mFvKv2gGrVS6NkyF6qs+Q==",
+      "dev": true,
+      "dependencies": {
+        "@babel/helper-annotate-as-pure": "^7.18.6",
+        "@babel/helper-environment-visitor": "^7.18.9",
+        "@babel/helper-function-name": "^7.21.0",
+        "@babel/helper-member-expression-to-functions": "^7.21.0",
+        "@babel/helper-optimise-call-expression": "^7.18.6",
+        "@babel/helper-replace-supers": "^7.20.7",
+        "@babel/helper-skip-transparent-expression-wrappers": "^7.20.0",
+        "@babel/helper-split-export-declaration": "^7.18.6"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      },
+      "peerDependencies": {
+        "@babel/core": "^7.0.0"
+      }
+    },
+    "node_modules/@babel/helper-create-regexp-features-plugin": {
+      "version": "7.21.4",
+      "resolved": "https://registry.npmjs.org/@babel/helper-create-regexp-features-plugin/-/helper-create-regexp-features-plugin-7.21.4.tgz",
+      "integrity": "sha512-M00OuhU+0GyZ5iBBN9czjugzWrEq2vDpf/zCYHxxf93ul/Q5rv+a5h+/+0WnI1AebHNVtl5bFV0qsJoH23DbfA==",
+      "dev": true,
+      "optional": true,
+      "dependencies": {
+        "@babel/helper-annotate-as-pure": "^7.18.6",
+        "regexpu-core": "^5.3.1"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      },
+      "peerDependencies": {
+        "@babel/core": "^7.0.0"
+      }
+    },
+    "node_modules/@babel/helper-define-polyfill-provider": {
+      "version": "0.3.3",
+      "resolved": "https://registry.npmjs.org/@babel/helper-define-polyfill-provider/-/helper-define-polyfill-provider-0.3.3.tgz",
+      "integrity": "sha512-z5aQKU4IzbqCC1XH0nAqfsFLMVSo22SBKUc0BxGrLkolTdPTructy0ToNnlO2zA4j9Q/7pjMZf0DSY+DSTYzww==",
+      "dev": true,
+      "optional": true,
+      "dependencies": {
+        "@babel/helper-compilation-targets": "^7.17.7",
+        "@babel/helper-plugin-utils": "^7.16.7",
+        "debug": "^4.1.1",
+        "lodash.debounce": "^4.0.8",
+        "resolve": "^1.14.2",
+        "semver": "^6.1.2"
+      },
+      "peerDependencies": {
+        "@babel/core": "^7.4.0-0"
+      }
+    },
+    "node_modules/@babel/helper-define-polyfill-provider/node_modules/semver": {
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
+      "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw==",
+      "dev": true,
+      "optional": true,
+      "bin": {
+        "semver": "bin/semver.js"
+      }
+    },
+    "node_modules/@babel/helper-environment-visitor": {
+      "version": "7.18.9",
+      "resolved": "https://registry.npmjs.org/@babel/helper-environment-visitor/-/helper-environment-visitor-7.18.9.tgz",
+      "integrity": "sha512-3r/aACDJ3fhQ/EVgFy0hpj8oHyHpQc+LPtJoY9SzTThAsStm4Ptegq92vqKoE3vD706ZVFWITnMnxucw+S9Ipg==",
+      "dev": true,
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-explode-assignable-expression": {
+      "version": "7.18.6",
+      "resolved": "https://registry.npmjs.org/@babel/helper-explode-assignable-expression/-/helper-explode-assignable-expression-7.18.6.tgz",
+      "integrity": "sha512-eyAYAsQmB80jNfg4baAtLeWAQHfHFiR483rzFK+BhETlGZaQC9bsfrugfXDCbRHLQbIA7U5NxhhOxN7p/dWIcg==",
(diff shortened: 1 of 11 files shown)


package-lock.json
thetarnav, line 1: we're using `pnpm` in this project, so I'm not sure if workspaces will work properly without it. this file shouldn't be committed anyway

packages/presence-signal/README.md
```diff
@@ -0,0 +1,173 @@
...
+## This solution
+
+This utility provides a lightweight solution where the animating element is only mounted the minimum of time, while making sure the animation is fully visible to the user. The rendering is left to the user to support all kinds of styling solutions.
+
+## Example
+
+```jsx
+const FirstExample = () => {
+  const [showStuff, setShowStuff] = createSignal(true);
+  const { isVisible, isMounted } = createPresenceSignal(showStuff, {
```
thetarnav, line 23: I'm not sure if the name should include the `Signal` suffix. It's not a direct abstraction over `createSignal`, it uses more primitives inside than just a single signal, so it's creating more than that. Also, signals can be created anywhere and they don't hook into the current reactive owner in any way, they don't have to be disposed of using the ownership graph, but just garbage collected. `createPresenceSignal` is creating computations inside, so the place where it's used matters.
It could be just called `createPresence` and `createSwitchPresence`.

packages/presence-signal/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@solid-primitives/presence-signal",
```
thetarnav, line 2: similarly to the comment above, `presence` should be enough

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+type SharedTransitionConfig = {
+  /** Duration in milliseconds used both for enter and exit transitions. */
+  transitionDuration: number;
+};
+
+type SeparateTransitionConfig = {
+  /** Duration in milliseconds used for enter transitions (overrides `transitionDuration` if provided). */
+  enterTransitionDuration: number;
+  /** Duration in milliseconds used for exit transitions (overrides `transitionDuration` if provided). */
+  exitTransitionDuration: number;
```
thetarnav, line 12: Shouldn't these be optional?
Also why not just call it `duration`, `exitDuration`, and `enterDuration`? transition is the point of this primitive so it will implicitly involve transition
```suggestion
  enterTransitionDuration?: number;
  /** Duration in milliseconds used for exit transitions (overrides `transitionDuration` if provided). */
  exitTransitionDuration?: number;
```

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+  /** Opt-in to animating the entering of an element if `isVisible` is `true` during the initial mount. */
+  initialEnter?: boolean;
+};
+
+/**
+ * Animates the appearance of its children.
+ */
+export default function createPresenceSignal(
+  /** Indicates whether the component that the resulting values will be used upon should be visible to the user. */
+  isVisible: Accessor<boolean>,
```
thetarnav, line 29: why not just call it `source` or `sourceState`? The source signal having the same name as the returned `isVisible` makes it a bit confusing, you might think that this state is getting returned directly when it's not.

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+  const initialEnter = () => opts().initialEnter ?? false;
+  const [animateIsVisible, setAnimateIsVisible] = createSignal(
+    initialEnter() ? false : isVisible(),
+  );
+  const [isMounted, setIsMounted] = createSignal(isVisible());
+  const [hasEntered, setHasEntered] = createSignal(initialEnter() ? false : isVisible());
+
+  const isExiting = () => isMounted() && !isVisible();
+  const isEntering = () => isVisible() && !hasEntered();
+  const isAnimating = () => isEntering() || isExiting();
```
thetarnav, line 55: So we have individual signals for `isMounted` and `isVisible` - they can be listened to separately and will only trigger change when the value actually changes.
While `isExiting`, `isEntering`, and `isAnimating` is derived from them, it's not done with a memo, so they can trigger updates even if the value didn't actually change.
I don't like the inconsistency here. As a user, if I get an accessor from a custom primitive, my impression would be that the value is memorized. How I'm I supposed to tell which accessors to memorize and which I don't have to?

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+      onCleanup(() => clearTimeout(timeoutId));
+    }
+  });
+
+  createEffect(() => {
+    if (isVisible() && isMounted() && !animateIsVisible()) {
+      // Force a reflow so the initial styles are flushed to the DOM
+      if (typeof document !== undefined) {
+        // We need a side effect so Terser doesn't remove this statement
+        (window as any)._usePresenceReflow = document.body.offsetHeight;
```
thetarnav, line 80: I don't think this is necessary. rollup and esbuild won't remove expressions like this.
Here you can see the same pattern being used by solid-transition-group https://github.com/solidjs-community/solid-transition-group/blob/main/src/TransitionGroup.ts#L66 and it never has caused any issues.

```suggestion
        document.body.offsetHeight;
```

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+      }, exitTransitionDuration());
+
+      onCleanup(() => clearTimeout(timeoutId));
+    }
+  });
+
+  createEffect(() => {
+    if (isVisible() && isMounted() && !animateIsVisible()) {
+      // Force a reflow so the initial styles are flushed to the DOM
+      if (typeof document !== undefined) {
```
thetarnav, line 78: Why do we check for document? the code inside `createEffect` won't ever run on the server

packages/presence-signal/src/createPresenceSwitchSignal.ts
```diff
@@ -0,0 +1,32 @@
...
+      } else if (item() !== undefined) {
+        setMountedItem(() => item());
+        setShouldBeMounted(true);
+      }
+    } else if (item() === undefined) {
+      setShouldBeMounted(false);
+    } else if (item() !== undefined) {
+      setShouldBeMounted(true);
+    }
+  });
```
thetarnav, line 25: There is awfully lot of state synchronization using `createEffect`. Couldn't this be expressed with `createMemo`?

packages/presence-signal/README.md
```diff
@@ -0,0 +1,173 @@
...
+}
+```
+
+### Example
+
+```jsx
+const SecondExample = () => {
+  const items = ["foo", "bar", "baz", "qux"];
+  const [item, setItem] =
+    (createSignal < typeof items[number]) | (undefined > items[0]);
```
thetarnav, line 113: Did something happen to the formatting here? it doesn't seem real

packages/presence-signal/src/createPresenceSignal.ts
```diff
@@ -0,0 +1,108 @@
...
+      : optsValue.transitionDuration;
+  };
+  const enterTransitionDuration = () => {
+    const optsValue = opts();
+    return "enterTransitionDuration" in optsValue
+      ? optsValue.enterTransitionDuration
+      : optsValue.transitionDuration;
+  };
+
+  const initialEnter = () => opts().initialEnter ?? false;
```
thetarnav, line 46: Why is this a function? the state of `initialEnter` never change, right?

thetarnav (commented):
Thanks for bringing this package to solid-primitives 🥳
I love the concept behind these primitives and I'm interested to see how the patterns around animations and transitions will evolve over time in solid.

package-lock.json
EthanStandel: This was a mistake and has been removed and I've reinstalled using `pnpm` locally 👍

packages/presence-signal/README.md
EthanStandel: I like this and think it fits the conventions already applied in the other packages in this repo. As a standalone package, these names made sense to me, but inside of the ecosystem of these packages I think the simpler name reads better.

Updated 👍

packages/presence-signal/package.json
EthanStandel: Done 👍

packages/presence-signal/README.md
EthanStandel: Good catch 😅  I think it was an autoformatting issue. This block was previously marked as `jsx` and not `tsx` but this line was supposed to be

```tsx
const [item, setItem] = createSignal<(typeof items)[number] | undefined>(items[0]);
```

Fixed 👍

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I do agree that `exitDuration` & `enterDuration` say everything that needs to be said, but I am concerned that `duration` might be too terse of an option property. Do you think it's obvious enough what it is a duration _of_?

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I find this case somewhat hard to recreate so I am a little cautious to change it. The logic comes from the original library (and as a memento, I have kept the property name the same). But are we sure that the reflow logic actually runs if you don't trigger a side-effect? From one of the original threads on StackOverflow that discussed this issue, they noted that the V8 itself would ignore this statement without any side effects (but of course, that answer just turned 6 years old at this point).

https://stackoverflow.com/a/21665117

❓

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I was thinking about that when I was converting this because I had the same thought in regards to the original library which had this logic in a `React.useEffect` which has similar constraints. What I realized is that React can be running in non-DOM clients, e.g. `react-native`. This is the only logic in this library that interacts with the DOM directly, so I like the idea of keeping the check. I notice that Solid itself even keeps it's client `render` function in `solid/web` which implies to me that Solid is built in such a way that there is the _potential_ of a `solid-native` concept in the future. Do you think that's enough of a reason to keep the check around ❓ 

It's just a thought, and not a hill I feel the need to die on.

packages/presence-signal/src/createPresenceSwitchSignal.ts
EthanStandel: I like that idea and it makes sense, however, I want to get the current version fully tested _and then_ start making changes to the core logic and validate against previous working tests. My main concern is the fact that lines 17 & 18 run next to each other here and if they were split up, I could see some potential jankiness being triggered from it.

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: Ah, this is just a side effect of my conversion over the original library. My methodology was to convert every piece of derived state into a function that returned that value. but that was entirely unecessary here as `initialEnter` is only ever used to initialize other signals. Fixed 👍

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I agree that the overloaded language is not optional. How would you feel about `preferredState` ❓

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: It hadn't actually crossed my mind that these could trigger updates that didn't need to happen, but you're definitely right. Fixed 👍 

Also did the same for the derived `exitDuration` & `enterDuration` signals seeing as they could potentially trigger reactive updates any time any of the options changed.

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I have updated `exitDuration` & `enterDuration` seeing as we agree on that change.

packages/presence-signal/src/createPresenceSwitchSignal.ts
EthanStandel: So I just added tests and started looking back at this idea, and I don't think it's possible. There are two states being synced here and I believe their relationships with other states requires an effect to keep synced.

## The reason we need an effect for `mountedItem`

The first line checks if `mountedItem() !== item()`, which inherently means that `mountedItem`'s future values can't be derived without being able to reference it's past values. So this flow has to be able to hold and reference previously existing state for that.

## The reason we need an effect for `shouldBeMounted`

The underlying `createPresence` call relies on the `shouldBeMounted` state, however, `createPresence` returns `isMounted` which the `shouldBeMountedState` itself utilizes to derive it's upcoming values. This seems like a chicken & the egg problem, but it's really just that `shouldBeMounted` is the true intended future state after all transitions whereas `isMounted` is the currently rendered state.

EthanStandel:
Thank you very much @thetarnav for the detailed review! I've responded to all comments and for the sake of quick-scan reading all of my resonses contain either 👍, ❓, or 🛑 emojis.

👍 means that I agree with the feedback and have updated accordingly
❓ means I have a follow up question about that feedback
🛑 means that I either disagree with the feedback or was unable to fulfill the request

Client & server tests have be added for `createPresence` & `createPresenceSwitch`, and so I have opened this up for review officially as a package proposal!

snnsnn:
I believe implementation could be simplified significantly like so:

```typescript
import {
  Accessor,
  createComputed,
  createSignal,
  on,
  untrack
} from 'solid-js';

interface Options {
  initialTransition?: boolean;
  duration: number | { enter: number; exit: number };
}

interface State {
  get isMounted(): boolean;
  get entering(): boolean;
  get entered(): boolean;
  get exiting(): boolean;
  get exited(): boolean;
}

type Status = 'entering' | 'entered' | 'exiting' | 'exited';

export function usePresence(show: Accessor<boolean>, options: Options): State {
  const { initialTransition, duration } = options;
  const enter = typeof duration === 'number' ? duration : duration.enter;
  const exit = typeof duration === 'number' ? duration : duration.exit;
  const initialVal = untrack(show) ? 'entered' : 'exited';
  const [state, setState] = createSignal<Status>(initialVal);

  const handleTransition = () => {
    const val = show();
    setState(val ? 'entering' : 'exiting');
    setTimeout(() => setState(val ? 'entered' : 'exited'), val ? enter : exit);
  };

  if (initialTransition) {
    handleTransition();
  }

  createComputed(on(show, handleTransition, { defer: true }));

  const getState: State = {
    get isMounted() {
      return state() !== 'exited';
    },
    get entering() {
      return state() === 'entering';
    },
    get entered() {
      return state() === 'entered';
    },
    get exiting() {
      return state() === 'exiting';
    },
    get exited() {
      return state() === 'exited';
    },
  };

  return getState;
}
```

Here state updates triggers single update, so we can avoid batching. 

The naming convention the react implementation uses a bit confusing,  I think clear and straightforward names are better.

You can see a working demo: https://stackblitz.com/edit/solidjs-templates-xbyq1m?file=src/use-presence.ts

atk:
Here the getters subscribe to any reactive change of state(), which is significantly less performant than using createSelector(). This is a low hanging fruit for optimization.

snnsnn:
> Here the getters subscribe to any reactive change of state(), which is significantly less performant than using createSelector(). This is a low hanging fruit for optimization.

Exactly. We can also encode status as union of numbers or single characters to shave off some bits. 

I wanted to see the initial reaction. I will open a PR if you or other maintainers say OK.

thetarnav:
I like the @snnsnn idea very much. We should aim to do as little work as possible inside these primitives, whether something should be kept in a separate signal to not over-execute is the users' decision. If we can maintain the state using a single signal, either by using an enum like @snnsnn suggested, or a plain object that is updated by restructuring.

We have utilities like [`createDerivedStaticStore`](https://github.com/solidjs-community/solid-primitives/tree/main/packages/static-store#createderivedstaticstore) or [`destructure`](https://github.com/solidjs-community/solid-primitives/blob/main/packages/destructure/README.md#destructure) to help memorize individual properties if needed.

As for the shape of the returned object, it either should be an object with getters, just like a store/props. Or a single signal (memo) returning the derived object state in an immutable way.

with getters
```ts
declare function createPresence(options): {
  readonly isMounted: boolean;
  readonly entering: boolean;
  readonly entered: boolean;
  readonly exiting: boolean;
  readonly exited: boolean;
}

const presense = createPresense(/* ... */)

createEffect(() => {
  console.log(presense.entering)
})

// manually memorize if needed
const entering = createMemo(() => presense.entering)
```

a single signal
```ts
declare function createPresence(options): Accessor<{
  readonly isMounted: boolean;
  readonly entering: boolean;
  readonly entered: boolean;
  readonly exiting: boolean;
  readonly exited: boolean;
}>

const presense = createPresense(/* ... */)

createEffect(() => {
  console.log(presense().entering)
})

// manually memorize if needed
const entering = createMemo(() => presense().entering)
```

packages/presence-signal/src/createPresenceSignal.ts
thetarnav: imo keeping it as `source` or `input` make sense as this is a reactive source that drives this primitive. Much like the `source` param in `createResource` / `createMemo`
https://github.com/solidjs/solid/blob/4dd99a7a590b5baa32175198c6c2013d6584c6fb/packages/solid/src/reactive/signal.ts#L569

snnsnn:
I like the idea of exposing the state directly:

```typescript
import { Accessor, createComputed, createSignal, on, untrack } from 'solid-js';

interface Options {
  initialTransition?: boolean;
  duration: number | { enter: number; exit: number };
}

type State = 'entering' | 'entered' | 'exiting' | 'exited';

export function usePresence(
  show: Accessor<boolean>,
  options: Options,
): [Accessor<boolean>, Accessor<State>] {
  const { initialTransition, duration } = options;
  const enter = typeof duration === 'number' ? duration : duration.enter;
  const exit = typeof duration === 'number' ? duration : duration.exit;
  const initialVal = untrack(show) ? 'entered' : 'exited';
  const [state, setState] = createSignal<State>(initialVal);

  const handleTransition = () => {
    const val = show();
    setState(val ? 'entering' : 'exiting');
    setTimeout(() => setState(val ? 'entered' : 'exited'), val ? enter : exit);
  };

  if (initialTransition) {
    handleTransition();
  }

  createComputed(on(show, handleTransition, { defer: true }));

  const isMounted = () => state() !== 'exited';

  return [isMounted, state];
}
```

Now we can take advantage of array destructuring:

```typescript
const [isMounted, state] = usePresence(show, {
  duration: 1000,
  initialTransition: true,
});
```

And about the memoization, we don't need memoization because transitioning from `false` to `false`, i.e `state() === 'entering' `, has no effect on the UI:

```typescript
<div
  style={{
    transition: 'all .5s linear',

    // Transitioning from false to false has no effect 
    ...(state() === 'entering' && {
      color: 'green',
    }),

    ...(state() === 'entered' && {
      color: 'red',
    }),
    ...(state() === 'exiting' && {
      color: 'orange',
    }),
    ...(state() === 'exiting' && {
      color: 'blue',
    }),
  }}
>
  Some Content!!
</div>
```

 Even if we use memo, we will run this computation inside the memo function, so in terms of net computation work, there is no difference. It is best to run the computation when it is needed rather than beforehand.

atk:
You could even use it in a branchless version

```ts
<Show when={isMounted()}>
  <div
    style={{
      transition: 'all .5s linear',
      color: ({
        entering: 'green',
        entered: 'red',
        exiting: 'orange',
        exited: 'blue'
      })[state()]
    }}
  >Some Content!!</div>
</Show>
```

I'd probably prefer a CSS variable, because that helps the browser optimize the color changes, but this should work, too.

snnsnn:
> I don't see a point in returning `isMounted`

Good point. I don't like redundancies either.

One last question, about the naming, presense is a facy name but does not convey the true nature of what we are doing here and is plainly confusing if you are not familiar with it. Can we come up with a better name?

thetarnav:
I'm only familiar with the naming from the work on [motionone](https://motion.dev/solid/presence), and here it's used to describe a similar pattern, so I quite like it.

snnsnn:
> I'm only familiar with the naming from the work on [motionone](https://motion.dev/solid/presence), and here it's used to describe a similar pattern, so I quite like it.

Name is good, memorable and distinct, my concern is its connotations, which is also stressed further by the proposed description, "A small SolidJS utility to animate the presence of an element"  gives you the idea of interacting with the underlying DOM element, all we do is delay the unmounting of the element through a derived signal.

thetarnav:
I'm looking at it as `crearePresence` as an implied shortcut for `createPresenceState`. A similar naming to the rest of the packages in this lib.

> A small SolidJS utility to animate the presence of an element

The usage with DOM elements should be only brought up as an example of one usecase. There are probably other places in the documentation to make the boundaries clear.

snnsnn:
> I'm looking at it as `createPresence` as an implied shortcut for `createPresenceState`. 

Yeah, this may make more sense since we are aiming to apply animations both when mounting and unmounting, while the name `extendPresence` applies only to the latter one.

OK, thanks. I will send a PR.

EthanStandel:
Apologies for not being in on this conversation all day. I've been trying to keep track of it, but I have an upcoming deadlines for official work stuff which generally makes me unavailable for OSS stuff during 9-5EST, but I wanted to make sure I got in on this discussion before a PR was opened against my fork.

I want to say that I have no problem with changing the API itself. The value of this package, to me, is to have an exposed state-primitive (or set of state primitives) that can represent the presence of elements/components/data. However, exactly how it is exposed to the end user is totally up for debate in this PR.

@snnsnn @thetarnav I do like the idea of turning this effectively into a strict state machine, but I also have a few concerns with how existing paradigms that are useful in [`solid-presence-signal`](https://www.npmjs.com/package/solid-presence-signal)/[`use-presence`](https://www.npmjs.com/package/use-presence) would be maintained. I like @snnsnn's StackBlitz example using text-color but I also think that the discussion has focused a bit to much on this as if it's a _use-case_ whereas I think that is more of a good _debugging tool_. I think a more realistic use-case for this is one that I have laid out in [the StackBlitz that I added to the README](https://stackblitz.com/edit/solidjs-templates-8eqpxz?file=src/App.tsx), where you have an element animating in stage-left and out stage-right. That only works because of an existing intermediate state where `isEntering` is `true` but `isVisible` is `false` which allows the element to first render in it's pre-entrance state with `opacity: 0; transform: translateX(-25px);`, at which point in the current implementation, we `requestAnimationFrame` and on that next frame we declare `isVisible` to be `true` which is the state that determines that we should actually be animating towards visibility. It doesn't appear that your state-machine model would ever render that intermediate state, thus making this kind of entrance impossible. I do have a few other concerns about the code in your first comment, but this is the most pressing issue.

While we're already considering changes to the API, I figure this one might be worth bringing to the table as well. I originally contributed the `usePresenceSwitch` hook to the [`use-presence`](https://www.npmjs.com/package/use-presence), and my original intention with that hook was for it to entirely replace `usePresence` because `usePresenceSwitch` is entirely capable of accepting a `boolean` as it's controlling state and handling the rest of the existing model the same. The owner of that repo preferred that it be written as a separate hook to keep the use-cases separate, so I went in that direction with the PR. When I forked the repo to make the original Solid based implementation, I intended to just keep the exact patterns and API. However, now that it's a larger community discussion I'd like to bring that topic back into frame: should `createPresence` & `createPresenceSwitch` just be merged and become a single export ❓

thetarnav:
> should `createPresence` & `createPresenceSwitch` just be merged and become a single export ❓

Honestly, it was one of my initial impressions when looking at the implementation as well. "Can I just pass a `boolean`?" So I'm glad you're bringing this up, and I'm down to go in this direction.
The only concerning part is extending the features of the primitive that may not be useful for everyone. But I think it's worth it considering that we get a single, unified interface to work with depending on our needs. Much more primitive-like™

And about the other proposed changes: the same things that were possible with the previous API, should be possible with this one too. When making it into a pure state-machine, how do we let the user force animations to happen, which is done by forcing a reflow internally now.
The idea of returning just a single enum signal is nice, but I'm not sure if this state can be expressed like this. Are the different "phases" really exclusive? Maybe is can be achieved with the help of some additional helpers as we did for `isMounted`. Not sure.

I think we need to establish what the primitive is supposed to do exactly, test it, and publish it as the initial implementation. Then we can switch the implementation and even the API details all we want, as long as the same things are still possible.
As the tests are in place, I think we are in a good spot? I guess that unit testing triggering reflows would be hard, so at least the demo should contain that functionality.

snnsnn:
>  you have an element animating in stage-left and out stage-right. 

@EthanStandel  Sorry but I could not understand your concern. State machine offers far greater flexibility than manually swapping states.

https://stackblitz.com/edit/solidjs-templates-2kjppq?file=src/App.tsx

In your implementation you have too many moving pieces and too many manual updates which will degrade performance considerably. Solid is fast but still lags if you have too many components. Also, you don't batch update, which triggers undesired intermediate state updates.

snnsnn:
> I think we need to establish what the primitive is supposed to do exactly, test it, and publish it as the initial implementation

Here is my understanding, which was written for the README file:

A small utility to have stronger control over an element's presense in the DOM in order to apply enter and exit transitions.

When an elements visibilty tied to a signal, the elements gets mounted and unmounted abruptly, not permitting to apply any css transitions.

```ts
<Show when={show()}>
  <div></div>
</Show>
```

Through `createPresence` we create a derived signal, through which we transition from `false` to `true` in two steps, `entering` and `entered`, and from `true` to `false` in two steps `exiting` and `exited`.  

```ts
const [show, setShow] = createSignal(true);

const state = usePresence(show, {
  duration: 500,
  initialTransition: true,
});

createEffect(() => console.log(state()));
```

This allows us to apply enter and exit transitions to an element but first we need to hand over the control of the element's visibilty to the `state`:

```ts
const isMounted = () => state() !== 'exited';
<Show when={isMounted()}>
  <div></div>
</Show>
```

Now, when `show` is set to `false`, `div` will not disappear immediately but wait for the `exiting` state to complete and move over to `exited`. In other words, we extended element's presence in the DOM for the duration of `exiting` state.

When `show` is set to `true`, `state` will move into `entering` and remain there for `500ms` then automaticall move to `entered` indefinitely. This will allow us to use show one style during `entering` and another one in `entered` state:

```ts
<div
  style={{
    transition: 'all .5s linear',
    
    ...(state() === 'entering' && {
      color: 'green',
    }),
    
    ...(state() === 'entered' && {
      color: 'red',
    }),
  }}
>
  Some Content!!
</div>
```

Setting `show` to `false` will move the state value `exiting` for `500ms` and then `exited`.

```ts
<div
  style={{
    transition: 'all .5s linear',
  
    ...(state() === 'exiting' && {
      color: 'orange',
    }),
    
    ...(state() === 'exited' && {
      color: 'blue',
    }),
  }}
>
  Some Content!!
</div>
```

EthanStandel:
@thetarnav I'd be happy to just combine them! When this was a React hook there was more risk of lost frames from triggering unnecessary state updates from the extra effect, but in Solid the user has more control over what a state update does and the state updates themselves are generally far less weighty. So I don't believe this actually is a relevant issue here.

@snnsnn Your understanding of the model is definitely correct, but the [new StackBlitz you shared](https://stackblitz.com/edit/solidjs-templates-2kjppq?file=src/App.tsx) isn't doing what you think it is. If you expand the transition time to be something longer, it's a little bit easier to see the problem.

When you trigger the presence of an element to enter, it's immediately rendering in a hidden state and then waiting `${enterDuration}ms` in that hidden state, and then only once you have it in the `"entered"` state does it actually begin transitioning in. So in your model, all entrance transitions actually take twice as long as they're supposed to. [Here's a fork with some small changes to make the problem a bit easier to see](https://stackblitz.com/edit/solidjs-templates-22zcmy?file=src/App.tsx).

I _think_ that if you just added _one_ more state (I think it's just one) to the flow, it could cover for this. [Here's a janky fork of with a potential fix](https://stackblitz.com/edit/solidjs-templates-5bxl1b?file=src/App.tsx). I got away with just awaiting `setTimeout` of `0` in this case rather than waiting for a whole animation frame like the original implementation, but I don't know if there are risks to this approach. It seems to work fine and would be overall more performant because the transition times wouldn't be lagged by a frame. The waiting out a full frame could have just been a React specific requirement.

snnsnn:
> When you trigger the presence of an element to enter, it's immediately rendering in a hidden state and then waiting ${enterDuration}ms in that hidden state, and then only once you have it in the "entered" state does it actually begin transitioning in. So in your model, all entrance transitions actually take twice as long as they're supposed to.

I see, timing and triggering transitions was different in my initial implementation but that is lost somewhere between the iterations. I will fix it. Thank you.

snnsnn:
Ok, I had time to look at it with a fresh look. The problem is, as you said, the element is mounted with the initial state and remain there until `entered` states kicks in, only then the css transitions applied, then we see the results.

In my earlier initial implementation (not shown in the discussion) did not suffer from this, because the element was mounted with `entering` style and the `entered` styles was flushed immediately, rather than waiting for the duration of the `enter` value like so:

```ts
setState('entering');
setTimeout(() => setState('entered'), 0);
```

I introduce this bug when I changed `0` to enter transition duration while I was trying to match your API.

Solution is simple, we add an initial state. Initial state should be there from the beginning because an element can be in one of these tree resting states, `initial`, `entered`, `exited` and two transition states, `entering` and `exiting`. Not having the `initial` state was a bug because we can not have two transition states between two resting states.

Now having the initial state, element will be mounted with `transform: 'translateX(-25px)'`, then will transition to ` transform: 'translateX(0)'`, that is our left-in transition :

```typescript
<div
  style={{
    background: 'gray',
    transition: 'all 3s linear',

    ...(state() === 'initial' && {
      "background-color": 'red',
      opacity: '1',
      transform: 'translateX(-25px)',
    }),
    
    ...(state() === 'entering' && {
      opacity: '1',
      "background-color": 'green',
      transform: 'translateX(0)',
    }),

    ...(state() === 'exiting' && {
      opacity: '1',
      "background-color": 'blue',
      transform: 'translateX(25px)',
    }),
  }}
>
  Some Content!!
</div>
```

When show is toggled off, state will move to `exiting`, moving the transform value to `transform: 'translateX(25px)',` and that is our right-out.

Without the `initial` state, we can not mount the element with the property of `transform: 'translateX(-25px)'`.  If we have it as an unconditional value like below, as soon we exit the transition state, element will move back to `-25px`:

```
<div
  style={{
    background: 'gray',
    transition: 'all 3s linear',
   transform: 'translateX(-25px)',
```

Your implementation also have similar issues. When we use the values from SecondExample in FirstExample, element does not have left-in right-out transition anymore.

https://stackblitz.com/edit/solidjs-templates-jpwjcz?file=src%2FApp.tsx

atk:
Since "initial" is basically an unmanaged state, it could also be an empty string, which would allow using the state as conditional, e.g. `if (state()) ...`.

EthanStandel:
@snnsnn I like this! I think this fulfills everything that I like about this model and with better performance! I would request that just for the sake of caution, we still have the reflow logic after `setState(visible ? 'initial' : 'exiting');`. And then also the `createPresence` hook is going to be merged with the `createPresenceSwitch` as well, which I'm sure will open up the small bit of logic left there to a bit more scrutiny. I intend to update this soon but this is a big deadline week in my dayjob which has been taking all my time.

If we agree that this is the ideal model going forward, do you want to open up a PR to my fork, or should we focus more on merging what I have first and then you can open up a PR with a major version change to update & solidify the API? I just want to make sure that you're credited for your work 😄 

@atk I personally prefer that it be legible and clear with exactly what that state is because it is distinctly its own state. I'm not 100% sure about `"initial"` but I think that state name is more clear than anything else I can think of 😅

snnsnn:
@EthanStandel This still feels icky for me because I feel like `createPresence` function should extend element's lifetime and it should not concern itself with transition duration because that is what we have css for. This will limit the proliferation of states and execute one less timeout and the logic for createPresence will be clear, extending element's lifetime for a specified duration.

I know this issue has been dragged so long, but css is tricky, os is coming up with good API.

Check this out:

```typescript
export function usePresence(
  show: Accessor<boolean>,
  options: Options
): Accessor<State> {
  const { initialTransition, duration } = options;
  const initialVal = untrack(show) ? 'entered' : 'exited';
  const [state, setState] = createSignal<State>(initialVal);

  let timeout: number;
  const handleTransition = () => {
    const val = show();
    clearTimeout(timeout);
    setState(val ? 'entering' : 'exiting');
    timeout = setTimeout(
      () => setState(val ? 'entered' : 'exited'),
      val ? 0 : duration
    );
  };

  if (initialTransition && initialVal === 'entered') {
    handleTransition();
  }

  createComputed(on(show, handleTransition, { defer: true }));
  return state;
}
```

https://stackblitz.com/edit/solidjs-templates-nrye3h?file=src/use-presence.ts

Here we flush `entered` state immediately with 0, let the css decide for transition duration.

snnsnn:
> I would request that just for the sake of caution, we still have the reflow logic after setState(visible ? 'initial' : 'exiting');

Do you mean if we flush our component with transition state immediately? React's version mounts elements then sets to updating css properties in two steps, but we mount element with initial proprerties, then we flush our component with transition.

EthanStandel:
@snnsnn I don't mean anything from a state/Solid perspective. I mean it specifically from a DOM perspective to ensure that the `"initial"` state truly gets painted because if it doesn't then the incoming transition will fail. That's why the initial `use-presence` library has [this line](https://github.com/amannn/react-hooks/blob/main/packages/use-presence/src/usePresence.tsx#L77). Here's a [SO thread on the topic](https://stackoverflow.com/questions/21664940/force-browser-to-trigger-reflow-while-changing-css).

I waffled on getting rid of the reflow when I first converted `use-presence` to Solid because we're waiting out a full animation frame in the interstitial state, so it should reliably update. But in your version we're only waiting out a single event-loop which will give the browser far less time to paint so I think having the reflow logic will be even more critical in your model.

snnsnn:
> mean it specifically from a DOM perspective to ensure that the `"initial"` state truly gets painted

Element is mounted with `initial` if placed under Show but I will look into cases where Show is not used and element kept being monted but gets hidden and visible between updates.

snnsnn:
I am somewhat satisfied with the result but still iffy about the passing enter and exit transition durations to `createPresence` because in my mind `createPresence` should deal with extending an element's lifetime for a specified duration and that is all it should do. The timing of the transitions and other css related stuff should be left to the css because there is no way to enforce transition durations via `createPresence`, we have to use css properties for transitions anyway. 

You may have notice I had a similar comment deleted but here writing it again.

If we use `createPresence` for extending element's lifetime only, we can mount the element in `entering` state and flush `entered` state immediately using setTimeout 0:

```typescript
setState('entering');
setTimeout(() => setState('entered'), 0);
```

Then leave everything to css, since css transition will take effect and animate towards the desired state (`entered`). 

If element does not have animation set, there won't be any inconsistencies because element is actually entered to the DOM. The point I am trying to make is, element's css will never be in an inconsistent state.

This way, the state reflects the actual status of the element, and we won't loose any functionality compared to the implementation we agreed upon, the one with the `initial` state. 

On the contrary, our implementation will have one less `setTimout` and our api will be cleaner and more straightforward. Plus, we will get rid of the confusion and discrepancy (because we have to manually adjust values both in css and js sides and keep them in sync)  brought by the transition values we pass to `createPresence` . 

```typescript
type State = 'entering' | 'entered' | 'exiting' | 'exited';

export function usePresence(
  show: Accessor<boolean>,
  options: Options
): Accessor<State> {
  const { initialTransition, duration } = options;
  const initialVal = untrack(show) ? 'entered' : 'exited';
  const [state, setState] = createSignal<State>(initialVal);

  let timeout: number;
  const handleTransition = () => {
    const val = show();
    clearTimeout(timeout);

    setState(val ? 'entering' : 'exiting');
    timeout = setTimeout(() => setState(val ? 'entered' : 'exited'), val ? 0 : duration);
  };
 
  // Animate element only when `show` set to true
  if (initialTransition && initialVal === 'entered') {
    handleTransition();
  }

  createComputed(on(show, handleTransition, { defer: true }));
  return state;
}
```

Here is a working demo: https://stackblitz.com/edit/solidjs-templates-nrye3h?file=src%2Fuse-presence.ts

thetarnav:
I like the API of `transition-group` primitives where the transition can be ended by calling the `done` function.
https://github.com/solidjs-community/solid-primitives/tree/main/packages/transition-group#createswitchtransition
So you can use a timeout, wait for transitionend event, or call it immediately if you want to skip the transition compleately.
Then `solid-transition-group` has a flexibility to choose when the transition should be ended:
https://github.com/solidjs-community/solid-transition-group/blob/main/src/common.ts#L60

snnsnn:
> I like the API of `transition-group` primitives where the transition is ended by calling the `done` function.

It looks very complex. I would prefer current implementation. Take a look at this:

```typescript
<div
  classList={{
    'animate-in': state() === 'entering',
    'animate-out': state() === 'exiting',
  }}
>
  Some Content!!
</div>
```

I would keep a simple internal state and and expose a handle to drive the progress or mark it done rather than setting listeners.

thetarnav:
@EthanStandel do you want to add any more changes before the initial release?
There is no point in keeping deferring the release and I don't mind merging this as it is now. It'll be easier to iterate in separate PRs and improving over time later as we use the primitive and learn from it.
But you mentioned the idea of combining `createPresence` and `createPresenceSwitch` together, and there are few unresolved comments from earlier.
Let me know what you think.

EthanStandel:
Sorry everyone, this week was a product release week for a client. The past two weeks have been really busy, but I'm going to put out some updates tonight to get it in an initial place that I'm happy with.

@snnsnn I do get the awkwardness & dislike of having to declare the transition time in both the primitive and in your component. My reccomendation for how people should use this (and how I've used `use-presence`) is as primitive to create re-usable components, but I wouldn't recommend sprinkling calls to `createPresence` all over your code.

However, if you don't actually wait for the transition time before setting the state to `"entered"` then we never capture the state that would identify if you should be currently animating. Capturing that state isn't useful in these cases we've talked about, but it would be useful if you were doing some kind of tweening like animating an up-count of a number to a particular target. I think there are reasons that this state is worth supporting. And seeing as you have to pass a duration for the exit flow anyways, I don't think it's worth removing the entering state for the entrance flow just because we can away with it in this particular use-case.

@thetarnav I agree with @snnsnn that I find that model in `transition-group` (and `react-transition-group`) to be a less optimal and part of my motivation for this package in the first place. I really like the idea of using the state primitives themselves to manage the current state of an animation, as opposed to having something more akin to events firing & handler. Imo, `transition-group` feels more imperative and `presence` feels more declarative.

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I'll go with `source` for the moment on this one. Though per our discussion in a future commit, I'm just going to merge them anyways, in which case this argument will just become `item`.

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I think that's reasonable, just wanted to double check. I'll go with the suggestion.

packages/presence-signal/src/createPresenceSignal.ts
EthanStandel: I'll remove the document validation from this and leave the reflow check. I think in the future we could consider removing it or not as this moves to the state machine model. That model, I think, would be easier for the user to introduce a reflow if they need it (though I lean towards keeping it generally, just because I don't want users to ever have to really consider it).

EthanStandel:
@thetarnav I put through updates to address our discussions. I chose not to _merge_ `createPresence` & `createPresenceSwitch`, but instead to just rename `createPresence` to `createPresenceBase` and `createPresenceSwitch` to `createPresence`, store them both in `createPresence.ts` and only export (the new) `createPresence`. I updated all tests, the dev app, readme, and package.json's `primitive.list` to reflect this change. I definitely think in the future these should be constructed as a single utility function, but I think it's worth waiting to do that for when @snnsnn takes a swing at the state-machine based API.

I also had previously only validated whether a switching item should be mounted based on `item() !== undefined`. I had to change this to also check `item() !== false`. I ended up going with this:

```ts
const itemShouldBeMounted = <ItemType>(item: ItemType) => item !== false && item != null;
```

With that, in my opinion this is ready for an initial stage 0 release. Let me know if it needs anything else!

snnsnn:
@thetarnav I have to travel on a short notice but I will send a PR for basic functionality this weekend. We can improve on it later on.

PR #418, opened 2023-04-25
Websocket rewrite

atk:
Rewrote the primitive, with a lot of simplification and improved composability.

### DIFF
--- a/.changeset/five-plums-yawn.md
+++ b/.changeset/five-plums-yawn.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/websocket": patch
+---
+
+websocket: rewrite
--- a/packages/websocket/package.json
+++ b/packages/websocket/package.json
@@ -18,12 +18,11 @@
     "list": [
       "makeWS",
       "createWS",
-      "createMessageWS",
       "makeReconnectingWS",
       "createReconnectingWS",
       "makeHeartbeatWS"
     ],
-    "category": "network"
+    "category": "Network"
   },
   "private": false,
   "type": "module",
@@ -44,7 +43,7 @@
   "sideEffects": false,
   "scripts": {
     "dev": "vite serve dev",
-    "page": "vite build dev",    
+    "page": "vite build dev",
     "build": "jiti ../../scripts/build.ts",
     "test": "vitest"
   },
--- a/packages/websocket/package.json
+++ b/packages/websocket/package.json
@@ -18,6 +18,7 @@
     "list": [
       "makeWS",
       "createWS",
+      "createMessageWS",
       "makeReconnectingWS",
       "createReconnectingWS"
     ],
--- a/packages/websocket/package.json
+++ b/packages/websocket/package.json
@@ -3,7 +3,9 @@
   "version": "0.4.0",
   "description": "Primitive to create a web socket connection",
   "author": "David Di Biase <dave.dibiase@gmail.com>",
-  "contributors": ["Alex Lohr <alex.lohr@logmein.com>"],
+  "contributors": [
+    "Alex Lohr <alex.lohr@logmein.com>"
+  ],
   "license": "MIT",
   "homepage": "https://github.com/solidjs-community/solid-primitives/tree/main/packages/websocket",
   "repository": {
@@ -39,7 +41,8 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "build": "jiti ../../scripts/build.ts"
+    "build": "jiti ../../scripts/build.ts",
+    "test": "vitest"
   },
   "keywords": [
     "websocket",
--- a/packages/websocket/package.json
+++ b/packages/websocket/package.json
@@ -1,10 +1,26 @@
 {
   "name": "@solid-primitives/websocket",
-  "version": "0.3.7",
-  "description": "Primitive to create a worker helper",
+  "version": "0.4.0",
+  "description": "Primitive to create a web socket connection",
   "author": "David Di Biase <dave.dibiase@gmail.com>",
+  "contributors": ["Alex Lohr <alex.lohr@logmein.com>"],
   "license": "MIT",
   "homepage": "https://github.com/solidjs-community/solid-primitives/tree/main/packages/websocket",
+  "repository": {
+    "type": "git",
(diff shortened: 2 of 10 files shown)


davedbase:
Thanks for tackling this Alex! A couple thoughts I wanted to throw in:

- Is the goal with the rewrite to take an approach similar to fetch?
- Should we consider adding a heartbeat option? VueUse and other hook libs seem to have it.
- Should we consider adding `delay` to WSReconnectOptions?
- What do you think about an interface that returns `const { status, data, send, open, close } = createWS()`?

atk:
One of the main drivers for the rewrite was the attempt to simplify things, once through exposing the actual web socket instance, but also by attempting to increase the composability.

I'll take some time to think about how to integrate the extra features in a minimal, composable way.

Also, there's already a delay, but it's called timeout. I should probably rename it, delay is more fitting.

atk:
> Is the goal with the rewrite to take an approach similar to fetch? ie. adding modifiers. I really like that pattern.

Actually, I found the modifier pattern has its place, but more in larger libraries and less in primitives. I prefer the composability of something like our resource package.

> Curious what you think might be the downside of an interface of const { status, data, send, open, close } = createWS()? I suppose the pattern you opted for is more in line with some of the other primitives though.

This needlessly adds size and abstracts the actual web socket away, so you cannot directly influence it. I think this is a bad thing.

I have made the first steps towards better composability and an improved DX for the reconnecting web sockets. I want to add an error signal wrapper, but I'm not sure about its name yet.

Also, still have to look into a way to support that heartbeat option.

.changeset/five-plums-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@solid-primitives/websocket": patch
```
thetarnav, line 2: this doesn't look like a patch lol
Why not make it a `major` since it's a breaking rewrite?
```suggestion
"@solid-primitives/websocket": major
```

.changeset/five-plums-yawn.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+websocket: rewrite
```
thetarnav, line 5: might be worth including new features and others that maybe got removed / work differently

packages/websocket/package.json
```diff
@@ -1,10 +1,30 @@
 {
   "name": "@solid-primitives/websocket",
-  "version": "0.3.7",
-  "description": "Primitive to create a worker helper",
+  "version": "0.4.0",
```
thetarnav, line 3:
```suggestion
  "version": "0.3.7",
```
let the changeset do the work ;)

packages/websocket/package.json
```diff
@@ -1,10 +1,30 @@
...
+    "stage": 0,
+    "list": [
+      "makeWS",
+      "createWS",
+      "createMessageWS",
+      "makeReconnectingWS",
+      "createReconnectingWS",
+      "makeHeartbeatWS"
+    ],
+    "category": "network"
```
thetarnav, line 26:
```suggestion
    "category": "Network"
```
not sure if our categories are case sensitive but better to stay consistent

packages/websocket/src/index.ts
```diff
@@ -1,72 +1,199 @@
...
  * ```ts
- * const [ connect, disconnect ] = createWebsocket('http://localhost', '', 3, 5000);
+ * const ws = createMessageWS(makeWS("ws://localhost:5000"));
+ * createEffect(on(ws.message, (msg) => console.log(msg)), { defer: true });
  * ```
  */
-const createWebsocket = (
+export const createMessageWS = (
+  ws: WebSocket
+): WebSocket & { message: Accessor<WSMessage | undefined> } => {
```
thetarnav, line 40: Cannot it just return the message instead of mutating the input? This seems like an odd pattern considering that the type won't be reflected in the original object.
```ts
const ws = makeWS("ws://localhost:5000")
createMessageWS(ws)
ws.message() // works, but ts doesn't know about it
```

It also doesn't support setting initial value, which could be useful.

Also, we have `createEventSignal` in `@solid-primitives/event-listener`. Why not promote using that as it can be used with any event target?

```ts
import { createEventSignal } from "@solid-primitives/event-listener"

const ws = makeWS("ws://localhost:5000")
const message = createEventSignal(ws, "message")

message() // WSMessage | undefined
```

packages/websocket/src/index.ts
```diff
@@ -1,72 +1,199 @@
...
+ */
+export const createReconnectingWS = (
+  url: string,
+  protocols?: string | string[],
+  options?: WSReconnectOptions
+) => {
+  const ws = makeReconnectingWS(url, protocols, options);
+  onCleanup(() => ws.close());
+  return ws;
+}
```
thetarnav, line 151: I like the strong make vs create separation 💪
btw you can type it using `typeof makeReconnectingWS` if youre making the exact same interface. Best way to ensure they are in sync
```suggestion
export const createReconnectingWS: typeof makeReconnectingWS = (url, protocols, options) => {
  const ws = makeReconnectingWS(url, protocols, options);
  onCleanup(() => ws.close());
  return ws;
}
```

packages/websocket/package.json
```diff
@@ -1,10 +1,30 @@
...
+  "primitive": {
+    "name": "websocket",
+    "stage": 0,
+    "list": [
+      "makeWS",
+      "createWS",
+      "createMessageWS",
+      "makeReconnectingWS",
+      "createReconnectingWS",
+      "makeHeartbeatWS"
```
thetarnav, line 24:
```suggestion
      "createWS",
      "createMessageWS",
      "createReconnectingWS",
      "makeHeartbeatWS"
```
Not sure if there is a point in listing the non-reactive variants here.
I see them more as additions.
Like you focus on `createWS` as the default, and say btw there is also a variant that doesn't autocleanup.

atk: I'm considering marking those primitives that work with other frameworks with solid in order to ease the migration of applications towards solid.

packages/websocket/src/index.ts
atk: I removed this one and settled on documenting the use of `createEventSignal`.

thetarnav (approved):

PR #425, opened 2023-05-06
Add start primitive

Tommypop2:
Added a primitive for solid start

Currently only contains createUserTheme, which, given a theme type, will handle isomorphic cookie parsing to retrieve the user's current theme


packages/start/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+export const createUserTheme = <T extends String>(
+  defaultTheme: T,
+  options?: UserThemeOptions,
+): [get: Accessor<T>, set: Setter<T>] => {
+  const { cookieMaxAge = 365 * 24 * 60 * 60, cookieName = "theme" } = options ?? {};
+  const event = useRequest();
+  const userTheme = parseCookie(
+    isServer ? event.request.headers.get("cookie") ?? "" : document.cookie,
+  )["theme"] as T | undefined;
```
thetarnav, line 25: Shouldn't this depend on the input?
```suggestion
  )[cookieName] as T | undefined;
```

packages/start/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @return Returns an accessor and setter to manage the user's current theme
+ */
+type UserThemeOptions = {
+  cookieMaxAge: Number;
+  cookieName: String;
+};
+export const createUserTheme = <T extends String>(
+  defaultTheme: T,
+  options?: UserThemeOptions,
+): [get: Accessor<T>, set: Setter<T>] => {
```
thetarnav, line 20:
```suggestion
): Signal<T> => {
```

thetarnav, line 20:
```suggestion
export const createUserTheme = (
  defaultTheme: string,
  options?: UserThemeOptions,
): [get: Accessor<string>, set: Setter<string>] => {
```
Can we be sure that the value received from a cookie will always be what we specified in the generic? No clue how it works, I guess we control both sides, but it feels like something that should be typed vaguely as `string` so that we force users to validate it. Or require a validation function in the options to be able to specify a generic.
But maybe it's fine, not sure

thetarnav (commented):
That looks like a good start, do you want to add docs for it too?

PR #426, opened 2023-05-07
Render playgrounds with solid-start site

thetarnav:
It shouldn't be too hard to be able to render our package playgrounds using the solid-start website in `/site`. Then we can develop the primitives while being able to see how they behave in SSR.
HMR seems to work fine too
Currently the playgrounds are using unocss for styling, which has to be moved to tailwind to be in sync with the site (I don't think we can move the tailwind config to unocss, am I right @aquaductape ? ) One issue with that is tailwind in JIT mode will only look for classes to load in the /site/src folder, and ignore the playgrounds. So maybe it's safer to just style with .scss in the playgrounds? Or be sure to add every used class to the safelist 🤷‍♂️
Unocss wouldn't have this problem I think 😕
There needs to be a way to start the development server directly from the package folder probably


thetarnav:
It actually builds 🎉
Here is a preview: https://645794d4077e6b592e2b1c26--solid-primitives.netlify.app/
You can access playgrounds by going to `/playground/pagination` or others
As you see the styles are not too hot

davedbase:
Woah! This is freaken amazing :) awesome work!

aquaductape:
This super sick! Back when I was building the site I wanted to use UnoCSS to build the site, but I was afraid to hit unforeseen problems and slow me down. But now UnoCSS has improved and fixed a lot of bugs since then.

I'll work on replacing tailwind with UnoCSS for the site x)!

thetarnav:
@aquaductape do you want to do that after or before this PR gets merged?
I added the playgrounds to the tailwind content list and it works.

image

Previously the buttons had a default browser style, and now the styles are being reset yo nothing so buttons without any styling look like this.
But what do we want to do about it? Add a default style for buttons? Add `.btn` class with some default styles? Export a button component from somewhere with the styles that can be used in the playgrounds?
Adding default styles to `<button>` and others would be the easiest for now, just making sure that it doesn't mess with the site styles.

image

aquaductape:
I'd prefer not to add default style to `<button>` globally in site styles, but instead group it with a selector where the playground is appended to like so.
```css
.playground-container button {
  /* ... */
}
```
This is so that the packages dev source should remain that way so that it's easier for authors to write the markup

thetarnav:
New deploy: https://645d7db5bcf1711ec4a04ca5--solid-primitives.netlify.app

Anything else to do?

atk:
I understand why you removed the page scripts, but why the dev scripts, too? Those were really helpful during development.

thetarnav:
@atk I agree, but how can the dev scripts be written now that you don't run a separate dev server in each package, but just run dev from root and work using the start site?
How should it be handled? Running dev in a package would somehow start the dev server from root and then navigate to a correct playground url?
I'll experiment with something today, but let me know if you have some ideas.

PR #429, opened 2023-05-10
i18n: Add useScopedI18n(scope: string)

mokeyish:
Add `useScopedI18n`, so that we can:

```ts

const dict = {
  en: { 
    login: {
      username: 'User name',
      password: 'Password',
      login: 'Login'
  },
  fr: {
     ...
  }
 }
}

export const LoginView () => {
  const [t] = useScopedI18n('login');
  return <>
      <div>{t('username')}<input /></div>
      <div>{t('password')}<input /></div>
      <button>{t('login') }</button>
  </>
}
```

### DIFF
--- a/.changeset/kind-maps-notice.md
+++ b/.changeset/kind-maps-notice.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/i18n": minor
+---
+
+Add `useScopedI18n` helper for creating a module-specific translate
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -142,6 +142,36 @@ const App: Component = () => {
 </I18nContext.Provider>;
 ```
 
+## `useScopedI18n`
+
+Use `useScopedI18n` to create a module-specific translate.
+
+```tsx
+
+const dict = {
+  en: { 
+    login: {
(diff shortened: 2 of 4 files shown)


thetarnav:
hello, that's an nice addition. I like the idea very much.
But since this feature doesn't require access to any internals and can be implemented on top of the existing API, I think it could be exported as a separate `const useScopedI18n = createScopedI18n(I18nContext)` function. Or added to the docs as a snippet to copy, "This is how you can make a scoped useI18n function for yourself" or something.
What do you think?

thetarnav:
Yup, that looks great.
Here is another idea since we only need to wrap the translate function:
```ts
export const LoginView () => {
  const [t] = useI18n();
  const loginT = scopedTranslate(t, 'login')
  return <>
      <div>{loginT('username')}<input /></div>
      <div>{loginT('password')}<input /></div>
      <button>{loginT('login') }</button>
  </>
}
```
But it may be more confusing than helpful.

Could you include the new API in the docs? Then I can go ached and release this

mokeyish:
I think it would be better to `useScopedI18n` directly, because it only needs one line of code to use.

PR #432, opened 2023-05-11
i18n: Add typings for `useI18n`

mokeyish:
Add typings for `useI18n`, so that we can:

图片

图片


thetarnav (approved):
Alright this looks great, thanks for improving the i18n package so much.

I think that we are missing a type/utility for creating typesafe dictionaries though. Like there is nothing checking if you have matching translations between languages. This could be done with a simple type + `satisfies` or a identity function.
But that's unrelated to the PR so I'm gonna merge this one.

PR #434, opened 2023-05-12
graphql: fix resource states

fogzot:
The current code passes `initialValue` unconditionally to the `createResource()` function, forcing an initial value even if the corresponding argument is undefined. This is bad because the sequence of states the resource goes through is:

* refreshing
* ready

while with an undefined `initialValue` the sequence is expected to be:

* pending
* ready

This patch fixes this problem and restores the correct resource states by passing a third argument to `createResource()` only when `initialState !== undefined`.

### DIFF
--- a/.changeset/sixty-dryers-float.md
+++ b/.changeset/sixty-dryers-float.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/graphql": patch
+---
+
+fix resource states by omitting `initialValue` option if passed `undefined`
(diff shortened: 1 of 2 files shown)


fogzot:
I did not added a changeset because I don't know if this is just a patch or a major change. Potentially it can break any code that somehow depend on never reading `pending' as the initial state.

thetarnav:
This is interesting
I wonder if this should be considered a bug of `createResource`?

thetarnav:
Ok will merge this if it turns out to be intended behavior.

Also, what do you think about allowing users to pass the whole resource options object, instead of just the initial value? It would solve this issue as well.

fogzot:
> Also, what do you think about allowing users to pass the whole resource options object, instead of just the initial value? It would solve this issue as well.

I would be in favor of this. But I think that writing a backward-compatible patch would be impossible, because `initialValue` can be any object (even an object that is shaped like `options`),

PR #437, opened 2023-05-15
Refactors prensence package to SM implementation

snnsnn:


packages/presence/README.md
```diff
@@ -32,143 +19,184 @@ yarn add @solid-primitives/presence
...
-```tsx
-const FirstExample = () => {
-  const [showStuff, setShowStuff] = createSignal(true);
-  const { isVisible, isMounted } = createPresence(showStuff, {
-    transitionDuration: 500,
-  });
+```typescript
+const App = () => {
+  const [show, setShow] = createSignal(true);
+  const state = createPresence(show, { duration: 300, initialRun: true });
```
atk, line 29: Since we need it in any case, wouldn't it be better to provide `state.mounted()`, which you can use as `<Show when={state.mounted()}>`?

atk (changes requested):

davedbase:

PR #449, opened 2023-05-21
Add useUserAgent

Tommypop2:
- Added useUserAgent primitive, which allows for the user to retrieve the user agent string on the client or the server
- Restructured code slightly, so that different start primitives are located within different files


packages/start/src/userAgent.ts
```diff
@@ -0,0 +1,13 @@
+import { useRequest } from "solid-start/server";
+import { isServer } from "solid-js/web";
+/** A primitive that allows for the user agent string to be accessed isomorphically on the client, or on the server
+ * @return Returns the user agent string, or null
+ */
+export function useUserAgent() {
+  const event = useRequest();
```
atk, line 7: This only works in Solid-Start. Please provide a way for users to add their own function that provides the request headers and keep useRequest as a default.

atk (changes requested):

packages/start/src/userAgent.ts
thetarnav: I think that was the point
To have a package with hard start dependency

atk: OK, everyone else can use the storage primitive anyways.

atk (approved):

Tommypop2:
This PR is quite old now and relies on old start, so shouldn't be merged in its current state. I'll happily update it to new start if there's still interest in this functionality though!

davedbase:

davedbase:

PR #452, opened 2023-05-28
Add `createLens` primitive

### LINKED ISSUE #453: Add `createLens` primitive

Utilities for working with nested reactivity in a modular way.

See previous discussion:
- https://github.com/solidjs/solid/issues/506
- https://github.com/solidjs/solid/discussions/944

Originally described by @clinuxrulz:

> Say our application state holds a collection of states of things in the tabs of the application. It can be handy to have a Lens to focus on a subset of the state in order to provide a store for use by the UI in the tabs state. So that rhe end-user does not need to alway push state changes to top level (under the hood the state still reaches the top, it is just hidden boiler plate).

### Suggest A Solution

Proposed primitives:

- `createLens` - Given a path within a Store object, return a derived or "focused"
getter and setter pair.

- `createFocusedGetter` - The first half of the lens tuple; a derived Signal
using path syntax on an object.

- `createFocusedSetter` - The second half of the lens tuple; a Setter
for a specific path within a Store.

I've started a PR to explore this concept (WIP): #452

It's essentially just syntactic sugar on top of regular Stores, with some tricks to maintain Typescript support.

In previous discussion people discussed whether there could be a performance gain to this approach as well, by caching a partially applied path within the Store object -- for example: https://github.com/solidjs/solid/issues/506#issuecomment-869300170.

I'm not as familiar under the hood of how the Store works, so my goal will be the replicate the intended DX and then unit test thoroughly enough that someone else could come in and work some performance magic. For now it is Lens-like DX with the same behavior/performance as a regular Store.

I'm also curious to get people's thoughts on:

- Does this seem useful?
- Are the types correct?
- What tests are missing?

Cheers 🍺

nathanbabcock:
Utilities for working with nested reactivity in a modular way.

- `createLens` - Given a path within a Store object, return a derived or "focused"
getter and setter pair.

- `createFocusedGetter` - The first half of the lens tuple; a derived signal
using path syntax on an object.

- `createFocusedSetter` - The second half of the lens tuple; a Setter
for a specific path within a Store.

## How to use it

```ts
// Start with an ordinary SolidJS Store
const storeTuple = createStore([
  { myString: 'first' }
])

// Create a lens to focus on one particular item in the Store.
// Any valid path accepted by `setStore` works here!
const [firstString, setFirstString] = createLens(storeTuple, 0, myString)

// Setters and Getters work just like ordinary Signals
setFirstString("woohoo") // equivalent to `setStore(0, "myString", "woohoo")
console.log(firstString()) // "woohoo"

```

### 1. Separation of Concerns

Components can receive scoped Setters for only the parts of state they need
access to, rather than needing a top-level `setStore` function.

### 2. Type-safety

Essentially, we are just [partially applying](https://en.wikipedia.org/wiki/Partial_application) a `setStore`
function with an initial path, and returning a function that will apply the
remainder of the path. It is just syntactic sugar, and under the hood
everything is using calls to native Store functionality.

The same approach can already be used by the Setter returned by `createStore`. However,
Typescript users will find it hard to maintain type-safety for the arguments
passed to a "derived"/partially-applied Setter. The type definitions for `SetStoreFunction` are...
[daunting](https://github.com/solidjs/solid/blob/44a0fdeb585c4f5a3b9bccbf4b7d6c60c7db3ecd/packages/solid/store/src/store.ts#L389).

The `lenses` package alleviates this friction by providing both `StorePath<T>`
and `EvaluatePath<T, P>` generic type helpers.

### 3. Shared path syntax between Getters and Setters

The path syntax defined in Solid Stores is incredibly expressive and powerful.
By introducing `createScopedGetter`, the same syntax can be also be used to
access Store values as derived Signals. This is particularly relevant to
child components which may both display and modify items from a Store
collection.

Closes #453

### DIFF
--- a/packages/lenses/.gitignore
+++ b/packages/lenses/.gitignore
@@ -1,0 +1,2 @@
+tsconfig.vitest-temp.json
+# (created while `vitest typecheck` is running in watch mode)
--- a/packages/lenses/LICENSE
+++ b/packages/lenses/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/lenses/README.md
+++ b/packages/lenses/README.md
@@ -1,0 +1,103 @@
+<p>
+  <img width="100%" src="https://assets.solidjs.com/banner?type=Primitives&background=tiles&project=lenses" alt="Solid Primitives lenses">
+</p>
+
+# @solid-primitives/lenses
+
+[![turborepo](https://img.shields.io/badge/built%20with-turborepo-cc00ff.svg?style=for-the-badge&logo=turborepo)](https://turborepo.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/lenses?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/lenses)
+[![version](https://img.shields.io/npm/v/@solid-primitives/lenses?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/lenses)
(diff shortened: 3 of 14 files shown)


nathanbabcock:
Source code is inside `packages/lenses`: https://github.com/nathanbabcock/solid-primitives/tree/lens/packages/lenses

thetarnav:
This is interesting, but why not rely on the "native" store scoping ability:
```ts
const [store, setStore] = createStore([
  {
    inner: {
      innerString: "first",
      innerNumber: 0,
    },
  },
])

const [inner, setInner] = createStore(store[0].inner)

setInner("innerString", "hello") // reflected in both stores as they share the same reference.
```
With stores you can rely on the references being the same, so any mutations to one will be reflected in all the other proxies that wrap the object.

I've replaced the `createLens` body with this, and all tests have passed:
```ts
const last = path[path.length - 1];
path = path.slice(0, -1);

const storeNode = store[0] instanceof Function ? store[0]() : store[0];
const value = getValueByPath(storeNode as StoreNode, path) as V;

const [get, set] = createStore(value as any);

return [() => get[last], (...args: any[]) => set(last, ...args)];
```
I don't hate the idea, but I want to know what exactly is the benefit of these primitives, as creating basic lenses is already in reach.

nathanbabcock:
I had no idea `createStore` could be used in this way. For some reason I thought it would double-wrap in a Proxy, or the references wouldn't stay consistent over the lifetime of the store.

I'm pretty sure that does everything I need it do unless I'm missing something.

clinuxrulz:
@thetarnav 
Thank you so much for that. I found I had a need for lenses on stores again, and had no idea ```createStore``` could be used like that. Was thinking using it like that would be invalid (double wrap).

That ```createStore``` lenses trick should be documented somewhere, like in the documentation for stores.

MrJohz:
@thetarnav, that won't work if the "lens" part can have a primitive value, right?  At least, Typescript complains about that case, and I can't see how that would work.

Here's an example from a codebase I'm working on at the moment:

```typescript
const [editedGraph, setEditedGraph] = createStore<{ graph: Graph | null }>({
    graph: null,
  });
```

Here, the `graph` attribute can be `null` (if there is no graph to be edited).  I can't just have `createStore(null)` because `null` is a primitive value and not an object, and so the store proxy magic wouldn't work, hence the wrapping in an object with the `graph` attribute.

I would like to be able to use lenses to hide this implementation detail from consumers.  In this situation, doing `createStore(editedGraph.graph)` won't work for the same reason that `createStore(null)` wouldn't work.  Hence the need for lenses that can handle this case.

The getter side is easy most of the time, because you can just wrap the `.graph` access in a function, or use a JS property.  And the setter side is easy in Javascript (i.e. without types) because you can wrap the property path magic:

```javascript
const setter = (...args) => setEditedGraph("graph", ...args);
```

But it's very difficult to type this function properly in such a way that it still behaves exactly like the normal `setStore` function (i.e. that you can call it with the property path, or pass different setter functions in).

Something like `createLens` would be really useful for this case where I want to wrap `setStore` with some specific path prefix, but I still want typing to work as expected.

PR #454, opened 2023-06-02
input-mask: new pattern display primitive

atk:
Fabio requested the feature to be able to display a pattern mask while editing. Here it is.

### DIFF
--- a/packages/input-mask/README.md
+++ b/packages/input-mask/README.md
@@ -9,7 +9,7 @@
 [![version](https://img.shields.io/npm/v/@solid-primitives/input-mask?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/input-mask)
 [![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-1.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
 
-Primitive that returns an event handler to mask the inputs of an text input element (`<input>`, `<textarea>`) when applied in `oninput` or `onchange`.
+Primitive that returns an event handler to mask the inputs of an text input element (`<input>`, `<textarea>`) when applied in `oninput` or `onchange` and set attributes to handle pattern display.
 
 ## Installation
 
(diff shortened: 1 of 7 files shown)


packages/input-mask/README.md
```diff
@@ -91,6 +91,34 @@ return (
...
+
+return <>
+  <style>{`
+    input { border: 0.1rem solid currentColor; padding: 0.5rem 0.75rem; }
+    label[data-mask-value] { position: absolute; border: 0.1rem solid transparent; padding: 0.5rem 0.75rem; }
+    label[data-mask-value]::before { content: attr(data-mask-value); color: transparent; }
+    label[data-mask-pattern]::after { content: attr(data-mask-pattern); opacity: 0.7; }
+  `}</style>
+  <label for="datefield">ISO Date:</label>
+  <label for="datefield"></label>
```
thetarnav, line 110: This label shouldn’t be duplicated, right?

packages/input-mask/README.md
```diff
@@ -91,6 +91,34 @@ return (
...
+
+## Showing the pattern during input
+
+The placeholder attribute is nice, but it cannot show the pattern during input. In order to rectify this shortcoming, this package has a `createMaskPattern` export:
+
+```tsx
+import { createInputMask, createMaskPattern } from "@solid-primitives/input-mask";
+
+return <>
+  <style>{`
```
thetarnav, line 103: I like including styles, but the online ones don’t have any syntax highlighting and look crammed together because of no formatting
Making a separate css code block could be clearer.

packages/input-mask/src/index.ts
```diff
@@ -119,3 +119,46 @@ export const createInputMask = <
   };
   return handler;
 };
+
+/**
+ * Adds a data-attribute with the remaining pattern to be used in ::after
+ * ```tsx
+ * <style>
+ * label[data-mask-value] {
```
thetarnav, line 127: Is this supposed to be empty?

thetarnav (approved):
Very nice :)

packages/input-mask/README.md
atk: Yes, it should. One is for the descriptive text, the other one for the pattern. It's described in the README.md.

packages/input-mask/README.md
atk: Good point.

packages/input-mask/README.md
atk: Maybe I should add a descriptive id to make its function more obvious, though.

packages/input-mask/src/index.ts
atk: Fixed this, too.

PR #458, opened 2023-06-13
Storage rewrite

atk:
The storage package was needlessly complex, large and led to frequent misunderstandings from developers who expected it to work like a store. I have learned from my mistakes and deprecated the old APIs with the exception of the cookieStorage. I also added a new smaller `makePersisted(signalOrStore)` primitive that should fulfil the general requirements and improved the cookieStorage.

### DIFF
--- a/configs/vitest.config.ts
+++ b/configs/vitest.config.ts
@@ -45,6 +45,12 @@ export default defineConfig(({ mode }) => {
     },
     resolve: {
       conditions: testSSR ? ["node"] : ["browser", "development"],
+      // fix issues with stores and node@20:
+      ...(!testSSR && { 
+        alias: {
+          "solid-js/store": "solid-js/store/dist/store.js"
+        },
+      }),
     },
   };
 });
--- a/.changeset/giant-steaks-double.md
+++ b/.changeset/giant-steaks-double.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/storage": major
+---
+
+new makePersisted primitive, deprecated old API
--- a/packages/storage/README.md
+++ b/packages/storage/README.md
@@ -46,6 +46,14 @@ type PersistedOptions<Type, StorageOptions> = {
 - values persisted in asynchronous storage APIs will not overwrite already changed signals or stores
 - setting a persisted signal to undefined or null will remove the item from the storage
 
+### Different storage APIs
+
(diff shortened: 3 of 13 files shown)


configs/vitest.config.ts
```diff
@@ -45,6 +45,12 @@ export default defineConfig(({ mode }) => {
     },
     resolve: {
       conditions: testSSR ? ["node"] : ["browser", "development"],
+      // fix issues with stores and node@20:
+      ...(!testSSR && { 
+        alias: {
+          "solid-js/store": "solid-js/store/dist/store.js"
```
thetarnav, line 51: Could you explain this? do we need the same thing for `"solid-js/web"` too? Also we have set the required node version to `18` for the repo. Maybe we should add a [`.nvmrc`](https://github.com/thetarnav/solid-devtools/blob/main/.nvmrc) file to enforce it?

atk: Node@20 partially messes up the resolver for vitest. I have not yet found the reasons, only two workarounds: make solid an external dependency or alias some files.

thetarnav (approved):
I like this change, it's going in the right direction with simplicity and composability.
Although I would split the primitive into multiple parts, instead of simply wrapping a store/signal tuple.
What if I have a custom state API that requires special reconciliation, or I'm using a store and want to reconcile manually with `produce`?
What if my state API does not provide a tuple?
What if I want to use it with `createResource`?
Like it would be nice to split the API into 'serialization', 'deserialization', 'getting initial value', 'updating storage on state change' and 'updating state on storage change'. This way the user would have more clarity into what the primitive will actually do, and how he can customize it, instead of just configuring the options object to some limited extend.

atk:
> What if I have a custom state API that requires special reconciliation, or I'm using a store and want to reconcile manually with produce? What if my state API does not provide a tuple?

Then you're welcome to write an adapter that projects your custom state API to something resembling either a signal or a store.
I want the simplest possible API for the default use cases, though I'm happy with splitting this like you suggested, as long as the main API remains the same. I'd prefer to get this out in any case. So I'd prefer to have this merged now and take some time to split this up afterwards.

> What if I want to use it with createResource?

Then you should use `createResource(fetcher, { storage: makePersistent(createSignal()) })`; anything else might make things unneccessary complex. But you're right, I should document that.

thetarnav:
> I'd prefer to get this out in any case. So I'd prefer to have this merged now and take some time to split this up afterwards.

yup go ahead, just wanted to get my thoughts out. The new API is definitely an improvement.

> I should document that.

💯😊

PR #459, opened 2023-06-13
update map/set constructor to match TypeScript's

AlexErrant:
There are two potential sources of type definitions:

1. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts
2. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.collection.d.ts

I'm going with `iterable` because it's more general.

### DIFF
--- a/packages/map/src/index.ts
+++ b/packages/map/src/index.ts
@@ -24,7 +24,7 @@ export class ReactiveMap<K, V> extends Map<K, V> {
   #keyTriggers = new TriggerCache<K | typeof $KEYS>();
   #valueTriggers = new TriggerCache<K>();
 
-  constructor(initial?: [K, V][]) {
+  constructor(initial?: Iterable<readonly [K, V]> | null) {
     super();
     if (initial) for (const v of initial) super.set(v[0], v[1]);
   }
@@ -127,7 +127,7 @@ export class ReactiveWeakMap<K extends object, V> extends WeakMap<K, V> {
   #keyTriggers = new TriggerCache<K>(WeakMap);
   #valueTriggers = new TriggerCache<K>(WeakMap);
 
-  constructor(initial?: [K, V][]) {
+  constructor(initial?: Iterable<readonly [K, V]> | null) {
     super();
     if (initial) for (const v of initial) super.set(v[0], v[1]);
   }
--- a/packages/set/src/index.ts
+++ b/packages/set/src/index.ts
@@ -18,7 +18,7 @@ const $KEYS = Symbol("track-keys");
 export class ReactiveSet<T> extends Set<T> {
   #triggers = new TriggerCache<T | typeof $KEYS>();
 
-  constructor(values?: readonly T[] | null) {
+  constructor(values?: Iterable<T> | null) {
     super();
     if (values) for (const v of values) super.add(v);
   }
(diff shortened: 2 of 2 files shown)


packages/map/src/index.ts
```diff
@@ -24,7 +24,7 @@ export class ReactiveMap<K, V> extends Map<K, V> {
   #keyTriggers = new TriggerCache<K | typeof $KEYS>();
   #valueTriggers = new TriggerCache<K>();
 
-  constructor(initial?: [K, V][]) {
+  constructor(initial?: Iterable<readonly [K, V]> | null) {
```
AlexErrant, line 27: c.f. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts#LL141C24-L141C59

packages/map/src/index.ts
```diff
@@ -127,7 +127,7 @@ export class ReactiveWeakMap<K extends object, V> extends WeakMap<K, V> {
   #keyTriggers = new TriggerCache<K>(WeakMap);
   #valueTriggers = new TriggerCache<K>(WeakMap);
 
-  constructor(initial?: [K, V][]) {
+  constructor(initial?: Iterable<readonly [K, V]> | null) {
```
AlexErrant, line 130: c.f. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts#LL147C40-L147C67

Note that TypeScript doesn't support undefined or null; I think this is a bug because [the spec](https://tc39.es/ecma262/#sec-weakmap-iterable) says:

> 4. If iterable is either undefined or null, return map.

[The TS PR](https://github.com/microsoft/TypeScript/pull/43396) making "Map constructor argument optional and nullable" didn't include `WeakMap`, which I believe is an oversight. We'll see if/when I get enough nerve to submit a PR TypeScript 😅

packages/set/src/index.ts
```diff
@@ -18,7 +18,7 @@ const $KEYS = Symbol("track-keys");
 export class ReactiveSet<T> extends Set<T> {
   #triggers = new TriggerCache<T | typeof $KEYS>();
 
-  constructor(values?: readonly T[] | null) {
+  constructor(values?: Iterable<T> | null) {
```
AlexErrant, line 21: c.f. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts#LL189C21-L189C42

I'm not sure why this is `readonly`; looking at the gitblame for the above line in the TS repo seems to indicate it was never `readonly`.

packages/set/src/index.ts
```diff
@@ -94,7 +94,7 @@ export class ReactiveSet<T> extends Set<T> {
 export class ReactiveWeakSet<T extends object> extends WeakSet<T> {
   #triggers = new TriggerCache<T>(WeakMap);
 
-  constructor(values?: readonly T[] | null) {
+  constructor(values?: Iterable<T> | null) {
```
AlexErrant, line 97: https://github.com/microsoft/TypeScript/blob/main/src/lib/es2015.iterable.d.ts#LL195C47-L195C60

The above comments also apply here - in particular the `readonly` comment and the fact that TS doesn't have `WeakSetConstructor` support null or undefined, which (again) doesn't conform to [the spec](https://tc39.es/ecma262/#sec-weakset-constructor):

> 4. If iterable is either undefined or null, return set.

packages/set/src/index.ts
thetarnav: `readonly` was there only to allow passing `readonly` arrays, instead of forcing you to copy to satisfy ts.
with iterable it's not an issue.

packages/map/src/index.ts
thetarnav: since we are doing a `if (initial)` check, we could allow any falsy type.
But we should probably limit to nullable if we decide to change the check

thetarnav:
Thank you :)

PR #462, opened 2023-06-18
Added visibility primitive

Tommypop2:
Hi,

This package creates a reactive wrapper around the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API). It returns an accessor containing a boolean - storing whether or not the page is currently visible.


Tommypop2:
Seems like this already exists

PR #463, opened 2023-06-19
fix link to jsx-tokenizer example

bigmistqke:
fixed link to jsx-tokenizer example (was `https://primitives.solidjs.community/playground/` before)


thetarnav:
thanks ✌

Issue #399, opened 2023-04-13
JSX Tokenizer: add tokens which can be parsed using `children`


thetarnav:
### Describe The Problem To Be Solved

Currently, the component returned by `createToken` will return a function element. This way we can display a friendly warning in the console if something tries to render the token, instead of an error from trying to insert a non-`JSX.Element` type.
The downside of that approach is that the token elements cannot be passed to a context provider, transition-group, or anything that tries to resolve the children using the `children` helper, even if it doesn't necessarily render anything yet.
So we lose a bit of functionality, and composition value, for a nicer misuse warning.

### Suggest A Solution

I see this being implemented in two ways.
Either the `createToken` component will be returning an object, instead of a function if the `fallbackFn` param is not provided.
Or there could be `createToken` and `createRenderableToken` as separate functions.

bigmistqke:
+1 the whole `not-resolving-tokens` in jsx-tokenizer is a bit finnicky and brings mental overhead that makes me less want to use it (or advice others to use it). Especially since solid itself perfectly allows for non-DOM to be returned (only typescript is unhappy about that). Typecasting objects to `JSX.Element`, in combination with the typing that `jsx-tokenizer` provides would be typesafe enough imo.

bigmistqke:
little experiment with simply casting: 
https://playground.solidjs.com/anonymous/5110caf3-7129-411b-9da3-7ff423f6d296

```tsx
const token = (args[0] ? args[0](props) : props) as unknown as TokenElement<T>;
token[$TOKENIZER] = symbol;
return token as unknown as JSX.Element;
```
then we can only accept objects as return-values, with the added benefit the abstraction looks a bit cleaner:

```tsx
const Token = createToken(() => ({ id: 'string' }));

const Parent = (props: {children: JSX.Element)) => {
  const token = props.children;
  token.id // 'string'
  // instead of
  token.data.id // 'string'
})
```

an extra bonus to requiring an object as returntype is that it will skip rendering if it accidentally gets into the dom-tree, while if we allow primitive values these would be rendered as-is.

Context becomes a lot easier too:
```tsx
const childs = <Context.Provider value={...}>{props.children}</Context.Provider>
```
just works.

This is definitely the way forward.

thetarnav:
It's more or less what I want to do
but I without adding a property to the token (`token[$TOKENIZER] = symbol`)
But instead use a `WeakMap` for identifying passed objects.

bigmistqke:
[...]

PR #464, opened 2023-06-19
withContext and withMultiContexts

bigmistqke:
Using context with `jsx-tokenizer` can be [quite painful](https://github.com/solidjs-community/solid-primitives/tree/main/packages/jsx-tokenizer#usage-with-context-api). To remedy this I have been using a utility-function `withContext` in my side-project [solid-canvas](https://github.com/bigmistqke/solid-canvas) [(see)](https://github.com/bigmistqke/solid-canvas/blob/main/src/utils/withContext.tsx). 

I modified this function into `withContext` and `withMultiContexts` to be more in-line with other `solid-primitives` like `MultiProvider`.

# API

## withContext

```ts
const NumberContext = createContext<number>
const children = withContext(
   () => props.children,
   NumberContext,
   1
)
```

## withMultiContexts

the API mirrors `MultiProvider` (for now I have kept it simple and only `[Context<T>, T]` is allowed), but the implementation is a bit different, since with `jsx-tokenizer` we have to be careful not to resolve the tokens [(see)](https://playground.solidjs.com/anonymous/0970d73c-e743-453f-abf0-af626e445106) this would cause a warning and prevent the token to be passed on.

```ts
const NumberContext = createContext<number>
const StringContext = createContext<string>
const children = withContext(
   () => props.children,
   [
     [NumberContext, 1],
     [StringContext, "string"]
   ]
)
```

Its usage is not limited to `jsx-tokenizer`, it could be used with 'regular' solid as well.

Demos: 
- [jsx-tokenizer](https://playground.solidjs.com/anonymous/4d018284-60f7-4c4a-8fb4-541e13c9dd70)
- [regular solid](https://playground.solidjs.com/anonymous/6f6ffd28-2151-44d7-af64-be389f1117e5)

### DIFF
--- a/packages/context/src/index.ts
+++ b/packages/context/src/index.ts
@@ -182,7 +182,7 @@ Type validation of the `values` array thanks to the amazing @otonashixav (https:
  * const NumberContext = createContext<number>
  * const StringContext = createContext<string>
  * const children = withContext(
- *    () => children(props.children),
+ *    () => props.children,
  *    [
  *      [NumberContext, 1],
  *      [StringContext, "string"]
--- a/packages/context/src/index.ts
+++ b/packages/context/src/index.ts
@@ -1,12 +1,11 @@
 import {
-  createContext,
-  createComponent,
-  useContext,
-  JSX,
+  Accessor,
   Context,
   FlowComponent,
-  Accessor,
-  untrack,
+  JSX,
+  createComponent,
+  createContext,
+  useContext,
 } from "solid-js";
 import type { ContextProviderComponent } from "solid-js/types/reactive/signal";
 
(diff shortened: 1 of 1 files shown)


bigmistqke:
Possibly less essential for `jsx-tokenizer` if we would change its implementation as proposed [here](https://github.com/solidjs-community/solid-primitives/issues/399) instead, but still could be handy I think.

packages/context/src/index.ts
```diff
@@ -119,3 +127,94 @@ export function MultiProvider<T extends readonly [unknown?, ...unknown[]]>(props
...
+
+export function withMultiContexts<T extends readonly [unknown?, ...unknown[]]>(
+  children: Accessor<JSX.Element | JSX.Element[]>,
+  values: {
+    [K in keyof T]: readonly [Context<T[K]>, [T[K]][T extends unknown ? 0 : never]];
+  },
+) {
+  let result: JSX.Element | JSX.Element[];
+
+  const fn = (index: number) => {
```
atk, line 202: Wouldn't it be simpler to use `values.reduce((children, [context, value]), children) => ...)`

bigmistqke: Something like
```ts
(values as [Context<any>, any]).reduce((acc, [context, value], index) => {
    return () =>
      context.Provider({
        value,
        children: () => {
          if (index === 0) result = acc();
          else acc();
        },
      });
  }, children)()
 ```
 ?
 
[playground](https://playground.solidjs.com/anonymous/c4747502-bcdb-4c5e-8c83-6fffad0b2856) 

I don't use `reduce` myself that much, so this code is a bit less readable to me personally, but that's mb more of a skill issue.

File: packages/context/src/index.ts [hunk 2]
```diff
@@ -119,3 +127,94 @@ export function MultiProvider<T extends readonly [unknown?, ...unknown[]]>(props
...
+          result = children();
+        }
+        return "";
+      }) as any as JSX.Element,
+    });
+  };
+
+  fn(0);
+
+  return () => result;
```
bigmistqke, line 219: The accessor is redundant. We can also return `result` directly.

packages/context/src/index.ts
```diff
@@ -119,3 +127,94 @@ export function MultiProvider<T extends readonly [unknown?, ...unknown[]]>(props
...
+
+  context.Provider({
+    value,
+    children: (() => {
+      result = children();
+      return "";
+    }) as any as JSX.Element,
+  });
+
+  return () => result;
```
bigmistqke, line 165: The accessor is redundant. We can also return result directly.

deminearchiver:
How would this function handle a custom context provider (in case the context is private)?

```tsx
const MyContext = createContext</*...*/>();
export const MyContextProvider: ParentComponent = (props) => { /* ... */ }
```

I think you should add support for generic `Component<T>` instead of `Context<T>`

Issue #466, opened 2023-06-28
Extend `presence` to reach feature parity with `transition-group`


nonphoto:
### Describe The Problem To Be Solved

`presence` is super helpful for transitioning between components without unmounting them. However, it is missing some features that are present in `transition-group` despite being useful in the same scenarios, mainly:

1. Modes: `parallel` and `in-out`
2. Callbacks for animation completion (instead of duration-based timeout)
3. List transitions

### Suggest A Solution

Update `createPresence` to match the API of `createSwitchTransition`. Add an additional hook, `createPresenceList` that matches the API of `createListTransition`. I don't anticipate any major difficulty in this because the implementation should mirror that of `transition-group` but I haven't looked into the source of either library too deeply yet.

I would be open to creating a PR for this as I will be building this functionality anyway as part of a private project.

PR #474, opened 2023-07-18
Refactor transition-group to use per-element state machines

nonphoto:
WIP

This PR is an effort to resolve Issue #466 by taking the best of both worlds from `transition-group` and `presence`

As mentioned in #466 `presence` primarily works by deriving element appearance from transition state. However, it is missing:
1. Modes: parallel and in-out
2. Callbacks for animation completion (instead of duration-based timeout)
3. List transitions

In opposition, `transition-group` works by attaching callbacks to elements that have already rendered. It is lacking:
1. Access to transition state from within the transitioning item's context
2. The ability to run different animations per transitioning element.

This PR resolves these differences by introducing the `TransitionItemContext` type, an instance of which holds the transition state for each element in a list. `createListTransition` now returns an array of tuples: `[T, TransitionItemContext]`. The user can use `TransitionItemContext` to register transition events or track the transition state independently for each element.

Currently only `createListTransition` has been modified. If this PR goes over well I will extend the functionality to `createTransition` as well.

---

This also has the added benefit of removing from jank from the playground's transitions.

Old:

https://github.com/solidjs-community/solid-primitives/assets/3441123/9fa290a1-4ec0-4afe-ab2d-321ebbd19bae

New:

https://github.com/solidjs-community/solid-primitives/assets/3441123/d3081081-7c7d-4c2e-a2eb-c995d7a3a98a

### DIFF
--- a/packages/transition-group/dev/list-page.tsx
+++ b/packages/transition-group/dev/list-page.tsx
@@ -1,5 +1,6 @@
 import {
   createEffect,
+  createMemo,
   createRenderEffect,
   createResource,
   mapArray,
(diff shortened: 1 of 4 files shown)


nonphoto:
Currently experiencing issues with animation sequencing: callbacks work correctly but animations are not playing in the correct sequence or measuring at the right moment for all options.

nonphoto:
I'm no longer working on this. Will re-open if there's still interest.

PR #479, opened 2023-07-31
input mask: optionals

atk:


packages/input-mask/test/index.test.tsx
```diff
@@ -13,6 +13,7 @@ const dispatchInputEvent = (node: HTMLElement) =>
 describe("createInputMask", () => {
   it("adds placeholder (e.g. to an iso date)", async () => {
     const { container } = render(() => <input onInput={createInputMask("9999-99-99")} />);
+    console.warn(container.innerHTML);
```
thetarnav, line 16: leftover?

configs/vitest.config.ts
```diff
@@ -22,6 +22,7 @@ export default defineConfig(({ mode }) => {
       isolate: !testSSR,
       passWithNoTests: true,
       environment: testSSR ? "node" : "jsdom",
+      deps: { registerNodeLoader: false },
```
thetarnav, line 25: What's that for?

thetarnav (approved):

packages/input-mask/test/index.test.tsx
atk: Yes. Removed.

configs/vitest.config.ts
atk: Unfortunately, the node loader is broken in node@20 - and it is still enabled by default in the solid plugin, because it was supposed to fix the problems it now causes. Fortunately, they are solved in vite's internal loader now so that it's fine to disable it on older node versions as well.

Issue #480, opened 2023-08-01
[storage] `cookieStorage` doesn't work with Solid-Start


yume-chan:
### Describe the bug

When using `cookieStorage` with `makePersisted`, the server console always prints "It seems you attempt to use cookieStorage on the server without having solid-start installed".

Workaround: Copy https://github.com/solidjs-community/solid-primitives/blob/d74956b93850672afcb519d917ca9b08e9457362/packages/storage/src/cookies.ts into my project and convert `require("solid-start/server")` to a normal top level `import`.

### Minimal Reproduction Link

https://stackblitz.com/edit/github-n4ho5j-u5pz4k?file=src%2Froutes%2Findex.tsx

atk:
The code you've shown is executed both on the server and the client - and it seems that the server import for useRequest fails. If you want it to only store cookies on the client in solid-start, you need to `import isServer from "solid-js/web"` and use

```ts
const [value, setValue] = isServer
  ? createSignal("initial")
  : makePersisted(
      createSignal("initial"),
      { name: "value", storage: cookieStorage, storageOptions: { path: "/", maxAge: 60 * 60 * 24 * 7 } }
    );
```

yume-chan:
> If you want it to only store cookies on the client in solid-start,

No, I want to read the value on SSR, that's why I used `cookieStorage`, but it doesn't work.

atk:
Understood.

atk:
I think I have an idea. It reduces the convenience a bit, but on the other hand makes it far more flexible: I'll accept a requestAccessor (that either returns Request or, like useRequest, { request: Request }) and a cookieSetter as optional options.

yume-chan:
https://github.com/solidjs/solid-start/blob/e83271fef5469b4e4e986566f8d07e84388558c9/packages/start/server/ServerContext.tsx#L7-L13

```js
export const useRequest = () => {
  return useContext(ServerContext)!;
};


export const useServerContext = () => {
  throw new Error("useServerContext is deprecated. Use useRequest instead.");
};
```

I think `useServerContext` is the old name and no longer be used.

atk:
Thanks, the docs are still on the old name. In any case, even though it should work out of the box in other cases than stackblitz, you can use the getRequest option to give useRequest to makePersisted manually and it will no longer rely on the attempted optional import that breaks in stackblitz - after [PR ](https://github.com/solidjs-community/solid-primitives/pull/482) is merged and released.

[...]

PR #482, opened 2023-08-01
storage: allow to finetune server behavior

atk:
Fixes https://github.com/solidjs-community/solid-primitives/issues/480

### DIFF
--- a/packages/storage/src/cookies.ts
+++ b/packages/storage/src/cookies.ts
@@ -2,6 +2,7 @@ import { isServer } from "solid-js/web";
 import { StorageProps, StorageWithOptions, StorageSignalProps } from "./types";
 import { addClearMethod } from "./tools";
 import { createStorage, createStorageSignal } from "./storage";
+import type { PageEvent } from "solid-start";
 
 export type CookieOptions = {
   domain?: string;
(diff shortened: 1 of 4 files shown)


packages/storage/src/cookies.ts
```diff
@@ -58,20 +61,28 @@ try {
...
  * ```
  * Also, you can use its _read and _write properties to change reading and writing
  */
 export const cookieStorage: StorageWithOptions<CookieOptions> = addClearMethod({
-  _read: isServer ? () => useRequest().request.headers.get("Cookie") : () => document.cookie,
+  _read: isServer
+    ? (options?: CookieOptions) => {
+      const eventOrRequest = options?.getRequest?.() || useRequest?.();
+      const request = eventOrRequest && ('request' in eventOrRequest ? eventOrRequest.request : eventOrRequest);
+      return request?.headers.get("Cookie") || "";
```
yume-chan, line 75: For more flexibility, the new option should return the whole Cookie string, because other framework's request object might not use `request.headers.get("Cookie")`

atk: Which solid-based framework gives you the cookies as string and doesn't provide an easy access to the Request object?

yume-chan: > Which solid-based framework gives you the cookies as string and doesn't provide an easy access to the Request object? 

I don't have a concrete example, it's only a possibility.

atk: Thank you for considering that. I'm hesitant to add stuff that's very probably never going to be used.

thetarnav:
so I tried looking at it and didn't understand a single thing :rofl:
have you tested that with solid-start? if it works then a job done I guess
the api didn't change too much, although have you considered giving the user the ability to override _write and _read entirely by own implementation?

atk:
It is already tested both locally with solid-start and also with stackblitz (then you have to provide the getRequest option because the automatic loading does not work there, but that's documented).

PR #486, opened 2023-08-05
State machine

thetarnav:
A primitive for creating a reactive state machine. For expressing possible states and transitions, and bounding reactive computations to the lifecycle of those states.

This takes a bit different and more minimalistic approach to state machines. Maybe you cannot even call it a state machine. But I think it fits well the already very declarative capabilities of solid's reactivity, and utilizes the lifecycle of computations to scope state relevant logic.

I've been using this pattern when defining app logic that could be described using exclusive states, and relying on solid's reactivity one could very easily achieve that. But the biggest issue were types, defining unions of possible states can be very annoying, and usually will be done inconsistently because of that.
This primitive mainly helps with types, while providing slight runtime convenience utilities.

more details and examples in the [README](https://github.com/solidjs-community/solid-primitives/tree/state-machine/packages/state-machine#readme)

The story around nested machines, events, and more is still unfinished, but as a simple primitive it's already very useful.

### DIFF
--- a/packages/state-machine/src/index.ts
+++ b/packages/state-machine/src/index.ts
@@ -1,4 +1,4 @@
-import { createMemo, createSignal, type Accessor } from "solid-js";
+import { createMemo, createSignal, type Accessor, untrack } from "solid-js";
 
 /**
  * Used for restricting the user type input in {@link createMachine} generic.
@@ -176,18 +176,21 @@ export function createMachine<T extends StatesBase<keyof T>>(
     EQUALS_OPTIONS,
   );
 
-  const to = (type: keyof T, input: any) => {
+  const to: any = (type: keyof T, input: any) => {
     setPayload({ type, input });
   };
 
   for (const key of Object.keys(states)) {
-    // @ts-expect-error
     to[key as any] = (input: any) => to(key, input);
   }
 
   const memo = createMemo(() => {
     const { type, input } = payload();
-    return { type, value: states[type](input, to as any), to };
+    return {
+      type,
+      value: untrack(() => states[type](input, to)),
+      to,
+    };
   }) as any;
 
   Object.defineProperties(memo, {
--- a/packages/state-machine/src/index.ts
+++ b/packages/state-machine/src/index.ts
@@ -28,8 +28,8 @@ export type StateValue<T extends { value?: any }> = T extends { value: infer Val
  * @example
  * ```ts
  * const states: MachineStates<{
- *   idle: { input: void; value: "foo"; to: ["loading"] };
- *   loading: { input: void; value: "bar"; to: ["idle"] };
+ *   idle: { value: "foo"; to: ["loading"] };
+ *   loading: { value: "bar"; to: ["idle"] };
  * }> = {
  *   idle: () => "foo",
  *   loading(input, next) {
@@ -115,6 +115,8 @@ export type MachineNext<T extends StatesBase<keyof T>, TKey extends keyof T> = {
   ...args: T[K] extends { input: infer Input } ? [to: K, input: Input] : [to: K, input?: undefined]
 ) => void);
 
+const EQUALS_OPTIONS = { equals: (a: { type: any }, b: { type: any }) => a === b };
+
 /**
  * Creates a reactive state machine.
  *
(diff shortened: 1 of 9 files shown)


packages/state-machine/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+ *   state.value // "bar"
+ * }
+ * ```
+ */
+export function createMachine<T extends StatesBase<keyof T>>(
+  options: MachineOptions<T>,
+): Accessor<MachineState<T, keyof T>> & MachineState<T, keyof T> {
+  const { states, initial } = options;
+
+  const [payload, setPayload] = createSignal(
```
atk, line 170: Instead of using payload and a memo to access type and value, why not use two signals for type and memo and batch the updates?

packages/state-machine/test/index.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        states: {
+          idle: i => i,
+          loading: i => i,
+        },
+      });
+
+      expect(state.type).toBe("idle");
+      expect(state.value).toBe("foo");
+
+      if (state.type === "idle") {
```
atk, line 69: This if is superfluous. The test should have already thrown otherwise.

packages/state-machine/test/index.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        state.to.loading(1);
+      }
+
+      expect(state.type).toBe("loading");
+      expect(state.value).toBe(1);
+
+      if (state.type === "loading") {
```
atk, line 76: Same here.

atk (approved):

packages/state-machine/test/index.test.ts
thetarnav: The if is for types. as you can see in test above that it won’t let you switch to different state before checking if that’s possible. So you have to narrow down the current state.

packages/state-machine/src/index.ts
thetarnav: I don’t see how that would be beneficial. you use more signals to have more fine grain changes, here it’s always meant to be together. a single point of change, even if the value is an object.

packages/state-machine/test/index.test.ts
atk: Ah, too bad that typescript can't infer the current state from previous statements.

thetarnav: yeah, and I'm wondering if the getters on the signal should be removed too, as type narrowing can be misleading there. With a function it's more clear that you have to call it again to get the new state.

image

But then typechecking on `<Show>`, `<Switch>`, etc. becomes annoying

Like with getters you can do this
```ts
<Show when={state.type === "Reading" && state}>
  {narrowed_state => {
    narrowed_state // state for "Reading"
  }}
</Show>
```

With function it's more like
```ts
<Show when={(() => {
  const v = state()
  return v.type === "Reading" && v
})()}>
  {narrowed_state => {
    narrowed_state // state for "Reading"
  }}
</Show>
```

packages/state-machine/src/index.ts
atk: Storing those values in separate signals saves you the cost of creating objects again and again. Using batch will guarantee the changes will stay in sync.

packages/state-machine/test/index.test.ts
mosheduminer: What about the opposite - just getters, no memo/signal to call at all?

thetarnav: The signal form is more idiomatic, it makes it obvious what changes, because in the end it is just a single signal. The getters are just an additional convenience, they are not separate signals, and they shouldn't be by default.
It also makes the usage with primitives expecting a function source easier: (`on`, `createSelector`, `createResource`, etc)
```ts
const isNodeDragged = createSelector(
  state,
  (node: Node, v) => v.type === "Dragging" && v.value.node === node,
)
// vs
const isNodeDragged = createSelector(
  () => ({ ...state }),
  (node: Node, v) => v.type === "Dragging" && v.value.node === node,
)
```

packages/state-machine/src/index.ts
thetarnav: Adding another signal and a batch function would bring it's own complexity, allocations, and issues with managing triggering updates from two separate signals. *(I only want to trigger the memo to rerun if the state changed, not if it was called again with different input)*
And the memo is necessary to achieve the intended behavior.
But I managed to remove one object creation without it.

PR #489, opened 2023-08-10
Disable test isolation

thetarnav:
This tries to disable the vitest's `isolate` option for client tests. In ssr it's already disabled which dramatically increase the test performance, as the environment doesn't have to be reset and recreated for every test suite.

This requires the tests to not cause any side effects—which in tests shouldn't happen anyway—by cleaning every change to the env manually, and don't rely on timeouts when testing, as the timing of those can change depending on the load—fake timers are way faster and more reliable anyway.

~~It already pretty much works, but some packages like `resource`, `filesystem` and `fetch` are a bit flaky, and sometimes fail. Not sure why yet, I'm not too familiar in what's happening there.~~ *I guess it works very well now 🤷‍♂️*

It reduces the time spend testing on ci from ~1.25 min to ~15 ms 🔥

https://github.com/solidjs-community/solid-primitives/assets/24491503/6e285143-3739-42fa-a275-bbe3048198a6


thetarnav:
https://github.com/solidjs-community/solid-primitives/assets/24491503/6e285143-3739-42fa-a275-bbe3048198a6

PR #498, opened 2023-08-19
Update `start` README with Examples Section for `createUserTheme`

zboyles:
Added an Examples section to the `README.md` file that includes code snippets from `root.tsx`, `root.css` and `Counter.css`.

### DIFF
--- a/packages/start/README.md
+++ b/packages/start/README.md
@@ -95,9 +95,11 @@ import {
 import Counter from "./components/Counter";
 import "./root.css";
 import { createUserTheme } from '~/primitives/start';
+
 export default function Root() {
-  const [theme, setTheme] = createUserTheme("dark-light-theme", "dark");
+  const [theme, setTheme] = createUserTheme("dark-light-theme", { defaultValue: "dark" });
   const toggleTheme = () => setTheme(currentTheme => currentTheme === "dark" ? "light" : "dark");
+
   return (
     // Set data-theme when using libraries like Daisyui:
     <Html lang="en" data-theme={theme()}>
@@ -116,9 +118,7 @@ export default function Root() {
               "light-variant": theme() === "light",
             }}
           />
-          <button
-            onClick={toggleTheme}
-          >
+          <button onClick={toggleTheme}>
             <span>Toggle Theme</span>
           </button>
         </div>
@@ -149,7 +149,6 @@ body.dark {
   background-color: black;
   color: aliceblue;
 }
-
 body.light {
   background-color: aliceblue;
   color: black;
@@ -158,11 +157,9 @@ body.light {
 .dark span {
   color: #b1d4ff;
 }
-
 .light span {
   color: #003677;
 }
-
 ```
 
 ### `Counter.css`
--- a/packages/start/README.md
+++ b/packages/start/README.md
@@ -79,7 +79,6 @@ This is the main entry file for SolidStart demonstrating usage of the `createUse
 - Provides additional examples on using variant classes to style components.
 
 ```tsx
-// @refresh reload
 import { Suspense } from "solid-js";
 import {
   A,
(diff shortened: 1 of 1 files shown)


packages/start/README.md
```diff
@@ -66,6 +66,122 @@ const [theme, setTheme] = createUserTheme("cookieName", {
...
+
+This is the main entry file for SolidStart demonstrating usage of the `createUserTheme` function.
+- Initializes the theme with the `createUserTheme` function.
+- Toggles between the "dark" and "light" themes using a button.
+- Sets the `data-theme` attribute adding compatibility between `createUserTheme` and libraries like [DaisyUI](https://daisyui.com/).
+- Includes a clean implementation of theming through a `Body` class.
+- Provides additional examples on using variant classes to style components.
+
+```tsx
+// @refresh reload
```
thetarnav, line 82: solid refresh pragma doesn’t add anything to the example

packages/start/README.md
```diff
@@ -66,6 +66,122 @@ const [theme, setTheme] = createUserTheme("cookieName", {
...
+    <Html lang="en" data-theme={theme()}>
+      <Head>
+        <Title>SolidStart - Bare</Title>
+        <Meta charset="utf-8" />
+        <Meta name="viewport" content="width=device-width, initial-scale=1" />
+      </Head>
+      <Body
+        classList={{
+          // Apply theme as css class to body:
+          [theme()]: true,
```
thetarnav, line 113: This could be simplified to
```tsx
class={theme()}
```

packages/start/README.md
```diff
@@ -66,6 +66,122 @@ const [theme, setTheme] = createUserTheme("cookieName", {
...
+        }}
+      >
+        <div>
+          <Counter
+            classList={{
+              // Within components conditionally apply css classes:
+              "dark-variant": theme() !== "light",
+              "light-variant": theme() === "light",
```
thetarnav, line 121: again, wouldn’t you want to switch it like this?
```tsx
class={theme() === "light" ? "light-variant" : "dark-variant"}
```

thetarnav (commented):
Thank you for the example, it would be a great thing to add.

packages/start/README.md
zboyles: I agree although I intentionally wrote it this way because of all the Solid code I see where people begin stacking into class like so:
```tsx
<Component class={`${theme()} text-style other-style`} />
```

This doc provided a good opportunity to show `classList` and doubly so using a computed property. With that said, I think you're right and `Body` should use `class` but I'd like to keep `classList` as a commented out alternative unless you're against it.

packages/start/README.md
zboyles: I could use both methods and comment out `classList` again.

zboyles: Looking at this again, I don't think you'd use the theme at the component level other than special situations and even then you would most likely apply theme dependent variants through `classList`.

packages/start/README.md
```diff
@@ -66,6 +66,117 @@ const [theme, setTheme] = createUserTheme("cookieName", {
...
+  return (
+    // Set data-theme when using libraries like Daisyui:
+    <Html lang="en" data-theme={theme()}>
+      <Head>
+        <Title>SolidStart - Bare</Title>
+        <Meta charset="utf-8" />
+        <Meta name="viewport" content="width=device-width, initial-scale=1" />
+      </Head>
+      // Apply theme as a class to the Body element:
+      <Body class={theme()}>
```
zboyles, line 110: Considering the demonstration of `classList` at the component level I decided to follow your recommendation and keep this simple.

packages/start/README.md
```diff
@@ -66,6 +66,117 @@ const [theme, setTheme] = createUserTheme("cookieName", {
...
+  Meta,
+  Routes,
+  Scripts,
+  Title,
+} from "solid-start";
+import Counter from "./components/Counter";
+import "./root.css";
+import { createUserTheme } from '~/primitives/start';
+export default function Root() {
+  const [theme, setTheme] = createUserTheme("dark-light-theme", "dark");
```
Tommypop2, line 99: Should be `createUserTheme("dark-light-theme", { defaultValue: "dark" });` I think

PR #499, opened 2023-08-19
Fix `start` `useRequest` import

zboyles:
Fixed the import statement directed at the `jsx` file:
`import { useRequest } from "solid-start/server/ServerContext.jsx";`

Resolution:
`import { useRequest } from "solid-start/server";`

### DIFF
--- a/packages/start/src/index.ts
+++ b/packages/start/src/index.ts
@@ -1,7 +1,7 @@
 import { createSignal, createEffect, Signal } from "solid-js";
 import { isServer } from "solid-js/web";
 import { parseCookie } from "solid-start";
-import { useRequest } from "solid-start/server/ServerContext.jsx";
+import { useRequest } from "solid-start/server";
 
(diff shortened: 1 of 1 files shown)


BierDav (approved):
LGMT

thetarnav:
As I remember `.jsx` was because of `NodeNext` node resolution.
SolidStart doesn't include extensions when reexporting in `server.ts`, which breaks the resolution.
I would like to "fix" the import as well, but solid-start needs fixing first probably.
You can see it breaking the build.
Also we should probably move to `@solidjs/start` package.

BierDav:
Also we should probably move to @solidjs/start package.

Then let's get this merged first https://github.com/solidjs-community/solid-primitives/pull/585 and then I will do a followup pr that fixes the import.

thetarnav:
Fixed in https://github.com/solidjs-community/solid-primitives/pull/660

PR #507, opened 2023-08-28
Fix ReactiveMap iterators tracking

thetarnav:
- `.forEach()` wasn't tracking changes to values, only keys

- `.keys()`, `.entries()` and `.values()` was tracking any changes to keys, even if the key was not seen
	Also some were iterating the map twice

- Same for `ReactiveSet`

### DIFF
--- a/.changeset/angry-planes-shout.md
+++ b/.changeset/angry-planes-shout.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/set": patch
+---
+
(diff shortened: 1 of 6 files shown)


thetarnav:
@NorbertBodziony np. does this solve your issue entirely? would you mind checking it locally before merging and publishing this?

NorbertBodziony:
@thetarnav yeah so far no issues. I will let you know if i find anything else after you release fix.

PR #518, opened 2023-09-24
Add selectors

ryoid:
Using [`createSelector`](https://www.solidjs.com/docs/latest/api#createselector) is good for a single values. I found that there are a lot of use cases for array selectors. I have been using this on a few projects and I think it would be a good addition to the library.

```tsx
const list: string[] = ["apple", "pear", "orange"]
const [selectedItems] = createSignal<string[]>(["apple"])
const isSelected = createArraySelector(selectedItems)

<For each={list}>
  {(item) => <li classList={{ active: isSelected(item) }}>{item}</li>}
</For>
```

Extending with a comparator function (instead of `b.includes(a)`) to support objects could be possible, but I decided to keep it simple. At that point it might be better to just create a custom selector directly.

### DIFF
--- a/packages/selectors/README.md
+++ b/packages/selectors/README.md
@@ -1,0 +1,44 @@
+<p>
+  <img width="100%" src="https://assets.solidjs.com/banner?type=Primitives&background=tiles&project=selectors" alt="Solid Primitives selectors">
+</p>
+
+# @solid-primitives/selectors
+
+[![turborepo](https://img.shields.io/badge/built%20with-turborepo-cc00ff.svg?style=for-the-badge&logo=turborepo)](https://turborepo.org/)
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/selectors?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/selectors)
+[![version](https://img.shields.io/npm/v/@solid-primitives/selectors?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/selectors)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
+
+A sample primitive that is made up for templating with the following options:
+
+`createArraySelcetor` - Provides a getter and setter for the primitive.
+
+## Installation
+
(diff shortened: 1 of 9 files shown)


packages/selectors/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+ * <For each={list}>
+ *   {(item) => 
+ *    <li classList={{ active: isSelected(item) }}>
+ *      {item}
+ *    </li>}
+ * </For>
+ */
+export function createArraySelector<T>(
+  source: () => Array<T>,
+  options?: { name?: string },
```
thetarnav, line 25: Better to reuse types from solid for the options (and any other type if possible), to show where it is coming from, and remove the need for updating the types when options in solid change.

packages/selectors/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@solid-primitives/selectors",
```
thetarnav, line 2: prefer `selector` singular to match the rest of the packages

packages/selectors/README.md
```diff
@@ -0,0 +1,44 @@
...
+```tsx
+const list: string[] = ["apple", "pear", "orange"]
+const [selectedItems] = createSignal<string[]>(["apple"])
+const isSelected = createArraySelector(selectedItems)
+
+<For each={list}>
+  {(item) => <li classList={{ active: isSelected(item) }}>{item}</li>}
+</For>
+```
+
```
thetarnav, line 37: Worth adding that it doesn't work when mutating the array:
https://playground.solidjs.com/anonymous/bfeda055-2f2c-465e-b438-ea9d671a4eb8

thetarnav, line 37: Worth showing that the implementation is basically an alias to `createSelector` with `Array.includes`
```ts
const createArraySelector = source => createSelector(source, (key, list) => list.includes(key))
```
I don't want anyone to get an impression that it's not possible to do this with `createSelector`

thetarnav (changes requested):
It's an interesting idea to have a package dedicated to the selector primitive. There might be other helpers this could include.
But right now I wouldn't advice anyone to use it with only `createArraySelector`, especially when it doesn't provide too much value.
Also I'm wondering if instead of exporting a `createSelector` wrapper, it would be better to just export the `(key, list) => list.includes(key)` part. Then one could compose it with the selector.
```ts
import { arrayIncludes } from "@solid-primitives/selector"

const isItemSelected = createSelector(selectedItems, arrayIncludes)
```

davedbase:

davedbase:

PR #519, opened 2023-09-26
Move Solid's `createMutable` to new `mutable` package

thetarnav:
This copies [the current implementation of `createMutable`](https://github.com/solidjs/solid/blob/063db14c4e69f411ac2c09f5044e2d262fa667b9/packages/solid/store/src/mutable.ts) in solid 1.7.12 one-to-one, as the primitive is planned to be removed in core in future releases.
As the API is still useful to many people, it will be continued to be developed and maintained here.
The v1 of this package should remain the same as it was in solid, to avoid breaking existing apps, but later it can be extended past the current implementation, as it won't be "held back" by the tight compatibility with `createStore`.
`createMutable` in solid was some internal logic from `createStore` implementation, so using both `createMutable` and `createStore` in your app will mean larger bundle size after the move.


configs/vitest.config.ts
```diff
@@ -46,12 +46,6 @@ export default defineConfig(({ mode }) => {
     },
     resolve: {
       conditions: testSSR ? ["node"] : ["browser", "development"],
-      // fix issues with stores and node@20:
-      ...(!testSSR && {
-        alias: {
-          "solid-js/store": "solid-js/store/dist/store.js",
```
thetarnav, line 52: @atk I needed to remove it so that both `solid-js/store` and `solid-js/web` resolve to the dev runtime version. Instead of the `store` resolving to prod, and `web` to dev, causing `isDev` to be `true` while the dev features were missing.

PR #522, opened 2023-10-08
Update index.ts

muzidudu:
Something seems wrong here

### DIFF
--- a/packages/pagination/src/index.ts
+++ b/packages/pagination/src/index.ts
@@ -187,7 +187,7 @@ export const createPagination = (
     {
       disabled: { get: () => page() >= opts().pages, set: noop, enumerable: true },
       children: { get: () => opts().nextContent, set: noop, enumerable: true },
-      page: { get: () => Math.max(opts().pages, page() + 1), enumerable: false },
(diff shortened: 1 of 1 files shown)


packages/pagination/src/index.ts
```diff
@@ -187,7 +187,7 @@ export const createPagination = (
     {
       disabled: { get: () => page() >= opts().pages, set: noop, enumerable: true },
       children: { get: () => opts().nextContent, set: noop, enumerable: true },
-      page: { get: () => Math.max(opts().pages, page() + 1), enumerable: false },
+      page: { get: () => Math.min(opts().pages, page() + 1), enumerable: false },
```
thetarnav, line 190: is this correct @atk ?
it does seam like the next page should be capped at the amount of pages.

But shouldn't the line 177 be corrected in this case as well?

atk: Yes, it should indeed be the minimum value here and in line 177.

thetarnav:
Thanks for the PR, I’ve merged #537 which fixes the same issue.

PR #525, opened 2023-10-09
solid-primitives destructure: Added new config option "normalize"

madaxen86:
Added new config option "normalize"
If true all values will be returned as reactive functions, meaning it will resolve values of type MaybeAccessor  to Accessor.
This helps a lot for library devs to provide a flexible interface to users and don't worry about having to make breaking changes later on and the handling of the destructured props is always as Accessor.

Maybe it would also make sense to make this the default behavior, but that would be a breaking change.
Test for spread object smart passes.


madaxen86:
I forgot to change the version in package.json @thetarnav

thetarnav:
Don't change the version. This does it: https://github.com/solidjs-community/solid-primitives/actions/workflows/release.yml

madaxen86:
Feel free to change the naming to normalize or should I do that? I also forgot to add the option in the annotations.

atk:
Please document the option in the README.md. Thanks!

thetarnav:
And add yourself to the contributors in package.json :)

thetarnav:
hmm looks like I’ve missed the change when `normalize` got combined with `memo`. This doesn’t allow for `normalize` and `memo` to be both enabled, why?

madaxen86:
When I experimented with different config combinations, I realised that as soon as normalised was enabled it didn't matter if memo was true or false. 

The returned values were the same, so I decided to merge it with the memo prop to avoid confusion. 

I also discovered that the returned types were not correct for all combinations.

That is all fixed now and also checked by the test I have added.

What it does now it the end is memo static values and returning functions as they are. So the memo config option and now works like charm with all the other options 🤓

packages/destructure/src/index.ts
```diff
@@ -4,30 +4,34 @@ import { access, MaybeAccessor, AnyObject, Values, AnyFunction } from "@solid-pr
 type ReactiveSource = [] | any[] | AnyObject;
 
 export type DestructureOptions<T extends ReactiveSource> = MemoOptions<Values<T>> & {
-  memo?: boolean;
+  memo?: boolean | "normalize";
```
thetarnav, line 7: Yeah I don't like the idea of combining the two. Normalize is about the inputs, while memo is about the outputs. I don't see a point in limiting options to either choosing one or another. `{ memo: boolean, normalize: boolean }` is simpler and lets the user decide on the bahavior.

e.g. here I would like to both normalize the inputs, and memoize the outputs:
```ts
const [count, setCount] = createSignal(1)

const {foo, bar} = destructure({
	get foo() {return count() > 5},
	bar: () => count() > 5
}, {memo: true, normalize: true})

createEffect(() => {console.log("foo", foo())})
createEffect(() => {console.log("bar", bar())})

setCount(2) // shouldn't rerun effects
setCount(6) // should rerun effects
```

or rely on the original singals:

```ts
const [count, setCount] = createSignal(1)

const {foo, bar} = destructure({
	get foo() {return count() > 5},
	bar: () => count() > 5
}, {normalize: true})

createEffect(() => {console.log("foo", foo())})
createEffect(() => {console.log("bar", bar())})

setCount(2) // should rerun effects
setCount(6) // should rerun effects
```

packages/destructure/src/index.ts
```diff
@@ -75,18 +81,24 @@ export function destructure<T extends ReactiveSource, O extends DestructureOptio
...
+  : Spread<T, O["memo"] extends "normalize" ? true : false> {
   const config: DestructureOptions<T> = options ?? {};
   const memo = config.memo ?? typeof source === "function";
-  const getter =
-    typeof source === "function"
-      ? (key: any) => () => source()[key]
-      : (key: any) => () => source[key];
+  const getter = (key: any) => {
+    const value = () => (typeof source === "function" ? source()[key] : source[key]);
+    if (config.memo !== "normalize") return value;
```
thetarnav, line 94: no need to check the config and if the source is a function every time a property is accessed

thetarnav (changes requested):

packages/destructure/src/index.ts
madaxen86: I added this this to the test.

packages/destructure/src/index.ts
madaxen86: moved that into a createMemo outside of the getter.

madaxen86:
Please review the latest commit.
Added tests for the comments and made adjustments accordingly.

packages/destructure/src/index.ts
```diff
@@ -106,7 +119,11 @@ export function destructure<T extends ReactiveSource, O extends DestructureOptio
     const calc = getter(key);
     if (config.deep && isReactiveObject(value))
       result[key] = destructure(calc, { ...config, memo });
-    else result[key] = memo ? createMemo(calc, undefined, options) : calc;
+    else
+      result[key] =
+        memo && (!config.normalize || calc.length === 0)
```
thetarnav, line 124: There is one thing I'm realizing about this.
```ts
const [callback, setCallback] = createSignal((a, b) => a + b)

const {cb} = destructure({get cb() { return callback() }})

setCallback(() => () => 69)

expect(cb()).toBe(69) // fail: still calls the original function
```
Even if we want to avoid `cb()(1, 2)` syntax of normal destructure, maybe it should still support "reactive callbacks", the same `props.callback(1, 2)` would if you passed `callback={someSignal()}` in JSX.
Although Solid doesn't allow for reactive callbacks in `on___` props, so it may be fine to restrict that as well.

thetarnav (commented):
Ok, besides the reactive callback stuff that I still need to think about and some minor stuff to correct and it can be merged I think.

packages/destructure/test/destructure.test.ts
```diff
@@ -1,6 +1,7 @@
-import { describe, test, expect } from "vitest";
 import { createComputed, createRoot, createSignal } from "solid-js";
+import { describe, expect, test } from "vitest";
 import { destructure } from "../src/index.js";
+import { get } from "@solid-primitives/utils/immutable";
```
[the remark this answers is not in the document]
madaxen86: Whooopsy 🤪

packages/destructure/test/destructure.test.ts
```diff
@@ -246,6 +247,142 @@ describe("destructure", () => {
...
+          },
+          _y: () => count() > 5,
+        },
+        { normalize: true },
+      );
+
+      const [_numbers, setNumbers] = createSignal({
+        a: 3,
+        b: () => (toggle() ? 2 : 3),
+        c: (a: number, b: number) => a * b,
```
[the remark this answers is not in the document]
madaxen86: I have adjusted the types.
It became a bit messy but now the types are inferred correctly.

thetarnav:
Another question about the callbacks.
How is it supposed to resolve this situation?
```ts
type MaybeAccessor<T> = T | Accessor<T>
type Props = {
	handleClick: MaybeAccessor<(e: MouseEvent) => void>
}
const logMousePos = (e: MouseEvent) => {
	console.log("mouse", e.x, e.y)
}

// both fine
const a = destructure<Props>({handleClick: () => logMousePos})
const b = destructure<Props>({handleClick: logMousePos})

a(e) // => works
b(e) // => works

// variadic params
const log = (...a) => {console.log(...a)}

// ts doen't mind
const a = destructure<Props>({handleClick: () => log}) // fine
const b = destructure<Props>({handleClick: log}) // this logs immediately

a(e) // ok
b(e) // runtime error - tried to call undefined
```

madaxen86:
Good call.
This is because the 'access' function of @solid-primitives/utils does not check for variadic params (it only does !func.length 
I created a new function 'getNormalizedValue' which fixes this issue and also added a test for any kind of combination with variadic params (including nested inside a function without params etc.).
Question:
In this line of the getter
´´´
if (typeof accessedValue() === "function" && hasParams(accessedValue())) return accessedValue();
´´´
shouldn't we only return if the config  normalize:true ?
Otherwise we create a breaking change.

madaxen86:
> Another question about the callbacks. How is it supposed to resolve this situation?
> 
> ```ts
> type Props = {
> 	handleClick: (e: MouseEvent) => void
> }
> const logMousePos = (e: MouseEvent) => {
> 	console.log("mouse", e.x, e.y)
> }
> 
> // fine
> const a = destructure<Props>({handleClick: logMousePos})
> 
> a.handleClick(e) // => works
> 
> // variadic params
> const log = (...a) => {console.log(...a)}
> 
> // ts doen't mind
> const b = destructure<Props>({handleClick: log}) // logs immediately
> 
> b.handleClick(e) // runtime error - tried to call undefined
> ```
Good call.
This is because the 'access' function of @solid-primitives/utils does not check for variadic params (it only does !func.length
I created a new function 'getNormalizedValue' which fixes this issue and also added a test for any kind of combination with variadic params (including nested inside a function without params etc.).
Question:
In this line of the getter

if (typeof accessedValue() === "function" && hasParams(accessedValue())) return accessedValue();

shouldn't we only return if the config normalize:true ?
Otherwise we create a breaking change.

madaxen86:
@thetarnav Any chance this moves forward or should I submit it as a new primitive?

thetarnav:
I'm sorry for not responding. Looking back at it, I'm still not convinced that this direction is the right one. Checking whether something is a signal or not goes against Solid's principles, it is intentionally designed to make such determinations difficult. When you need to resort to stringifying and testing a function's definition with regex look for keywords more common in callbacks than signals, it's a strong indicator that this may not be the best approach. It's not accurate, it's slow, and it's trying to be too clever.

Unfortunately, the alternative I have doesn't satisfy me either. Solid's compiler treats props starting with `on___` as callbacks (like `onClick`, `on:click`, `oncapture:click`, etc.), but it restricts their reactivity. So, when destructuring props, assuming that anything not starting with "on" could be a signal might not be foolproof. This heuristic doesn't apply to stores, which could also be used with `destructure`, and callbacks without the "on"  prefix will still be broken.

I'm hesitant to introduce something that users won't have the convenience that it will not cause more problems, when the only one that they wanted to solve was less typing.

PR #531, opened 2023-10-21
singlas -> signals in Readme

danieltroger:
I hate people that just make changes to README.md so it says on their github profile that they contributed but I legit got confused what singlas are and thought this (a PR) was more helpful than opening an issue


thetarnav:
All the singlas -> signals are welcome 😊

PR #537, opened 2023-11-02
Fix of back and next page values

muhammedf:
Back was always returning first page and next was returning last page.
This pr fixes that.


muhammedf:
Same thing with #522. I just saw it, right after I sent this pr. Well, I wrote a test too, so it is something.

thetarnav:
Thank you!

PR #567, opened 2024-01-17
Fix `i18n` docs

asqarslanov:
Just some small changes affecting only one line under the `Modules` section.

#### Before:

```md
For example if out app had…
```

#### After:

```md
For example, if our app had…
```


thetarnav:
thanks :+1:

Issue #550, opened 2023-12-15
Make i18n work with JSX


BierDav:
### Describe The Problem To Be Solved

Currently, there is no way how to format an i18n value. For example, you have a user variable, and you want to display the user in bold. Like this
```
"complete_registration_subtitle": "Hi <b>{{ name }}!</b"
```

This currently doesn't work.

### Suggest A Solution

I propose that all translations should return JSX so that such formatting can be done.

thetarnav:
you can do this
```tsx
"complete_registration_subtitle": () => <>Hi <b>{{ name }}!</b><>
```
or
```tsx
<div innerHtml={t.complete_registration_subtitle()} />
```
or create a custom template resolver that parses html and returns html elements.

BierDav:
Unfortunately this causes weird issues with solid start (or seroval 🙂) on the latest available version 0.3.10:

```
Error
Seroval caught an error during the parsing process.
  
Error
The value [object Function] of type "function" cannot be parsed/serialized.
      
There are few workarounds for this problem:
- Transform the value in a way that it can be serialized.
- If the reference is present on multiple runtimes (isomorphic), you can use the Reference API to map the references
```

thetarnav:
Thats interesting. To be honest I have no idea what solid start chooses to serialize and what not, I don't think the i18n lib is doing anything wring, in the end you are just calling a function and getting some values back, and here the values happen to be html elements.
A reproduction would be helpful regardless if this is an issue with solid start or solid primitives.

BierDav:
Here you go: https://stackblitz.com/~/github.com/BierDav/solid-primitives-i18n-bug

The problem is that createResource is not able to work with functions as result value I guess: https://github.com/BierDav/solid-primitives-i18n-bug/blob/e27d9bed2350c0c20800b6c6502318443a42d077/src/App.tsx#L18-L24

BierDav:
`ssrLoadFrom: "initial"` fixes the issue for me, thanks for rubber ducking 🙈

BierDav:
No, actually this is still an issue, because I don't want to use ssrLoadFrom: "initial", because I need to fetch the current language in ssr, I can't simply render everything in english.

BierDav:
[...]

PR #568, opened 2024-01-18
Add support for Generic Template Resolvers

BierDav:
Also added NullableChainedTranslators

This should resolve: #550

### DIFF
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -213,7 +213,7 @@ t2("hello", { name: "John" }); // => 'hello John!'
 
 Splitting the dictionary into multiple modules can be useful when you have a large dictionary and want to avoid loading the entire dictionary at once.
 
-For example if out app had a separate `login` and `dashboard` modules, we could split the dictionary into 3 modules: (`common`, `login` and `dashboard`).
+For example, if our app had a separate `login` and `dashboard` modules, we could split the dictionary into 3 modules: (`common`, `login` and `dashboard`).
 
 ```
 i18n/


BierDav:
Btw. this would be an possible implementation for a template resolver that makes use of this generics:
```

type CtI18nTemplateArgs = string | number | boolean | ((children?: JSX.Element) => JSX.Element);
  function resolveJSXTemplate(string: string, wrappers: {
        [key: string]: ((children?: JSX.Element) => JSX.Element)
    }, regex: RegExp): JSX.Element {
        const matches = Array.from(string.matchAll(regex))
        const resultArray = []
        let index = 0
        for (let match of matches) {
            resultArray.push(string.slice(index, match.index))
            index = match.index! + match[0].length
            const wrapper = wrappers[match[1]]
            resultArray.push(wrapper(match[3] ? resolveJSXTemplate(match[3], wrappers, regex) : undefined))
        }
        resultArray.push(string.slice(index))
        return resultArray
    }

    function resolveTemplate(string: string, args?: BaseTemplateArgs<CtI18nTemplateArgs>) {
        if (!args)
            return string;
        const wrappers: { [key: string]: ((children: JSX.Element) => JSX.Element) } = {}
        for (let prop in args)
            if (typeof args[prop] === 'function') {
                wrappers[prop] = args[prop] as ((children: JSX.Element) => JSX.Element)
                delete args[prop];
            }
        const result = i18n.resolveTemplate(string, args as BaseTemplateArgs)
        if (Object.keys(wrappers).length == 0)
            return result
        const regexExp = new RegExp(`{{\\s*(${Object.keys(wrappers).join("|")})\\s*}}((.*?){{\\s*(\\1)_end\\s*}})?`, "g")
        return resolveJSXTemplate(result, wrappers, regexExp)
    }

    const translator = i18n.translator(dict, resolveTemplate);
```

```
{
"hello_world": "Hello {{link_wrapper}}Link{{link_wrapper_end}}"
}
```

BierDav:
Can someone review this, the test is not failing due to an error of mine

thetarnav:
I should not be allowed near git lol, I'll try to fix this

BierDav:
I think I cannot reopen the merge request either, btw. hmm i haven't thought about that usecase yet, but fell free to make string a default, if it makes sense

BierDav:
ahh i know why you cannot reopen, because of an force push on the main branch, but if you have the changes in a separate branch it doesn't matter anyways

PR #570, opened 2024-01-20
Add sync functionality to `makePersisted`

BierDav:
Add deprecation documentation to all typescript definitions for createStorage, because some IDE's don't support transferring documentation across definitions

### DIFF
--- a/packages/storage/src/persisted.ts
+++ b/packages/storage/src/persisted.ts
@@ -1,9 +1,9 @@
-import {Accessor, onCleanup, onMount, Setter, Signal} from "solid-js";
-import {createUniqueId, untrack} from "solid-js";
-import type {SetStoreFunction, Store} from "solid-js/store";
-import {reconcile} from "solid-js/store";
-import type {AsyncStorage, AsyncStorageWithOptions, StorageWithOptions} from "./types.js";
-import {getSyncBroadcastName} from "./tools.js";
+import { Accessor, onCleanup, onMount, Setter, Signal } from "solid-js";
+import { createUniqueId, untrack } from "solid-js";
+import type { SetStoreFunction, Store } from "solid-js/store";
+import { reconcile } from "solid-js/store";
+import type { AsyncStorage, AsyncStorageWithOptions, StorageWithOptions } from "./types.js";
+import { getSyncBroadcastName } from "./tools.js";
 
 export type PersistenceBaseOptions<T> = {
   /**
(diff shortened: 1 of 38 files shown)


packages/storage/src/persisted.ts
```diff
@@ -1,23 +1,44 @@
-import type { Accessor, Setter, Signal } from "solid-js";
-import { createUniqueId, untrack } from "solid-js";
-import type { SetStoreFunction, Store } from "solid-js/store";
-import { reconcile } from "solid-js/store";
-import type { AsyncStorage, AsyncStorageWithOptions, StorageWithOptions } from "./types.js";
+import {Accessor, onCleanup, onMount, Setter, Signal} from "solid-js";
```
atk, line 1: What's with the formatting? Did someone change the settings?

packages/storage/src/persisted.ts
```diff
@@ -152,33 +173,73 @@ export function makePersisted<T, O extends Record<string, any> = {}>(
...
+    const value = storage.getItem(name, storageOptions);
+    if (value)
+      if (typeof value === "object" && "then" in value) // Check if value is a promise
+        value.then(data => unchanged && data && set(data));
+      else set(value);
+  }
+
+  fetchFromStorage() // Initialize the underlying signal
+
+  let broadcastChannel: BroadcastChannel | undefined
```
atk, line 192: I'd prefer a mechanism that allows to define your own IPC, with BroadcastChannel as a fallback.

packages/storage/src/persisted.ts
```diff
@@ -152,33 +173,73 @@ export function makePersisted<T, O extends Record<string, any> = {}>(
...
+        storage.setItem(name, value, storageOptions);
+        onChanged()
+      },
   ] as typeof signal;
 }
+
+
+interface SyncBroadcastMessage {
+  /**
+   * This property makes it easier for developers to understand the origin of the message if they accidentally use the
```
atk, line 240: That's a good idea, but couldn't we instead have a Set of registered names and warn users on re-use?

atk (changes requested):

packages/storage/src/persisted.ts
BierDav: what do mean by that?

packages/storage/src/persisted.ts
BierDav: i don't really see the use case, because if you want to handle it your self you can just wrap the signal with `sync: false` and handle the sync for your self. BroadcastChannels are the officially preferred way of inter window and tab communication so i also don't see why we should give first party support for something like `localstorage` change events.

packages/storage/src/persisted.ts
atk: Shouldn't the name be bound to the key in which they property is persisted? So we should avoid binding the same property to different keys as well as different keys to the same property.

packages/storage/src/persisted.ts
atk: But otherwise, we carry useless code that is difficult to tree-shake, and in most cases, you can get by without sync. So IMO it would be better to have an API and one BroadcastChannel-based plugin to use it.

atk:
This is now already solved.

PR #571, opened 2024-01-24
docs: duplicate keys

Joehoel:


thetarnav:
Thanks.
`createDropzone` allows many callbacks, so I'm not sure if there was supposed to be other one used instead. But better to remove it than confuse people.

PR #573, opened 2024-02-22
Change method used for tweening

jamiemlaw:
This alternative implementation calculates the tweened value as a function of its start and target values, rather than its current and target values.

This fixes an issue when, for example, setting a duration of five seconds and a linear easing function, the tween would not animate linearly, and would be close to its target value in under a second.


packages/tween/src/index.ts
```diff
@@ -21,21 +21,32 @@ export default function createTween<T extends number>(
...
+  let startValue = target();
+  let start = 0;
+  let cancelId: number;
+
+  function tick(t: number) {
+    const elapsed = t - start;
+
+    if (elapsed < duration) {
+      // @ts-ignore
+      setCurrent(startValue + ease(elapsed / duration) * (target() - startValue));
```
thetarnav, line 34: `target() - startValue` - this probably could be calculated in the effect

packages/tween/src/index.ts
```diff
@@ -21,21 +21,32 @@ export default function createTween<T extends number>(
...
-  );
+  let startValue = target();
+  let start = 0;
+  let cancelId: number;
+
+  function tick(t: number) {
+    const elapsed = t - start;
+
+    if (elapsed < duration) {
+      // @ts-ignore
```
thetarnav, line 33: Are the `@ts-ignore` still necessary?
I think it would be simpler to just make the `current` have a `number` type, and typecast it when returning.

packages/tween/src/index.ts
```diff
@@ -21,21 +21,32 @@ export default function createTween<T extends number>(
...
-        const elapsed = t - (start() || 0) + 1;
-        // @ts-ignore
-        setCurrent(c =>
-          elapsed < duration ? (target() - c) * ease((elapsed / duration) as T) + c : target(),
-        );
-      });
-      onCleanup(() => cancelAnimationFrame(cancelId));
-    }),
-  );
+  let startValue = target();
```
thetarnav, line 25: why assign `startValue` here? It will be assigned in the effect.

thetarnav (changes requested):
I link this implementation more. Could you add a unit test that ensures the mentioned issue is working? This package doesn't have any so even a single one would be great.

packages/tween/src/index.ts
jamiemlaw: Not a bad idea. I will call it something like, `delta`.

packages/tween/src/index.ts
jamiemlaw: Unless there's something I've missed, it seems a lot simpler to use `number`s for everything instead of generics. Maybe when there's an update that allows `createTween` to tween between arbitrary types like strings or colours, there would be reason to revisit this. In any case, it fixes all the TS errors.

jamiemlaw:
I started writing some tests, but I think they could be improved.

The core thing I was trying to improve with this PR was an issue where the animation seemed to be taking the wrong amount of time, and wasn't easing in the way I would expect. To test for that sort of thing required stubbing `requestAnimationFrame`, which I've done with a third party library.

One thing I notice I haven't done is wrap my tests in a call to `createRoot`. When I did this, the timers stopped working correctly, and I wasn't sure why. So I've committed what I've done so far, all the tests pass, and they also don't pass when testing against `main`, which is a good thing.

packages/tween/src/index.ts
thetarnav: I'm not sure why it was generic either, it's definitely easier to just remove it.
But I like the idea of making it tween arbitrary types.

thetarnav:
It might be simpler to mock the raf without a dependency, this way no one has to research that library to figure out what exactly it's doing. like here: https://github.com/solidjs-community/solid-transition-group/blob/main/test/index.test.tsx. But it's fine if you don't want to.

> One thing I notice I haven't done is wrap my tests in a call to `createRoot`. When I did this, the timers stopped working correctly, and I wasn't sure why.

Maybe solid's scheduling gets in the way of effects, you don't want update any signals or call any actions in the root/effect synchronously, but outside of it so that the effects queue is flushed immediately, just like it would when you would do that in eg a event handler. https://github.com/solidjs-community/solid-primitives/blob/i18n-generic-resolvers/packages/transition-group/test/list-transition.test.ts#L48

PR #577, opened 2024-03-04
Colorize scheduled playground

noriapi:
Changed the scheduled DEMO to make it easier to understand.
Specifically, each cell is now colored in the same color when the source is the same.


thetarnav:
thats a nice idea, thank you!

PR #579, opened 2024-03-05
Fix package.json homepages

deluksic:
As agreed on Discord, this PR fixes all `homepage` fields in packages to point to the correct package on `primitives.solidjs.community`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "solid-primitives",
   "private": true,
   "description": "A collection of high-quality, community contributed building blocks.",
+  "homepage": "https://primitives.solidjs.community/",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/solidjs-community/solid-primitives.git"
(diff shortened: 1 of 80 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "solid-primitives",
   "private": true,
   "description": "A collection of high-quality, community contributed building blocks.",
+  "homepage": "https://primitives.solidjs.community/",
```
deluksic, line 5: I understand this is not a package, but still it might be nice to have the homepage here.

thetarnav:
Thank you
although maybe a link to https:\/\/primitives.solidjs.community/package/package_name would be better?

deluksic:
> maybe a link to https://primitives.solidjs.community/package/package_name would be better?

I thought about this, but didn't want to make this change now, just wanted to fix the existing setup. I can also make this PR but later.

deluksic:
I checked all links and website does not have entries for `analytics`, `gestures`, `utils` and `template`.

thetarnav:
utils is "internal" so it never had a readme or homepage

the other two were not finished so are omitted as well

thetarnav:
can I merge this or are you doing something still?

deluksic:
good to go

PR #583, opened 2024-03-16
List component with reactive value & index

maciek50322:
Hello,
There's `<For>` with reactive index, `<Index>` with reactive value, and here is `<List>` with reactive index & value.
Because it "merges" their behaviours, it has priorities for each element. If higher priority doesn't work, next one is applied:

1) when new and old values and indexes match: don't change anything about element
2) when new and old values match: change old index to new
3) when new and old index match: change old value to new
4) when nothing about new element matches old element: change index and value of old element to new
5) when new element left but no old elements: create new element
6) when old element left, but no new elements: delete old element

Also made signals for value & index to be created only if used, so you can have signal only for index without value if you don't need it. Also `(...args) => ...` works.

This is my first pr here, happy to hear feedback.


thetarnav:
Very interesting, thanks for putting your time into this, it appears very solid.
I never thought about that, but now that I see it, it seems obvious to have that "For variant". Not sure whats the usecase though yet.
Here is some feedback from first looks:
- It would be wise to mention in the docs that it creates a "unkeyed" control flow - there the html elements are shared between different values, so it cannot be used with libraries like transition-group in the same way.
  For example you can look at [Index](https://docs.solidjs.com/reference/components/index-component) or [createRootPool](https://github.com/solidjs-community/solid-primitives/tree/main/packages/rootless#usage-with-for) docs
- Do you have some examples of "real world" usages of this pattern? Like why would I necessarely use it instead of `For`. The docs only say how it works, not why should I care.
- You've mentioned 6 behaviors in your message, can you add simple unit tests for those so we have a sanity check thats it's working correctly?
- > Also `(...args) => ...` works.
  
  You might want to propose the solution for that for Solid's `For` as well, I don't remember if something like that was considered.

fabiospampinato:
I think the use case is basically just better performance in some cases. I use something like this for virtualized lists, and together with a built-in pooling capability it unlocks doing very little work when scrolling the list, less work than either For or Index would be able to achieve.

maciek50322:
Thanks for Feedback @thetarnav 

- About unkeyed, it's not quite unkeyed, because sometimes it is, sometimes not. It is when the index signal is changed and isn't when value & index signals are changed, so you might have all animations going correct way until it's more efficient to reuse existing node. So not sure what to do about that.
- For real world examples, I don't think there's something you can't achieve when there's already `<For>` and `<index>`. I actually treat this as optimization of `<For>`, when you don't need it to be necessarily keyed (don't use animation of sorting etc.).  Also I treat this as solution for 'most cases', when some are covered by `<For>`, some by Index, this has behaviours of both, so very useful for beginners, eg. sometimes you hear that "If you don't know which one to use, use <For>". I think this should be better solution, for example
```tsx
<For each={store}>
{(v, i) => <input value={v} onInput={(e) => setStore(i(), e.target.value)} />}
</For>
```
each time you type you lose focus of input. `<Index>` here should be fine, but also the `<List>`
- I'll try to write these tests
- I'll propose it later

fabiospampinato:
> About unkeyed, it's not quite unkeyed, because sometimes it is, sometimes not. 

I think for more people the definition of "unkeyed" is "something that is not always keyed", so this qualifies.

> when you don't need it to be necessarily keyed (don't use animation of sorting etc.)

By the way there's yet another version of this kind of component that doesn't break animations, I haven't implemented it yet but presumably it works, the idea is that a result for an item is not reused for another item right away, but only after the browser has rendered once, this way the animation state of the elements should be reset and the animation bugs disappear, while technically nodes.

maciek50322:
> By the way there's yet another version of this kind of component that doesn't break animations, I haven't implemented it yet but presumably it works, the idea is that a result for an item is not reused for another item right away, but only after the browser has rendered once, this way the animation state of the elements should be reset and the animation bugs disappear, while technically nodes.

Interesting, I'll see if I can do that too here

maciek50322:
So 
- added tests, 
- mentioned "unkeyed", 
- fixed signal creation if used (scope catched changing variable, and using signal "later" would have it set to wrong value)
- added pnpm-lock - please check if it is set right, because some other things changed
- for animations on second thought I think it might be achieved with some wrapper, that can be maybe reused for `<For>`, so I'll leave that for now

thetarnav:
I don't have anything else to add, this looks great. Gonna do an initial release and look for places to try it out.

PR #584, opened 2024-03-20
Switch to `vinxi` and `@solidjs/start` + Fix issues that cookieStorage had with stackblitz and solid start

BierDav:
Recently I noticed that `solid-js/web` has an export function called getRequestEvent which completely solves the need to dynamically load a module using import. This should make it possible to solve the issues we had with this dynamic import.

I also have fixed the formatting of some code in the README.md, removed a seamingly outdated documentation (even the word `_cookies` is not contained anywhere in the whole codebase and removed the part that explained the issues we had with stackblitz and solid start.

I also have updated the peer dependency of `solid-start` to the new package `@solidjs/start` and with that also used the latest versions of solid and solid-start.

The site package now also utalizes the new package `@solidjs/start` and the build tool that is recommended for solidjs `vinxi`.

I'm not 100% sure if it works, because I haven't tested it. For that I need an dev or alpha version to be released. Can somebody do that for me please?


BierDav:
Ok guys, I have now also ported the documentation site to vinxi, which most likely broke some of the ci. Please can someone help me, fix the rest. There is also minor bug, that only some of the text remains black in dark mode which was white in the version before.

BierDav:
The import from `@solidjs/start/server` doesn't work, because the types are not exported properly. Therefore we have to wait for this pr of mine before merging: https://github.com/solidjs/solid-start/pull/1406

But nevertheless I would be cool if someone could look at the site package in the meanwhile, if that looks good. I'm not quite sure if I though of everything and tested everything well

atk:
I hate to be that guy, but this are too many changes in one PR to reasonably review.

BierDav:
@atk most of the files are only because of pnpm format which also updated many files to lf from clrf, i have now added a .gitattribute file so that git automatically converts files to lt

atk:
I am not against these changes, but I would prefer to have them in separate PRs. I just tested the function to get the cookies, which works fine in solid-start. It would be great if we could use a similar way to set cookies, too.

BierDav:
If you hit `Hide Whitespaces` only 30 files have really changed:
image

BierDav:
> I am not against these changes, but I would prefer to have them in separate PRs. I just tested the function to get the cookies, which works fine in solid-start. It would be great if we could use a similar way to set cookies, too.

hmm ok, i try to remove the cookie stuff from here and focus on the solid-start and vinxi transition in this pr, because this is neccessary for the cookie stuff

atk:
Maybe we can base my upcoming changes upon yours.

BierDav:
Jup, i will not not make a second pr with the remaining changes, because you have already included everything. I will close this pr now in favor of my new more specific one

atk:
Thanks for your work, I couldn't have done it without your help.

PR #586, opened 2024-03-21
Storage reloaded

atk:
This is a major rework of the storage/persistence API and the first step before fully deprecating the old APIs. This adds the sync API and muxers (multiplerxers) for sync and storage and also a withOptions method that makes handling cookieStorage with options a lot more convenient.

This needs more documentation (ie. recipes how to use cookieStorage on a server with solid classic SSR instead of solid-start).

### DIFF
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -65,18 +65,10 @@
     "@solid-primitives/utils": "workspace:^"
   },
   "peerDependencies": {
-    "solid-js": "^1.6.12",
-    "@solidjs/start": "^0.6.0",
-    "vinxi": "^0.3.9"
-  },
-  "peerDependenciesMeta": {
-    "@solidjs/start": { "optional": true },
-    "vinxi": { "optional": true }
+    "solid-js": "^1.6.12"
   },
   "devDependencies": {
-    "solid-js": "^1.8.15",
-    "@solidjs/start": "^0.6.0",
-    "vinxi": "^0.3.9"
+    "solid-js": "^1.8.15"
   },
   "typesVersions": {}
 }
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -16,14 +16,15 @@
     "name": "storage",
     "stage": 3,
     "list": [
-      "createStorage",
-      "createCookieStorage",
-      "createAsyncStorage",
-      "createStorageSignal",
-      "createLocalStorage",
-      "createSessionStorage",
+      "makePersisted",
       "cookieStorage",
-      "makePersisted"
+      "multiplexStorage",
+      "storageSync",
+      "messageSync",
+      "wsSync",
+      "multiplexSync",
+      "addClearMethod",
+      "addWithOptionsMethod"
     ],
     "category": "Browser APIs"
   },
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -66,16 +66,17 @@
   },
   "peerDependencies": {
     "solid-js": "^1.6.12",
-    "solid-start": ">=0.2.26"
+    "@solidjs/start": "^0.6.0",
+    "vinxi": "^0.3.9"
   },
   "peerDependenciesMeta": {
-    "solid-start": {
-      "optional": true
-    }
+    "@solidjs/start": { "optional": true },
+    "vinxi": { "optional": true }
   },
-  "typesVersions": {},
   "devDependencies": {
-    "solid-js": "^1.8.7",
-    "solid-start": "^0.3.10"
-  }
+    "solid-js": "^1.8.15",
+    "@solidjs/start": "^0.6.0",
+    "vinxi": "^0.3.9"
+  },
+  "typesVersions": {}
 }
--- a/packages/storage/src/cookies.ts
+++ b/packages/storage/src/cookies.ts
@@ -63,10 +63,9 @@ const setCookie = isServer
       return;
     }
     const serializedOptions = serializeCookieOptions(options);
-    responseHeaders.set(
+    responseHeaders.append(
       "Set-Cookie",
-      `${responseHeaders.get("Set-Cookie") ? ", " : ""}${key}=${value}${serializedOptions ? "; " : ""
-      }${serializedOptions}`,
+      `${key}=${value}${serializedOptions ? "; " : ""}${serializedOptions}`,
     );
   }
   : () => undefined;
--- a/packages/storage/src/cookies.ts
+++ b/packages/storage/src/cookies.ts
@@ -1,4 +1,4 @@
-import { isServer } from "solid-js/web";
+import { isServer, getRequestEvent, type RequestEvent } from "solid-js/web";
 import { StorageProps, StorageSignalProps, StorageWithOptions } from "./types.js";
 import { addClearMethod } from "./tools.js";
 import { createStorage, createStorageSignal } from "./storage.js";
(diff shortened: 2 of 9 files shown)


packages/storage/src/cookies.ts
```diff
@@ -52,20 +51,25 @@ function deserializeCookieOptions(cookie: string, key: string) {
...
+    const responseHeaders = (
+      getRequestEvent() as unknown as RequestEvent & { response?: { headers: Headers } }
+    )?.response?.headers;
+    if (!responseHeaders) {
+      return;
+    }
+    const serializedOptions = serializeCookieOptions(options);
+    responseHeaders.set(
+      "Set-Cookie",
+      `${responseHeaders.get("Set-Cookie") ? ", " : ""}${key}=${value}${serializedOptions ? "; " : ""
```
BierDav, line 68: What is this `,` at the beginning for?

packages/storage/src/cookies.ts
```diff
@@ -89,45 +93,16 @@ try {
...
-      }
+    ? (key: string, value: string, options?: CookieOptions) =>
+      (options?.setCookie ?? setCookie)?.(key, value, options as CookieProperties)
     : (key: string, value: string, options?: CookieOptions) => {
-        document.cookie = `${key}=${value}${serializeCookieOptions(options)}`;
-      },
+      document.cookie = `${key}=${value}${serializeCookieOptions(options)}`;
+    },
   getItem: (key: string, options?: CookieOptions) =>
     deserializeCookieOptions(cookieStorage._read(options), key),
```
BierDav, line 107: This `window.dispatchEvent` call should be unnecessary when we already have this new shiny sync or isn't it?

packages/storage/src/persisted.ts
```diff
@@ -162,23 +179,88 @@ export function makePersisted<T, O extends Record<string, any> = {}>(
...
+        (signal[1] as any)(...args);
+        const value = serialize(untrack(() => signal[0] as any));
+        options.sync?.[1](name, value);
+        // @ts-ignore
+        storage.setItem(name, value, storageOptions);
+        unchanged = false;
+      },
   ] as typeof signal;
 }
+
```
BierDav, line 217: Btw. I am desperatly missing a Broadcast Sync, because as far as I know this is the official preferred way to communicate between tabs and windows

packages/storage/src/tools.ts
```diff
@@ -17,3 +17,81 @@ export const addClearMethod = <S extends Storage | StorageWithOptions<any>>(
...
+      } as Partial<S>,
+    ) as unknown as W;
+  return storage as S & { withOptions: (options: O) => W };
+};
+
+const callStorageMethod = (
+  storages: (Storage | AsyncStorage)[],
+  method: string,
+  args: any[] = [],
+  first = false,
```
BierDav, line 56: I'm not sure if Promise.race is beneficial for debugging. Because if one storage is faulty it might result in errors that are not consistantly reproducable. I think we should add an option that enables this and mention it in the documenation that this is an optimization, but the developer should be aware of what happens in the background

packages/storage/package.json
```diff
@@ -66,16 +66,17 @@
   },
   "peerDependencies": {
     "solid-js": "^1.6.12",
-    "solid-start": ">=0.2.26"
+    "@solidjs/start": "^0.6.0",
+    "vinxi": "^0.3.9"
```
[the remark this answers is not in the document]
atk: It is indeed no longer needed and has been removed.

packages/storage/src/cookies.ts
atk: If there are already cookies in the request, we do not want to overwrite them, but to append.

packages/storage/src/cookies.ts
```diff
@@ -89,45 +93,16 @@ try {
...
-          result +=
-            responseHeaders
-              .get("Set-Cookie")
-              ?.split(",")
-              .map(cookie => !cookie.match(/\\w*\\s*=\\s*[^;]+/))
-              .join(";") ?? "";
-        }
-        return `${result};${request?.headers?.get("Cookie") ?? ""}`; // because first cookie will be preferred we don't have to worry about duplicates
-      }
+      const headers = options?.getRequestHeaders?.() || getRequestHeaders();
```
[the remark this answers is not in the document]
atk: I'm appending cookies now, so that is no longer an issue.

packages/storage/src/cookies.ts
atk: That is indeed correct; it never worked as expected in any case, since it is not propagated to other tabs.

packages/storage/src/tools.ts
atk: But what would be the other option? We cannot infer which API is faulty and which is not, even if we wait for all of the results, which is why I stress in the README that you have to await for writing access in order to avoid race conditions.

packages/storage/src/persisted.ts
atk: Good point. I'll add it tomorrow, it's a bit late for me now.

packages/storage/src/cookies.ts
```diff
@@ -52,20 +51,24 @@ function deserializeCookieOptions(cookie: string, key: string) {
...
+const setCookie = isServer
+  ? (key: string, value: string, options: CookieProperties) => {
+    const responseHeaders = (
+      getRequestEvent() as unknown as RequestEvent & { response?: { headers: Headers } }
+    )?.response?.headers;
+    if (!responseHeaders) {
+      return;
+    }
+    const serializedOptions = serializeCookieOptions(options);
+    responseHeaders.append(
```
BierDav, line 66: I have a bad feeling there, because we make the set cookie string indefinitely longer each time a vhange happens neither in the deserialization nor in the serialization we remove the unnecessary values, we should definitely consider cleaning up the set cookies header.

packages/storage/src/cookies.ts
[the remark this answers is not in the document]
BierDav: Not a fan of that, because the request header is in my opinion considered immutable, because the request didn't change and when i am querying the request header i want exactly the request real request headers and not the response ones

packages/storage/src/tools.ts
BierDav: hmm okay i guess, i fine with that

packages/storage/src/cookies.ts
atk: Good point. That's why I originally switched to set.

packages/storage/src/cookies.ts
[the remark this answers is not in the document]
atk: But we're handling the response headers here, not the request headers.

packages/storage/src/cookies.ts
atk: No, I wanted set to clean up previous cookies for the same key. Let's do that in two steps, I guess.

packages/storage/src/cookies.ts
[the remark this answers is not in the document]
atk: Ah, I misunderstood which code you were referring to. Now that is a problem. I will take a bit of time to think of a better solution.

packages/storage/src/cookies.ts
atk: Now I only need to reconcile and cleanup request and response cookies (the latter should take precedence) and then we can proceed.

packages/storage/src/persisted.ts
atk: I changed postMessageSync to messageSync, so it can handle both post message and broadcast channel API.

packages/storage/src/cookies.ts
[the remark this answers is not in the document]
BierDav: My solution was to concat the request header and the response header, so that the response header overwrites the values of the request header.

packages/storage/src/cookies.ts
```diff
@@ -63,9 +63,10 @@ const setCookie = isServer
       return;
     }
     const serializedOptions = serializeCookieOptions(options);
-    responseHeaders.append(
+    const cleanedCookies = responseHeaders.get("Set-Cookie").replace(new RegExp(`(^|, )${key}=[^,]+,?`, "g"), "$1");
```
BierDav, line 66: why replacing with`$1` and not with an emtpy string?

packages/storage/src/persisted.ts
```diff
@@ -229,11 +229,11 @@ export const storageSync: PersistenceSyncAPI = [
 ];
 
 /**
- * postmessageSync - synchronize over post message API
+ * messageSync - synchronize over post message or broadcast channel API
  */
-export const postmessageSync: PersistenceSyncAPI = [
+export const messageSync = (channel: Window | BroadcastChannel = window): PersistenceSyncAPI => [
```
BierDav, line 234: Good job, that's quite elegant in my opinion

packages/storage/src/cookies.ts
atk: Because otherwise, I would potentially remove too many commas; but maybe, I'll just remove the `,?` and then use `""`.

packages/storage/src/persisted.ts
atk: Thanks!

packages/storage/src/cookies.ts
BierDav: yeah do that, it would be a bit weird if in the response header randomly a `$1` appears, because as far as i can see this is never replaced against anything different

atk: $1 will be replaced with the first capture, in this case an empty string or ", ".

BierDav: Ahh ok, didn't know that. If it is so than everything should be fine

packages/storage/src/cookies.ts
atk: I completely rewrote the server read/write part and even changed the API so we now get the headers for the response as well (since we need it for reading).

packages/storage/src/cookies.ts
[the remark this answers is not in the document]
atk: I now read the request headers and then the response headers; the last value will overwrite previous ones, which should coincide with the behavior of most servers (unless you forcefully implement to do otherwise).

packages/storage/src/cookies.ts
atk: It was still wasting CPU cycles, so I've fixed it.

thetarnav:
It looks good to me. It's a bigger pr so just merge it when you feel like it's time to release.

PR #590, opened 2024-03-24
Make `props` argument optional

duckradu:
The signature for the `Directive` doesn't allow it to be used as a ref as shown in the docs. If the props are optional then the function signature matches.

image

### DIFF
--- a/packages/pagination/src/index.ts
+++ b/packages/pagination/src/index.ts
@@ -268,26 +268,29 @@ export type _E = JSX.Element;
  * @method `end` is a boolean indicator for end of the page
  * @method `setEnd` allows to manually change the end
  */
-export function createInfiniteScroll<T>(fetcher: (page: number) => Promise<T[]>): [
+export function createInfiniteScroll<T, C extends number | string>(
+  fetcher: (page: C) => Promise<T[]>,
+  getNextCursorValue: (pages: Accessor<T[]>, cursorValue?: C) => C,
+): [
   pages: Accessor<T[]>,
   loader: (el: Element) => void,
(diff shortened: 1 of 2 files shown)


thetarnav:
I don’t think you’re using the type correctly.
`ref` is not a directive.

duckradu:
I was following the example from the docs - https://primitives.solidjs.community/package/pagination#createinfinitescroll

thetarnav:
ah nvm I get it
I have to think about it
Making it optional hurts the proper usage

thetarnav:
this? `use:infiniteScrollLoader`?

duckradu:
The docs show that it can be used both as a directive and ref:

image

(feel free to dm me on discord if it's easier to chat there)

thetarnav:
It looks like it shouldn’t be typed as a directive if it is meant to be used with both ref and as a directive. because it just doesn’t use the second arg at all.

duckradu:
Would a new custom type be okay in that case? Something like `ElementInstance` or idk

thetarnav:
no need to make it a generic helper. `createInfiniteScroll` just should be typed correctly.

duckradu:
Something like this? 

image

thetarnav:
yes although there is no need in making it generic
I think `loader: (el: Element) => void` would be enough

duckradu:
Done! Want me to update this PR or should I just close it?

thetarnav:
It seems that this PR is about something completely different now?
Can you make a separate PR for the changes that are changing the API, and leave only the type-fix here?

PR #591, opened 2024-03-24
Make i18n translator resolver full generic

BierDav:
Make a few types more specific especially Scopes. Now we get proper types for available scopes and added support for nested scopes such as for example `app.test`


packages/i18n/src/index.ts
```diff
@@ -186,39 +190,39 @@ export const resolveTemplate: TemplateResolver = (string: string, args?: BaseTem
...
  * Template resolver that does nothing. It's used as a fallback when no template resolver is provided.
  */
-export const identityResolveTemplate = (v => v) as TemplateResolver;
+export const identityResolveTemplate = (v => v) as TemplateResolver<any, any>;
 
-export type Resolved<T, O> = T extends (...args: any[]) => infer R ? R : T extends O ? O : T;
+export type Resolved<Dict, Out> = Dict extends (...args: any[]) => infer R ? R : Out;
 
-export type ResolveArgs<T, O> = T extends (...args: infer A) => any
+export type ResolveArgs<Dict, Out, Args> = Dict extends (...args: infer A) => any
```
thetarnav, line 197: This should have a default value for `Args`, right?

packages/i18n/src/index.ts
```diff
@@ -241,41 +245,49 @@ export type NullableTranslator<T extends BaseRecordDict, O = string> = <K extend
...
+): Translator<Dict, Out, Args> {
+  return (path, ...args) => {
     const value = dict()?.[path];
-
     switch (typeof value) {
       case "function":
         return value(...args);
       case "string":
-        return resolveTemplate(value, args[0]);
+        return resolveTemplate(value, ...args);
```
thetarnav, line 266: Why? template args can only be a single object

thetarnav (changes requested):
There are too many changes to the scope/chained/proxy functionality.
Please just limit it to adding the generic for parameters when necessary.

packages/i18n/src/index.ts
```diff
@@ -137,14 +137,17 @@ export const prefix: <T extends BaseRecordDict, P extends string>(
...
+type DefaultResolveArgs = string | number | boolean;
+type DefaultResolved = string
+
+export type BaseTemplateArgs<Args = DefaultResolveArgs> = Record<string, Args>;
 
 /**
  * A string branded with arguments needed to resolve the template.
  */
-export type Template<T extends BaseTemplateArgs> = string & { __template_args: T };
+export type Template<T extends BaseTemplateArgs<Args>, Args> = string & { __template_args: T };
```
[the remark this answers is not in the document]
BierDav: Not sure what you mean by that, but I have removed the Args anyways because it doesn't really make sense to introduce a generic. The reson is that the type of the resolver cannot really be transferred to the template call. See my comment on the `template` function:

> Keep in mind that you may only use arguments of a type that is supported by your template resolver

If the arguments used by the template are not supported by the resolver than I have introduced this sketchy safety machanism:

```
export type ResolveArgs<Value, Args> = Value extends (...args: infer A) => any
  ? A
  : Value extends Template<infer R>
    ? (R extends BaseTemplateArgs<Args> ? [args: R] : [`Translator resolver doesn't fully support the argument types used in this template`])
    : Value extends BaseRecordDict ?
      [] :
      [args?: BaseTemplateArgs<Args>];
```

I also don't like it a lot, is everything else than pretty, but it works reliable and does it's job. The other thing we could have done is to just ignore those entries by returning an empty array, but I think that confuses developers, because there is no way one can conclude out of the fact that the template translation key is not available, that they have used an unsupported argument.

packages/i18n/src/index.ts
BierDav: I can definetly add it yes

packages/i18n/src/index.ts
BierDav: You're right, it was just the easiest way to make the types work. I have reverted it and annotated it with ts-expect-error, because casting this would be really messy

BierDav:
> There are too many changes to the scope/chained/proxy functionality. Please just limit it to adding the generic for parameters when necessary.

I know, that's because of the now nicely working scope api. To get all the types right was really a challange and involved hours of hard work, so please can we also merge this 👉👈

thetarnav:
@BierDav  can you fix this as well? :)

thetarnav:
Everything seems to be working as before. on solid-site too.
I added a little test for the generic resolver types.
Could you add tests for the improved functionality of scopes?

BierDav:
Sorry oversaw the pr. I've added a test that should ensure that nested translation works, the other part with the types could also be checked by typechecking the test files, don't know if that is already done or not.

marvin-j97:
Is there anything else needed for this feature? The templates are pretty limited as of now, compared to some other i18n libraries where you can e.g. pluralization in the templates, like: `no apples | one apple | {count} apples` (vue i18n)

BierDav:
This is not supported out of the box, but you can implement your own resolver which would allow you to support that syntax

BierDav:
@thetarnav is this ready to merge?

PR #598, opened 2024-03-31
Bugfix for newly introduced sync api

BierDav:
Fix that channel wasn't even used in messageSync
Remove location origin, because it is unnecessary and on the other hand Broadcast Channel doesn't have this parameter

@atk can you take a look and merge? It's just a small fix, but neccessary for usage sync api usage with Broadcast Channel.


BierDav:
Just wanted to add a note: The current implementation with window messages is quite bad if you already use window messages for other things. Currently we can't determine who the sender of the message was. This is a problem because if we send any message it will get processed which immediately results in errors, whether it is because the message data was not a JSON or because it is an arbitrary JSON object

atk (approved):

atk:
Please add a changeset: `npx changeset`.

atk:
Sorry, I should have been more precise; the changeset needs to be added on the root path, otherwise it won't work.

BierDav:
> Sorry, I should have been more precise; the changeset needs to be added on the root path, otherwise it won't work.

what do you exactly mean by that?

atk:
You ran the changeset command from a package directory.

atk (approved):

Issue #599, opened 2024-04-03
messageSync fails with invalid JSON when present with other librairies


jean343:
### Describe the bug

I am created a persisted signal with `messageSync()` and setting it this way.

```JavaScript
  return makePersisted(createSignal(value, options), {
    sync: isServer ? undefined : messageSync(),
    storage: cookieStorage.withOptions!({ path: "/", sameSite: "Lax", secure: true }),
  });
```

The problem with `messageSync` is that it expects any messages on the domain to be a JSON string. In our app, we are getting JSON objects from `"react-devtools-content-script"`. I would expect `messageSync` to accept any types of messages without throwing exceptions.

![image](https://github.com/solidjs-community/solid-primitives/assets/7070881/8d6c2002-094b-4ee4-b884-34762de3701a)

PR #602, opened 2024-04-03
storage: safe sync deserialization

atk:
addresses https://github.com/solidjs-community/solid-primitives/issues/599

### DIFF
--- a/packages/storage/src/persisted.ts
+++ b/packages/storage/src/persisted.ts
@@ -1,6 +1,6 @@
 import type { Accessor, Setter, Signal } from "solid-js";
 import { createUniqueId, untrack } from "solid-js";
-import { isServer } from "solid-js/web";
+import { isServer, isDev } from "solid-js/web";
 import type { SetStoreFunction, Store } from "solid-js/store";
 import { reconcile } from "solid-js/store";
 import type { AsyncStorage, AsyncStorageWithOptions, StorageWithOptions } from "./types.js";
@@ -176,13 +176,17 @@ export function makePersisted<T, O extends Record<string, any> = {}>(
         try {
           const value = deserialize(data);
           (signal[1] as any)(() => value);
-        } catch (e) { }
+        } catch (e) {
+          if (isDev) console.warn(e);
+        }
       }
       : (data: string) => {
         try {
           const value = deserialize(data);
           (signal[1] as any)(reconcile(value));
-        } catch (e) { }
+        } catch (e) {
+          if (isDev) console.warn(e);
+        }
       };
   let unchanged = true;
 
@@ -249,7 +253,9 @@ export const messageSync = (channel: Window | BroadcastChannel = window): Persis
     channel.addEventListener("message", ev => {
       try {
         subscriber(JSON.parse((ev as MessageEvent).data));
-      } catch (e) { }
+      } catch (e) {
+        if (isDev) console.warn(e);
+      }
     }),
   (key, newValue) =>
     channel.postMessage(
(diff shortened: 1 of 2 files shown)


packages/storage/src/persisted.ts
```diff
@@ -172,8 +172,18 @@ export function makePersisted<T, O extends Record<string, any> = {}>(
   const init = storage.getItem(name, storageOptions);
   const set =
     typeof signal[0] === "function"
-      ? (data: string) => (signal[1] as any)(() => deserialize(data))
-      : (data: string) => (signal[1] as any)(reconcile(deserialize(data)));
+      ? (data: string) => {
+        try {
+          const value = deserialize(data);
+          (signal[1] as any)(() => value);
+        } catch (e) { }
```
thetarnav, line 179: Shouldn't catching errors be the job of the (de)serializer?
Instead of try catching it here, I would modify the default JSON serializer to prevent throwing.

```ts
const default_serialize: (data: any) => string = 
	data => {
		try {
			return JSON.stringify(data)
		} catch (e) {
			if (isDev) console.warn(e)
		}
	}
	
const default_deserialize: (data: string) => any = 
	data => {
		try {
			return JSON.parse(data)
		} catch (e) {
			if (isDev) console.warn(e)
		}
	}

// in makePersisted
const serialize: (data: T) => string = options.serialize || default_serialize
const deserialize: (data: string) => T = options.deserialize || default_deserialize
```

atk: I considered it, but this would mean you'd have to repeat the code for every serializer. For the sake of minimal bundle size, I decided against it.

thetarnav: only for the serializes that can themselves throw, right?
but fine, if that's the trade-off you're choosing
but can you add some kind of dev-only warning like I did?
Right now you will have no idea that serialization failed silently.

atk: In case of web socket or broadcast channel or pusher, it might receive data that fails to deserialize because it is unrelated... but a warning doesn't hurt, so that's fine.

thetarnav (approved):

packages/storage/src/persisted.ts
jean343: I would prefer not having the warning, the library should only warn if it fails at decoding messages sent by itself. It would be very confusing if we keep receiving warnings that third party messages aren't decoded properly as they should not be decoded in the first place.

atk: Since we cannot know in advance if a message is sent by ourselves, this requirement is a bit problematic.

jean343: Why do we need to serialize the messages to string, I have patched it this way and it worked well for me.
```JavaScript
            (subscriber) =>
              window.addEventListener("message", (ev) => {
                if (ev.data.type === "createCookieSignal") subscriber(ev.data);
              }),
            (key, newValue) => {
              postMessage({ key, newValue, type: "createCookieSignal", url: location.href });
            },
```

jean343: If I apply the current patch as-is, I get hundreds of logs per minute

image

Those messages are caused by a bug in the React DevTools Chrome extension, but they could very well be legitimate.
https://github.com/facebook/react/issues/27992

This library can't warn on non-JSON messages.

atk: Ah, the original postMessage API only supported string; I see this shortcoming was fixed in a later version. Since that is the case, we can certainly drop the serialization for post message / broadcast channel sync.

jean343 (approved):
Awesome, thank you!!

PR #603, opened 2024-04-04
Fix dev dark mode color

2hu12:
The text color of badges would be black under dark mode in local dev
image


thetarnav:
Thanks, I think the issue has been already fixed in the meantime.

PR #604, opened 2024-04-04
Make messageSync ssr compatible

BierDav:
Sorry in my last pr i have missed a few things, this shoud now patch this. 
Also now I have made only a patch bump and not a minor one as I did last time.


atk:
Sorry, this was already fixed in another PR.

PR #605, opened 2024-04-04
Fix script loading issue in Firefox fixes #601

### LINKED ISSUE #601: createScriptLoader's onLoad fails from time to time with FF
### Describe the bug

In Firefox-based browsers, the onload callback of createScriptLoader is not fired systematically.

It works perfectly in Chrome, but it fails in Firefox.

I've set up a reproduction using Cypress, which allows testing the behavior in different browsers.

Instructions are in the README.

### Minimal Reproduction Link

https://github.com/jchatard/script-loader-ff

jchatard:
In Firefox the onload attribute is not always triggered on script element so we instead use script.addEventListener('load')

### DIFF
--- a/packages/script-loader/src/index.ts
+++ b/packages/script-loader/src/index.ts
@@ -8,7 +8,7 @@ export type ScriptProps = Omit<ComponentProps<"script">, "src" | "textContent"> 
   [dataAttribute: `data-${string}`]: any;
 };
 
-const OMITTED_PROPS = ["src"] as const;
+const OMITTED_PROPS = ["src", "onload", "onLoad"] as const;
 
 /**
  * Creates a convenient script loader utility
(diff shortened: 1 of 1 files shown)


atk:
This loses us the ability to use `[handler, argument]` event props. We also seem to have some issues with the calling of spread() that need to be addressed. I would probably go and replace it. In the meantime, can you try on:load as a workaround?

jchatard:
I'm not sure I fully understand what you suggest with `on:load`.

My understanding is that `on:` is for JSX markup.

But neither in this PR or in my user land code, do I have JSX involved.

My user land code:

```ts
  createScriptLoader({
    src: "/assets/glide.min.js",
    async: true,
    onLoad() {
      setHasGlideLoaded!(true);
    },
  });
```

What do you suggest exactly?

atk:
Use `'on:load': () => ...`.

atk:
I just checked and 'on:load' seems to work flawlessly, so we can pull off the following trick:

```ts
const OMITTED_PROPS = ["src", "onLoad", "onload"] as const;
//...
  const script = document.createElement("script");
  const [local, scriptProps] = splitProps(props, OMITTED_PROPS);
  const onload = local.onload || local.onLoad ? { 'on:load': local.onload || local.onLoad } : {};
  setTimeout(() => spread(script, mergeProps(scriptProps, onload), false, true));
```

atk:
But I'm afraid onError might have the same issue?

jchatard:
I hope to have time to test this today.

atk:
I just checked, on:load is transpiled by JSX expressions, so that does not help, we need to set the events ourselves. Let's use the actual events, but for all events. I'll create a PR.

atk:
Closed; being fixed in #637

Issue #600, opened 2024-04-03
storage.withOptions will convert the key to an array and create infinite cookies


jean343:
### Describe the bug

This is my `makePersisted` code. I am using `withOptions` to set the path to our cookies.

```JavaScript
  return makePersisted(createSignal(value, options), {
    storage: cookieStorage.withOptions!({ path: "/", sameSite: "Lax", secure: true }),
    deserialize: (data) => {
      try {
        return JSON.parse(data);
      } catch (e) {
        return value;
      }
    },
  });
```

Problem is that the key which is initially a string gets transformed to an array.

![image](https://github.com/solidjs-community/solid-primitives/assets/7070881/fb0d5698-e200-423f-a236-0c219e24d07e)

![image](https://github.com/solidjs-community/solid-primitives/assets/7070881/9001db73-e7fc-4185-a544-072ef42630fc)

![image](https://github.com/solidjs-community/solid-primitives/assets/7070881/d43c342d-5003-47f5-ae00-f56b3caf9c96)

And, finally, we get unlimited cookies

![image](https://github.com/solidjs-community/solid-primitives/assets/7070881/abae3465-2d5c-40d7-8a1e-0a8dafd4c4f9)

Sorry for not creating a reproduction link, it's difficult to create/handle cookies in there.

jean343:
Additionally, I have tried patching the library to add an `apply`, as such:
```JavaScript
storage[key].apply(null, [...args, options]);
```

It sends the values correctly to `cookieStorage`, however, `storageOptions` is always the last parameter, and `options` is ignored by the cookie storage methods.

In the mean time, I am passing path to `storageOptions` instead of `withOptions`.

atk:
Thank you for the feedback, I'll have a look.

atk:
I think this fix should work:

```ts
          (wrapped as any)[key] = (...args: Parameters<(typeof storage)[typeof key]>) => {
            const argsWithOptions = args.slice(0);
            argsWithOptions[storage[key].length - 1] = options;
            storage[key](argsWithOptions);
          }
```

atk:
I still need to test it, so I will submit a PR later in the evening.

atk:
there were a few errors, but I've submitted a PR.

PR #606, opened 2024-04-04
storage: fix withOptions

atk:


jean343:
This would fix #600

thetarnav (approved):

jean343:
@atk there is a missing `return` in this PR. Should have been:
```JavaScript
      if (typeof storage[key] === "function") {
        wrapped[key] = (...args) => {
          args[storage[key].length - 1] = options;
         return storage[key](...args);
        };
      }
```

PR #611, opened 2024-04-09
Patch Media: Remove event-listenever package and replaced with local function

davedbase:
This patch is designed to support older iOS/MacOS Safari browsers that don't support addEventListener on MediaQuery. Since it only supports addListener, we have to add a fallback. The PR changes two details:

1. Removes the `event-listener` package
2. Adds a `makeListener` function replacing `makeEventListener` that uses `addListener` as a callback
3. Cleans up

This patch has passed tests however I haven't not been able to verify that it is fully functional.

### DIFF
--- a/packages/media/src/index.ts
+++ b/packages/media/src/index.ts
@@ -1,42 +1,12 @@
 import { Accessor } from "solid-js";
 import { isServer } from "solid-js/web";
-import { entries, noop, createHydratableSignal, tryOnCleanup } from "@solid-primitives/utils";
+import { makeEventListener } from "@solid-primitives/event-listener";
+import { entries, noop, createHydratableSignal } from "@solid-primitives/utils";
 import { createHydratableStaticStore } from "@solid-primitives/static-store";
 import { createHydratableSingletonRoot } from "@solid-primitives/rootless";
 
 /**
- * Make listener helper to support binding events specifically for MQ.
- * This function doesn't use the standard makeEventListener as iOS 13
- * did not support addEventListener, hence this function checks for it
- * before binding addListener instead.
- * @param mq - ref to MediaQueryList
- * @param type - name of the handled event
- * @param handler - event handler
- * @param options - addEventListener options
- * @returns Function clearing all event listeners form targets
- */
-type EventHandler<T> = (event: T) => void;
-export function makeListener(
-  mq: MediaQueryList,
-  type: string,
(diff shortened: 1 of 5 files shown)


packages/media/src/index.ts
```diff
@@ -1,12 +1,42 @@
...
+  if ('addEventListener' in mq) {
+    mq.addEventListener('change', handler);
+    rm = mq.removeEventListener.bind(mq, type, handler as EventHandler<Event>, options);
+  } else {
+    // @ts-expect-error deprecated API
+    mq.addListener(handler);
+    // @ts-expect-error deprecated API
+    rm = mq.removeListener.bind(mq, handler, options);
+  }
+  return tryOnCleanup(rm);
```
thetarnav, line 35: do we want to add onCleanup to this?
I know that some of the make__ apis, call onCleanup, but I think it's a mistake, because it breaks the vanilla js vs reactive separation.
but I'm not sure
definitely would keep onCleanup in `makeMediaQueryListener` to not break existing behaviors

packages/media/src/index.ts
```diff
@@ -1,12 +1,42 @@
...
+ * @param mq - ref to MediaQueryList
+ * @param type - name of the handled event
+ * @param handler - event handler
+ * @param options - addEventListener options
+ * @returns Function clearing all event listeners form targets
+ */
+type EventHandler<T> = (event: T) => void;
+export function makeListener(
+  mq: MediaQueryList,
+  type: string,
```
thetarnav, line 21: Is there even a point in passing `type`? Wouldn't you just want `"change"`? If you pass something else, then you cannot call `addListener`.
Actually, in the body you are adding listener with `"change"` hardcoded and removing it with `type` param.

thetarnav (commented):
Is this media query specific or does other web apis have this issue as well?
Also, what browsers do we even want to support?
We could say in the documentation that we require specific dom features, and ask users to polyfill themselves, instead of polyfilling ourselves.
Users can simply do this to support the browsers youve mentioned, and keep using this package:
```js
MediaQueryList.prototype.addEventListener = function(callback) {
	this.addListener(callback)
}
MediaQueryList.prototype.removeEventListener = function(callback) {
	this.removeListener(callback)
}
```

thetarnav:
All good
Just use `pnpm changeset` or [this](https://github.com/solidjs-community/solid-primitives/pull/611#issuecomment-2045585486) next time :)

PR #616, opened 2024-04-22
Fix `trackStore` tracking multiple same-object references. (Fixes #582)

### LINKED ISSUE #582: `trackStore()` stops working in an edge case
### Describe the bug

If you open link and click the button twice, you will see that `trackStore` log statement stops logging. If you comment out, the `modifyMutable()` call, you will notice this goes away.

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/e52dc215-b5f1-4b10-a151-7134b56168e3

thetarnav:
Seems like I made an error when writing a test.
`set("a", "a.a", "a.a.b", "thoughts")`
instead of
`set("a", "a.a", "a", "a.b", "thoughts")`
which caused me to make a flawed implementation to satisfy the test 🫠
seems to be good now

### DIFF
--- a/.changeset/beige-suits-add.md
+++ b/.changeset/beige-suits-add.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/deep": patch
+---
+
+Fix `trackStore` tracking multiple same-object references. (Fixes #582)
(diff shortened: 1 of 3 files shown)


astoilkov:
Nice! It works!

PR #619, opened 2024-04-26
Remove Directive type from createInfiniteScroll

onx2:
I'm getting an error in a project of mine which has a custom component. This component accepts a `ref` essentially as `ref?: Ref<HTMLSpanElement>` _(other details omitted for brevity)_.

When consuming the component and passing in the loadingRef from this primitive I get this error

image

If I'm understanding the intent of the `Directive` return type here, it really is just a ref - this is correct? If this approach isn't preferred, does anyone have a suggestion on how I can accommodate for `Directive` and `Ref`?


thetarnav:
Don’t use Ref, but a maybe a better type could be inlined there instead.

onx2:
`Ref` is the type exposed from solid for ref, I'm not following what you mean. Do you mind elaborating?

thetarnav:
forgot solid now exports `Ref` lol
but I mean this: `(el: Element) => void`
Instead of using aliases like `Directive` or `Ref`

onx2:
I see - yes that seems to work though I no longer give additional constraints for the type:

```ts
export type TypographyProps = ParentProps & {
  color?: keyof Theme["colors"];
  spacing?: Theme["spacing"];
  wrap?: boolean;
  variant?: "heading" | "subheading" | "title" | "subtitle" | "body";
  ref?: Element | ((el: Element) => void);
};
```

**Before**
```ts
/**
 * Using these extra types allows the generic type passed into `Ref` to be narrowed
 * based on the variant that the consumer passes in.
 */
type HeadingVariantProps = {
  variant: "heading" | "subheading" | "title" | "subtitle";
  ref?: Ref<HTMLHeadingElement>;
};

type BodyVariantProps = {
  variant: "body";
  ref?: Ref<HTMLParagraphElement>;
};

type DefaultVariantProps = {
  variant?: undefined;
  ref?: Ref<HTMLSpanElement>;
};

export type TypographyProps = ParentProps & {
  color?: keyof Theme["colors"];
  spacing?: Theme["spacing"];
  wrap?: boolean;
} & (HeadingVariantProps | BodyVariantProps | DefaultVariantProps);
```

thetarnav:
I mean't that `createInfiniteScroll` should return `(el: Element) => void` instead of `Directive` or `Ref` as it is neither a directive nor a ref, it should just be able to be used with those. Using `Ref` in your component is fine, thats what it was made for.

onx2:
> I meant that `createInfiniteScroll` should return `(el: Element) => void` instead of `Directive` or `Ref` as it is neither a directive nor a ref, it should just be able to be used with those. Using `Ref` in your component is fine, thats what it was made for.

Understood 👍🏼  I can update this PR to make that change. Thank you for talking through this with me.

onx2:
@thetarnav Does this look okay now to merge?

atk (approved):

PR #622, opened 2024-04-26
storage: type fixes

atk:
The following code led to type errors:

```ts
const [x, setX] = makePersisted(createSignal(""), { storage: cookieStorage, name: "test" })
```

These type errors are now fixed and the types are even inexplicably simplified. Cheers!


thetarnav:
ya broke the build tho!

Issue #530, opened 2023-10-20
[storage] `cookieStorage`'s `maxAge` option doesn't work


yume-chan:
### Describe the bug

Create a persisted signal using `cookieStorage` and passing the `maxAge` option, but the created cookie is a session cookie.

In the linked StackBlitz project, click "Open in New Tab" button at top-right corner. Looks like it can't set cookies from an iframe.

```ts
  const [maxAge, setMaxAge] = makePersisted(createSignal('initial'), {
    name: 'maxAge',
    storage: cookieStorage,
    storageOptions: { path: '/', maxAge: 60 * 60 * 24 * 7 },
  });
  const [expires, setExpires] = makePersisted(createSignal('initial'), {
    name: 'expires',
    storage: cookieStorage,
    storageOptions: {
      path: '/',
      expires: new Date(Date.now() + 60 * 60 * 24 * 7 * 1000),
    },
  });

  createEffect(() => {
    setMaxAge('hello');
    setExpires('hello');
  });
```

![image](https://github.com/solidjs-community/solid-primitives/assets/1330321/628d7322-ecbd-42ab-8c44-d4b6942d00a8)

### Minimal Reproduction Link

https://stackblitz.com/edit/github-n4ho5j-nffe5j?file=src%2Froutes%2Findex.tsx

sh1man999:
I confirm this error

apatrida:
The problem is `serializeCookieOptions` doesn't convert `maxAge` to `max-age` and other similarly named object properties that do not match the name of the attribute expected by the browser.

PR #625, opened 2024-04-29
storage: fix cookie serialization

atk:
FIxes https://github.com/solidjs-community/solid-primitives/issues/530

### DIFF
--- a/packages/storage/src/cookies.ts
+++ b/packages/storage/src/cookies.ts
@@ -29,22 +29,15 @@ const cookiePropertyKeys = [
 ] as const;
 
 function serializeCookieOptions(options?: CookieOptions) {
-  if (!options) {
-    return "";
-  }
-  let memo = "";
-  for (const key in options) {
-    if (!cookiePropertyKeys.includes(key as keyof CookieProperties)) continue;
-
-    const value = options[key as keyof CookieProperties];
-    memo +=
-      value instanceof Date
-        ? `; ${key}=${value.toUTCString()}`
-        : typeof value === "boolean"
-          ? `; ${key}`
-          : `; ${key}=${value}`;
-  }
-  return memo;
+  if (!options) return "";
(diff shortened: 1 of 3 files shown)


packages/storage/src/cookies.ts
```diff
@@ -29,22 +29,15 @@ const cookiePropertyKeys = [
...
-      value instanceof Date
-        ? `; ${key}=${value.toUTCString()}`
-        : typeof value === "boolean"
-          ? `; ${key}`
-          : `; ${key}=${value}`;
-  }
-  return memo;
+  if (!options) return "";
+  return Object.entries(options).map(([key, value]) => {
+    key = key.replace(/([a-z])([A-Z])/g, (_, a, b) => `${a}-${b.toLowerCase()}`);
```
apatrida, line 34: This breaks httpOnly and sameSite? which is not hyphenated?

I think only max-age is hyphenated

packages/storage/test/cookies.test.ts
```diff
@@ -12,6 +12,12 @@ describe("cookieStorage", () => {
     cookieStorage.removeItem("test");
     expect(cookieStorage.getItem("test")).toBe(null);
   });
+
+  it("serializes options correctly", () => {
+    const set = vi.spyOn(document, "cookie", "set");
+    cookieStorage.setItem("test3", "good", { domain: "https://localhost:3000", path: "/", httpOnly: true, maxAge: 60 * 60 * 24 * 7 });
+    expect(set).toHaveBeenCalledWith("test3=good; domain=https://localhost:3000; path=/; http-only; max-age=604800");
```
apatrida, line 19: invalid test.  `httpOnly` is not `http-only`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

atk: Fixed.

PR #629, opened 2024-05-12
Add createSpring primitive.

Blankeos:
## Summary
- This PR adds the `createSpring` primitive. Inspired and Forked from https://svelte.dev/docs/svelte-motion#spring
- The API works exactly like createSignal, just that the setter will actually interpolate the value.
- It also has familiar opts as the one with svelte-motion/spring.
- I come from using Svelte and it's one of the most essential primitives I use for animation. I've been using Solid more and more and have come to love it. I think this primitive is one of the few missing for me here. I think Solid would greatly benefit from a primitive like this since we already have a `tween` primitive.

Example Usage:
```ts
export default function SpringyPage() {
  const [progress, setProgress] = createSpring(0);

  const [radialProgress, setRadialProgress] = createSpring(0, {
    stiffness: 0.05,
  });

  const [xy, setXY] = createSpring(
    { x: 50, y: 50 },
    { stiffness: 0.08, damping: 0.2, precision: 0.01 }
  );

  function toggleProgress() {
    if (progress() === 0) setProgress(1);
    else setProgress(0);
  }
  // ...
```

Witness, the springiness:

https://github.com/solidjs-community/solid-primitives/assets/38070918/7c4fa01f-7959-4a67-9588-e28448f7f20d

https://github.com/solidjs-community/solid-primitives/assets/38070918/4dcf9e6f-7a2e-44f8-b15a-d29d6551f561

Todos:
- [x] Tests
- [x] README.md
- [x] package.json


packages/spring/src/index.ts
rcoopr, line 1: Currently it allows you to pass a string as an input (or an array of strings, object with string values etc), but you could use a type like this to provide some validation

```ts
type SpringTargetPrimitive = number | Date;
type SpringTarget =
  | SpringTargetPrimitive
  | { [key: string]: SpringTargetPrimitive | SpringTarget }
  | SpringTargetPrimitive[]
  | SpringTarget[];
```

Blankeos:
Update: I also figure of adding `createSpringDerived` by the way, so that there's a similar API that works like the current `createTween`'s API which doesn't return a setter, but takes in an Signal accessor.

```tsx
const springValue = createTweenDerived(signal, opts)
```

Will add this tonight.

packages/spring/src/index.ts
Blankeos: While implementing this, I ran into a bit of a hiccup. Didn't know generics behaved this way lol. They seem to convert the data primitives into constants. (e.g. passing `0` will make the type `0`, `1` will be `1`, etc.)

image

Luckily the fix was easy:

image

packages/spring/test/index.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    let setSpringed!: (
+      newValue: number,
+      opts?: { hard?: boolean; soft?: boolean },
+    ) => Promise<void>;
+    let springed!: () => number;
+
+    createRoot(d => {
+      dispose = d;
+      [springed, setSpringed] = createSpring(0);
+    });
```
thetarnav, line 33: You can just return from `createRoot`
```suggestion
	const {dispose, springed, setSpringed} = createRoot(dispose => {
		const [springed, setSpringed] = createSpring(0)
		return {dispose, springed, setSpringed}
	})
```

packages/spring/test/index.test.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    expect(springed()).toBe(0);
+    setSpringed(50); // Set to 100 here.
+
+    // Not sure if this will be erratic.
+    await new Promise(resolve => setTimeout(resolve, 300));
+
```
thetarnav, line 40: Yeah timeouts in tests are my worst nightmare. If you can mock the animation frame and test it without using timeout it would be much better.
The tween package does that so possibly it could be copied: https://github.com/solidjs-community/solid-primitives/blob/main/packages/tween/test/index.test.ts

packages/spring/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "url": "git+https://github.com/solidjs-community/solid-primitives.git"
+  },
+  "bugs": {
+    "url": "https://github.com/solidjs-community/solid-primitives/issues"
+  },
+  "primitive": {
+    "name": "spring",
+    "stage": 0,
+    "list": [
+      "createTween"
```
thetarnav, line 20:
```suggestion
      "createSpring",
      "createDerivedSpring"
```

packages/spring/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+
+type TaskEntry = { c: TaskCallback; f: () => void };
+
+// src/internal/client/timing.js
+
+/** SSR-safe RAF function. */
+const request_animation_frame = isServer ? () => {} : requestAnimationFrame;
+
+/** SSR-safe now getter. */
+const now = isServer ? () => Date.now() : () => performance.now();
```
thetarnav, line 40: Isn't both `Date.now` and `performance.now` available in both node and browser?

thetarnav (changes requested):
Thanks, this is a very nice addition.

packages/spring/test/index.test.ts
```diff
@@ -0,0 +1,164 @@
+import { describe, expect, it } from "vitest";
+import { createRoot, createSignal } from "solid-js";
+import { createDerivedSpring, createSpring } from "../src/index.js";
+
+// ===========================================================================
+// Tests
+// ===========================================================================
```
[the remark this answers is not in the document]
Blankeos: Removing it lol.

packages/spring/src/index.ts
atk: Yes. performance.now() will even get you higher resolution data on the server than on the client.

packages/spring/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+  const [current_token, setCurrentToken] = createSignal<object>();
+
+  const [lastValue, setLastValue] = createSignal<T>(initialValue);
+  const [targetValue, setTargetValue] = createSignal<T | undefined>();
+
+  const [inv_mass, setInvMass] = createSignal<number>(1);
+  const [inv_mass_recovery_rate, setInvMassRecoveryRate] = createSignal<number>(0);
+  const [cancelTask, setCancelTask] = createSignal<boolean>(false);
+
+  const set: SpringSetter<T> = (newValue, opts = {}) => {
```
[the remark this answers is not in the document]
Blankeos: Ok, this is new to me. Thanks! So I should just wrap the `set` function in [`untrack`](https://docs.solidjs.com/reference/reactive-utilities/untrack) yeah?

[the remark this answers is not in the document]
thetarnav: Or just `untrack(springValue)`, since you should only read it once

packages/spring/src/index.ts
Blankeos: Yup, just checked `performance.now()` and `Date.now` are both available in both node and browser. Just mostly copied this from the svelte source code. So I have no idea what should be preferred here:
https://github.com/sveltejs/svelte/blob/main/packages/svelte/src/internal/client/timing.js

In Svelte:
```
const now = BROWSER ? () => performance.now() : () => Date.now();
``` 

---

I'll take your advice on this one though, should I still use a ternary or just stick with one?

packages/spring/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+  const [task, setTask] = createSignal<Task | null>(null);
+
+  const [current_token, setCurrentToken] = createSignal<object>();
+
+  const [lastValue, setLastValue] = createSignal<T>(initialValue);
+  const [targetValue, setTargetValue] = createSignal<T | undefined>();
+
+  const [inv_mass, setInvMass] = createSignal<number>(1);
+  const [inv_mass_recovery_rate, setInvMassRecoveryRate] = createSignal<number>(0);
+  const [cancelTask, setCancelTask] = createSignal<boolean>(false);
```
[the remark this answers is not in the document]
Blankeos: Great catch btw. Thanks! Didn't realize I didn't need to have signals all over the place 😬.

packages/spring/src/index.ts
thetarnav: Just pick one.
Will this code path even hit on the server? After it's inlined to createSpring? The whole primitive should be a normal signal on the server, no need to read current time then.

Blankeos: So I tried `Date.now()` it doesn't work, it broke the spring. I'll stick with `performance.now()`.

packages/spring/test/index.test.ts
Blankeos: Unfortunate news, I can't mock it. I tried:

1. vi.useFakeTimers and vi.advanceTimersByTime
```ts
describe("createSpring", () => {
  it("updates toward target over time", async () => {
    vi.useFakeTimers(); // Use fake timers

    const { dispose, springed, setSpringed } = createRoot(dispose => {
      const [springed, setSpringed] = createSpring(0);
      return { dispose, springed, setSpringed };
    });

    expect(springed()).toBe(0);
    setSpringed(100); // Set target to 100

    // Advance timers and check intermediate states
    vi.advanceTimersByTime(100);
    expect(springed()).toBeGreaterThan(0);
    expect(springed()).toBeLessThan(100);

    vi.advanceTimersByTime(500);
    expect(springed()).toBeGreaterThan(50);
    expect(springed()).toBeLessThan(100);

    vi.advanceTimersByTime(1000);
    expect(springed()).toBe(100); // Should have reached the target

    dispose();
  });
});
``` 

2. Complete copy of the `_flush_raf` function in `tween`
```ts
let _last_id = 0;
let _raf_callbacks_old = new Map<number, FrameRequestCallback>();
let _raf_callbacks_new = new Map<number, FrameRequestCallback>();

function _flush_raf(time: number) {
  _raf_callbacks_old = _raf_callbacks_new;
  _raf_callbacks_new = new Map();
  for (const callback of _raf_callbacks_old.values()) {
    callback(time);
  }
  _raf_callbacks_old.clear();
}

function _mocked_requestAnimationFrame(callback: FrameRequestCallback): number {
  const id = _last_id++;
  _raf_callbacks_new.set(id, callback);
  return id;
}
function _mocked_cancelAnimationFrame(id: number): void {
  _raf_callbacks_new.delete(id);
}

vi.stubGlobal("requestAnimationFrame", _mocked_requestAnimationFrame);
vi.stubGlobal("cancelAnimationFrame", _mocked_cancelAnimationFrame);

afterEach(() => {
  _raf_callbacks_old.clear();
  _raf_callbacks_new.clear();
  _last_id = 0;
});

describe("createSpring", () => {
  it("updates toward target over time", async () => {
    vi.useFakeTimers(); // Use fake timers

    const { dispose, springed, setSpringed } = createRoot(dispose => {
      const [springed, setSpringed] = createSpring(0);
      return { dispose, springed, setSpringed };
    });

    expect(springed()).toBe(0);
    setSpringed(100); // Set target to 100

    // Advance timers and check intermediate states
    _flush_raf(start + 25);
    expect(springed()).toBeGreaterThan(0);
    expect(springed()).toBeLessThan(100);

    _flush_raf(start + 50);
    expect(springed()).toBeGreaterThan(50);
    expect(springed()).toBeLessThan(100);

    _flush_raf(start + 100);
    expect(springed()).toBe(100); // Should have reached the target

    dispose();
  });
});

```

3. Exported `raf` from the implementation and overriding it.
4. Mocking `raf`

To no avail. In any case, I'll just refactor the tests to the best I can for now (based on your revisions). Thanks again!

thetarnav:
I don't necessarly like that `createSpring` by default tries to handle various types of data recursively.
The core idea of a spring is just a simple math equation for tweening numbers, but the source code reads like something much more complex.
I wonder if it would be smart to separate it into more functions rather than one-does-all primitive. Maybe a separate primitive for numbers and objects?
It would be nice to have a vanilla-js `spring(from, to, time, options)` function as well.
Also maybe the primitive could be added to the `tween` package? The use and concepts are similar. Separate package for each tweening strategy would probably be a lot.
I still need to think about that.

packages/spring/src/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  | SpringTargetPrimitive[]
+  | SpringTarget[];
+
+/**
+ * "Widen" Utility Type so that number types are not converted to
+ * literal types when passed to `createSpring`.
+ *
+ * e.g. createSpring(0) returns `0`, not `number`.
+ */
+export type WidenSpringTarget<T> = T extends number ? number : T extends Date ? Date : T;
```
thetarnav, line 127: I understand `T extends number`, but what is `T extends Date` for?

Blankeos: Hmm. Come to think of it, you're right.

`number` sometimes becomes 0 or 1 (depending on what was passed) so this is needed for that.

`Date` doesn't really get inferred as anything else but Date, so no need for that `T extends Date`

Can just do:
`export type WidenSpringTarget<T> = T extends number ? number : T`

packages/spring/src/index.ts
thetarnav: Just a side note but the "validation" will not catch this:
```ts
const [spring, setter] = createSpring({foo: [1,2,3]} as {foo: number[]} | number)
setter(101)
// runtime error: Cannot read properties of undefined (reading '0')
```

thetarnav: And it seems to break with `readonly number[]`

thetarnav: Also interfaces

image

thetarnav: Also in source code there is a `current_value == null` check, which I think is impossible with the validation.

thetarnav:
I tried to inline the tasks thing from svelte to the primitive. The tests pass and it works as it should in my eyes, but could you double check @Blankeos?
Probably will merge after that.

packages/spring/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+      return Promise.resolve();
+    }
+
+    if (opts.soft) {
+      inv_mass_recovery_rate = 1 / (typeof opts.soft === "number" ? opts.soft * 60 : 30);
+      inv_mass = 0; // Infinite mass, unaffected by spring forces.
+    }
+
+    if (raf_id === 0) {
+      time_last = performance.now()
```
rcoopr, line 145: I think it should be something like `Number(document.timeline.currentTime || performance.now())` ala [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame#timestamp)

In cases where `document.timeline.currentTime` is available, the first frame could have a negative `timeDelta`

packages/spring/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    value_target = typeof param === "function" ? param(value_current) : param;
+
+    if (opts.hard || (stiffness >= 1 && damping >= 1)) {
+      cleanup()
+      setSignal(_ => value_current = value_last = value_target);
+      return Promise.resolve();
+    }
+
+    if (opts.soft) {
+      inv_mass_recovery_rate = 1 / (typeof opts.soft === "number" ? opts.soft * 60 : 30);
```
rcoopr, line 140: This hard-codes an assumed 60 fps into the inverse mass recovery feature. On different hardware, `opts.soft` will behave differently (inv_mass will return to 1 more quickly with higher framerate) and more importantly the `velocity` will be incorrect.

I'm not sure the best fix for fps-independence here, but in my own adaptation I have gone for tracking the fps as `Math.max(fps, 1000 / (time - timeLast))` in the `frame` function above

packages/spring/src/index.ts
thetarnav: hmm youre right
I'm consistently getting negative values with `performance.now`.

image

But with `document.timeline.currentTime` it's always zero.
Which makes me think that maybe this would be better:
```ts
requestAnimationFrame(prev_time => {
	function frame(time) {
		let delta = time-prev_time
		prev_time = time
		requestAnimationFrame(frame)
	}
	requestAnimationFrame(frame)
})
```
Since the first iteration will be ran with `0` delta anyway which won't progress the animation at all, so it could be skipped.
Instead of relying on `document.timeline.currentTime`, which may be and may not be there, and falling back to `performance.now` which is wrong.

packages/spring/src/index.ts
rcoopr: Turns out tracking fps this way is very flawed. At best it gives an inaccurate value (consistently getting ~180 fps when I should really be getting 166) and at worst is can go to extreme values and be very inconsistent.

I can look into frame-independent impls unless somebody else wants to jump in with a solution

thetarnav: My intuition would be to keep the recovery rate as a constant, only depending on `opts.soft`.
And use time delta in `inv_mass = Math.min(inv_mass + inv_mass_recovery_rate * time_delta, 1)`
But I'm curious about other directions.
I'll try to add a test for framerate-independence since we are mocking the raf anyway.

rcoopr: I was able to fix the `inv_mass_recovery_rate` to actually recover in the correct amount of time, but couldn't accomplish the same for the `tick` function

The frame-dependent `time_delta` is baked into this implementation in `tick`. It doesn't quite cancel out even though `d` is multipled by `time_delta` in the end either, but I'm not totally sure what is supposed to be going on instead.

https://github.com/pqml/spring/blob/master/lib/Spring.js - this looks promising, it is based on picking a timestep, checking how many of those timesteps have passed since the last frame, and solving for the position that many times

thetarnav (approved):
I think I’m going to merge this in the current state as it’s already useful and matches sveltes version.
The implementation could be improved to be frame-independent in future PRs  as it shouldn’t change the current interface.

Blankeos:
Awesome work guys! Couldn't really check the past week, it's been a very busy month for me. 

Thanks for adding in your insights @thetarnav @rcoopr!  🦾🥳

PR #634, opened 2024-05-22
Fix unable to set Writable Memos twice to the same value, when they changed in between 

Fxxxlei:
Removed the equality check from the inner signal of a writable memo.
The equality check is performed by the outer memo anyway, and previously prevented the outer memo from being reset.


packages/memo/src/index.ts
```diff
@@ -191,7 +191,7 @@ export function createWritableMemo<T>(
   options?: MemoOptions<T | undefined>,
 ): [signal: Accessor<T>, setter: Setter<T>] {
   let combined: Accessor<T> = () => value as T;
-  const [signal, setSignal] = createSignal(value as T),
+  const [signal, setSignal] = createSignal(value as T, { equals: false }),
```
thetarnav, line 194:
```suggestion
  const [signal, setSignal] = createSignal(value as T, EQUALS_FALSE_OPTIONS),
```

thetarnav:
@Fxxxlei could you add a test covering the fixed case?

Fxxxlei:
Got it done.

PR #653, opened 2024-07-06
Handle rejected promises with raceTimeout #635

### LINKED ISSUE #635: raceTimeout does not catch promise
### Describe the bug

When a promise passed to raceTimeout rejects, it is uncaught, even if the returned promise has a catch. There is no .catch on the internal Promise.race.

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/8bcf575a-f246-446c-a461-722740790f2b

maciek50322:
Resolves #635

### DIFF
--- a/packages/promise/test/intex.test.ts
+++ b/packages/promise/test/intex.test.ts
@@ -75,6 +75,18 @@ describe("raceTimeout", () => {
     await vi.advanceTimersByTimeAsync(100);
     expect(done).toBe(reason);
   });
+
+  // test("rejects before timeout", async () => {
+  //   let done: unknown;
+  //   const reason = new Error("raceTimeout rejection reason");
+  //   const rejectetPromise = Promise.reject();  //vitest complains about unhandled promises, even though it is handled later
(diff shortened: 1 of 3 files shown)


packages/promise/test/intex.test.ts
```diff
@@ -75,6 +75,18 @@ describe("raceTimeout", () => {
     await vi.advanceTimersByTimeAsync(100);
     expect(done).toBe(reason);
   });
+
+  // test("rejects before timeout", async () => {
+  //   let done: unknown;
+  //   const reason = new Error("raceTimeout rejection reason");
+  //   const rejectetPromise = Promise.reject();  //vitest complains about unhandled promises, even though it is handled later
```
maciek50322, line 82: There seems to be an issue involving rejected promises with vitest.
If the rejected promise isn't immediately handled with `then` or `catch`, vitest will show errors about unhandled errors after all tests, even though the test with rejected promise handles it later and test passes fully and successfully

thetarnav: this is unfortunate.
tried this locally with the same effect

PR #654, opened 2024-07-07
`createWritableMemo` fix for reassigning previous memo value after computation

### LINKED ISSUE #592: `createWritableMemo()` doesn't work with `undefined`
### Describe the bug

Passing `undefined` to the `createWritableMemo()` setter doesn't update the memo to`undefined`.
```ts
import { createWritableMemo } from "@solid-primitives/memo";

const [count, setCount] = createSignal(1);
const [result, setResult] = createWritableMemo(() => count() * 2);
result(); // => 2
setResult(undefined); // overwrites calculation result
result(); // => 2 ❌ incorrect it should be undefined
```

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/76de3e3d-d0e1-457f-8d50-1cebfead6c1c

### LINKED ISSUE #633: `createWritableMemo()` cannot be set to the same value twice in a row
### Describe the bug

Given distinct values A and B, when the value of the writable memo is set to A, later computes as B, and then is set to A again, the writable memo holds value B, even though it should hold value A.

This error also happens if the second parameter passed to `createWritableMemo()` is A.

The reason for this behaviour is probably because the internal signal that is set already has value A, even though the writable memo has value B. Setting the inner signal to A won't propagate to the writable memo, since the inner signal did not change. 

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/19a210e3-aeb0-40a4-a82c-e7319790a3de

maciek50322:
resolves #592
resolves #633

### DIFF
--- a/packages/memo/src/index.ts
+++ b/packages/memo/src/index.ts
@@ -191,7 +191,7 @@ export function createWritableMemo<T>(
   options?: MemoOptions<T | undefined>,
 ): [signal: Accessor<T>, setter: Setter<T>] {
   let combined: Accessor<T> = () => value as T;
-  const [signal, setSignal] = createSignal(value as T),
+  const [signal, setSignal] = createSignal(value as T, { equals: false }),
     memo = createMemo(
       callbackWith(fn, () => combined()),
       value,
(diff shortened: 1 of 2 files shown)


maciek50322:
ah see it's already done in #634

PR #662, opened 2024-07-23
Attempt to fix vinxi build on netflify (remove fetch package playground)

maciek50322:
The `fetch` package's playground seems to fail vinxi build on netlify, even though locally it builds.
Exception is thrown in prerendering phase of build, error says `The value [object Function] of type "function" cannot be parsed/serialized`, doesn't say where, but can't exactly check it, because when built locally this error isn't thrown.

Besides that the `fetch`'s playground didn't work, because requested assets are not loaded to final build.

So removing the playground to see it the issue was actually there (there might be more, only 1/4 of all packages were built before the error).


atk:
Could you then also remove the link from the README of the package, please? I wanted to deprecate the fetch package in favor of the resource package in any case.

maciek50322:
@atk There is no link to playground in readme, just "TODO". There's one on https://primitives.solidjs.community/package/fetch but that's automatically generated.

atk (approved):

PR #668, opened 2024-07-27
storage: add localforage support

### LINKED ISSUE #666: Please add an example makePersisted using IndexedDB or localforage

My problem is that very large values fail to be saved via `makePersisted` on `localStorage`, 

The docs mention "there is also `localForage`" but there's no code example and the types don't quite match if you try to use it directly?

image

### Suggest A Solution

---

atk:
+ remove outdated scripts

fixes #666


packages/storage/README.md
```diff
@@ -99,6 +99,20 @@ const [state, setState] = makePersisted(createSignal(), {
 });
 ```
 
+#### LocalForage
+
+LocalForage allows to use indexedDB among others to greatly increase the size of what can be stored. Just drop it in as a storage (only supported in the client):
```
sparecycles, line 104: nit: grammar "LocalForage allows to use ..." is a bit off.

```suggestion
LocalForage uses indexedDB (if available) to greatly increase the size of what can be stored. Just drop it in as a storage (only supported in the client):
```

packages/storage/src/types.ts
```diff
@@ -63,7 +63,7 @@ export type AsyncStorageWithOptions<O> = {
   clear?: () => Promise<void> | void;
   getItem: (key: string, options?: O) => Promise<string | null> | string | null;
   getAll?: () => Promise<any>;
-  setItem: (key: string, value: string, options?: O) => Promise<void> | void;
+  setItem: ((key: string, value: string, options?: O) => Promise<void> | void);
   removeItem: (key: string) => Promise<void> | void;
   key: (index: number) => Promise<string | null> | string | null;
   readonly length: Promise<number> | number | undefined;
```
sparecycles, line 69: LocalForage's length property is a [_function_ returning a promise](https://github.com/localForage/localForage/blob/master/typings/localforage.d.ts#L26), not a property.

 (unless there's some adapter logic I'm missing.)

sparecycles (commented):
Thanks for jumping on this so quickly :).  I have two contributions to the discussion:

packages/storage/src/types.ts
atk: We're not using length at all in makePersisted. Also, we use localforage's type directly if available, so there's no issue here.

atk:
Update: after a short discussion on discord, I decided not to delay the planned changes for this primitive any longer. I removed the dependency on localforage again and simpified the types for makePersisted (which only uses getItem, setItem and removeItem) so additional properties would no longer be an issue.

PR #672, opened 2024-08-06
Add createScrollTo function for scrolling

lidarbtc:
With options and example


packages/scroll/README.md
```diff
@@ -124,7 +124,6 @@ Get an `{ x: number, y: number }` object of element/window scroll position.
 
 _PRs Welcome :)_
 
-- `createScrollTo` - A primitive to support scroll to a target
```
atk, line 127: scroll*ing* to a target

atk (approved):

packages/scroll/README.md
lidarbtc: Thanks for the review, does that mean replacing "scroll" with "scrolling"?

atk (approved):
While you're at it, please add a changeset using `npx changeset`.

lidarbtc:
> While you're at it, please add a changeset using `npx changeset`.

Thanks for noticing. I added it and changed "scroll" to "scrolling".

atk (approved):

packages/scroll/src/index.ts
```diff
@@ -106,3 +106,53 @@ export function createScrollPosition(
...
+    if (typeof options === "number") {
+      options = {
+        left: options,
+        top: y,
+      };
+    }
+
+    const { left, top, behavior = "auto" } = options;
+
+    if (currentTarget instanceof Window) {
```
thetarnav, line 152: Whats this check for? both branches do the same thing

packages/scroll/src/index.ts
```diff
@@ -106,3 +106,53 @@ export function createScrollPosition(
...
+      currentTarget.scrollTo({ left, top, behavior });
+    } else {
+      currentTarget.scrollTo({ left, top, behavior });
+    }
```
thetarnav, line 156: I assume it could be simplified to
```ts
currentTarget.scrollTo(typeof options === "number" ? {left: options, top: y} : options)
```

packages/scroll/src/index.ts
```diff
@@ -106,3 +106,53 @@ export function createScrollPosition(
...
+ * // Using with a specific element
+ * const [element, setElement] = createSignal(document.getElementById('myElement'));
+ * const elementScrollTo = createScrollTo(element);
+ * elementScrollTo({ left: 50, behavior: 'smooth' });
+ *
+ * // Using with x and y coordinates
+ * scrollTo(100, 200);
+ */
+export function createScrollTo(
+  target: Accessor<Element | Window | undefined> | Element | Window = window,
```
thetarnav, line 136: default value of `window` could throw on the server. `() => window` should work.

packages/scroll/src/index.ts
```diff
@@ -106,3 +106,53 @@ export function createScrollPosition(
...
+) {
+  return (options: ScrollToOptions | number, y?: number) => {
```
thetarnav, line 138: Since this is just returning a callback, it could be simplified to this to avoid currying
```ts
function scrollTo(target: Accessor<Element | Window | undefined> | Element | Window, options: ScrollToOptions | number, y?: number): void
```
this way it's clear that the utility is not reactive
functions with `create__(source: Accessor<T>)` signature usually scream that they are reactive

thetarnav (changes requested):
I know that the primitive was listed in the ideas to implement. But I wonder whats the actual usecase for this primitive.

- that it works with both signals and element values? I doubt thats usually a problem, and can be solved with `utils.access` or just checking the type.
- that it is a noop on the server? since this is an action-something what you call after user interaction or on mount, it shouldn't ever run on a server, and if it does it's a bug and it probably should error then.
- it doesn't really leverage the reactivity in any way

Why yould you use this instead of just calling `el.scrollTo`, it's not much more code and it will be clearer. in this current form I don't really see it.

Maybe this would make more sense?
```tsx
function createScrollTo(el: Accessor<Element | Window | null> | Element | Window, options: ScrollToOptions): void

let el
createScrollTo(() => el) // scroll to ref on mount
<div ref={el} />


createScrollTo(signal, {top: 50, behavior: "smooth"}) // scroll to a signal value whenever it changes
```

It would actually laverage solid effects to call scrollTo when the component is mounted.

Although I'm still not sure when would I use this or reach for a library if really needed.

thetarnav:
Now that I think about it

> - `createScrollTo` - A primitive to support scroll to a target

it says "scroll to a target" not "use `el.scrollTo`", since those are different ideas

PR #675, opened 2024-08-11
Attempt to fix vinxi build on netflify part 2

maciek50322:
Continuation of #662
The fix worked, but some next package (i18n) failed due to "out of memory".
Locally the package i18n threw and stopped server when loaded playground, but with different error, that's fixed here.
It might still fix the build if the error that netlify reported was wrong.

Also checked rest of the following packages and fixed it if they stopped dev server.

Not sure if vinxi need to preload all of the playgrounds to start server and throw if single page doesn't work.
Also if the issue is actually "out of memory", disabling preloading might fix it (if that's possible, eg. on dev server it doesn't preload every playground).


thetarnav:
lockfile is outdated?

maciek50322:
> lockfile is outdated?

Yes, removed `solid-app-router`, 
added new lockfile, should be fine now

Issue #673, opened 2024-08-09
`makePersisted` with cookieStorage does not deserialize umlauts on the server correctly 


madaxen86:
### Describe the bug

When persisting a data with umlauts (probably also other special characters) like "öäü" makePersisted does not deserialize those chars correctly but returns instead: "Ã¶Ã¤Ã¼ÃÃÃ"
Happens only during SSR. On client side navigation the values are returned correctly.

``

See GitHub repo for reproduction.
1. ON the index page click on "set umlauts" to set the text.
2. reload the page to trigger SSR deserialization.

### Minimal Reproduction Link

https://github.com/madaxen86/make-persisted-umlauts

atk:
Thank you for the reproduction repo. I'm just back from my vacation, will investigate and fix it this evening (I'm pretty sure we're just missing the encoding in the server part of cookieStorage.

atk:
Interestingly, the cookie is internally utf-8, but the header that the browser sends seems to already have the wrongful encoding.

atk:
Interestingly, this is actual browser behavior. A few RFCs tried to rectify the situation, but still aren't supported by most major browsers. I guess I have to encode the cookie both ways. Give me a bit more time, please.

madaxen86:
No problem, it's not urgent.
As a workaround I am setting/updating the cookie myself and JSON.stringify the value. 
Works perfectly.
But still would be great if it worked out of the box.
I also tried to use the serialize/deserialize props and add do the same JSON.stringify and .parse but without any success.

atk:
Sorry for the delay, my attempts to make \u... escapes failed and the solution became too complex, so I had to pivot to use encodeURIComponent - but then, cookies are a limited resource and I only wanted to escape those characters not supported; long story short, I'm just opening a new PR that fixes the issue.

madaxen86:
No problem.
I figured it would not be easy.
I'll test it when merged and published 🙏

PR #676, opened 2024-08-19
storage: escape utf-8 cookies correctly

atk:
fixes https://github.com/solidjs-community/solid-primitives/issues/673

### DIFF
--- a/packages/storage/src/cookies.ts
+++ b/packages/storage/src/cookies.ts
@@ -1,9 +1,6 @@
 import { getRequestEvent, isServer, type RequestEvent } from "solid-js/web";
-import {
-  SyncStorageWithOptions,
-  addClearMethod,
-  addWithOptionsMethod,
-} from "./index.js";
+import { SyncStorageWithOptions } from "./index.js";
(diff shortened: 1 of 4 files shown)


packages/storage/src/cookies.ts
```diff
@@ -1,6 +1,9 @@
 import { getRequestEvent, isServer, type RequestEvent } from "solid-js/web";
-import { SyncStorageWithOptions } from "./persisted.js";
-import { addClearMethod, addWithOptionsMethod } from "./tools.js";
+import {
+  SyncStorageWithOptions,
+  addClearMethod,
+  addWithOptionsMethod,
+} from "@solid-primitives/storage";
```
thetarnav, line 6: probably should import from a relative path

thetarnav (approved):
Could you add some information to the docs about the approach taken here? I imagine that someone might seek this information weather the package supports encoding utf8 by default and how it is done.
Also if there are any gotchas to this approach they could be listed as well.
Feel free to merge this pr though, docs can be improved afterwards.

thetarnav:
ah the tests are still failing, didn't even notice that, probably some issue with imports again

PR #677, opened 2024-08-28
storage: expose init

atk:


thetarnav (commented):
You could probably alias the return type as `PersistedReturn` if it's being used many times and has grown more complex

hmm it looks like you already do as `EnhancedSignalOrStore`, but the type is different and not actually used anywhere besides the jsdoc

the rest is fine

atk:
I weighted the succinctness of types against the number of indirections that merely added to the confusion and decided for the output type to be `PersistedState<SignalType<S>>` .

PR #593, opened 2024-03-27
Fix and improve ReactiveSet, ReactiveMap and Trigger

Exelord:

PR #688, opened 2024-09-08
Fixes for ReactiveSet

Exelord:
Hi @thetarnav and @atk  :) 

Decided to split [this draft PR](https://github.com/solidjs-community/solid-primitives/pull/593) into 2 chunks Set and Map fixes. Starting with Set here. Best to review commit by commit. Let me know where more reasoning is required, I will add comments.

Thanks!


packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
+    return super.has(value);
   }
 
-  *keys(): IterableIterator<T> {
-    for (const key of super.keys()) {
-      this.#triggers.track(key);
-      yield key;
-    }
-    this.#triggers.track($KEYS);
+  keys(): IterableIterator<T> {
```
Exelord, line 40: Keys are now a proxy on values (not the opposite) as per Set specification.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
+    return this.values();
   }
-  values(): IterableIterator<T> {
-    return this.keys();
+
+  *values(): IterableIterator<T> {
```
Exelord, line 44: Iterators no longer track specific keys. That optimization was working under the assumption that a BREAK will always happen in specific iteration, which might not be always a case as it can depend on external logic. All iterators should always be rerun if any op to the Set will happen.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
+    this.#triggers.track($KEYS);
+
+    for (const value of super.values()) {
+      yield value;
+    }
   }
+
   *entries(): IterableIterator<[T, T]> {
```
Exelord, line 52: Iterators no longer track specific keys. That optimization was working under the assumption that a BREAK will always happen in specific iteration, which might not be always a case as it can depend on external logic. All iterators should always be rerun if any op to the Set will happen.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
-    for (const key of super.keys()) {
-      this.#triggers.track(key);
```
Exelord, line 48: Iterators no longer track specific keys. That optimization was working under the assumption that a BREAK will always happen in specific iteration, which might not be always a case as it can depend on external logic. All iterators should always be rerun if any op to the Set will happen.

packages/set/src/index.ts
Exelord, line 44: Iterators no longer track specific keys. That optimization was working under the assumption that a BREAK will always happen in specific iteration, which might not be always a case as it can depend on external logic. All iterators should always be rerun if any op to the Set will happen.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
 
-  [Symbol.iterator](): IterableIterator<T> {
-    return this.values();
+    for (const entry of super.entries()) {
+      yield entry;
+    }
   }
-  forEach(callbackfn: (value: T, value2: T, set: this) => void) {
+
+  forEach(callbackfn: (value1: T, value2: T, set: Set<T>) => void, thisArg?: any): void {
```
Exelord, line 60: Same as iterators forEach does not track keys, even if in theory a BREAK could happen as a consequence of error. We simply cannot assume it will happen consistently.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
     this.#triggers.track($KEYS);
-    super.forEach(callbackfn as any);
+    super.forEach(callbackfn, thisArg);
```
Exelord, line 62: Added support for `thisArg` as per forEach spec

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
       });
     }
-    return r;
+
+    return result;
   }
+
   clear(): void {
     if (super.size) {
+      super.clear();
```
Exelord, line 92: Clear should always happen before dirtying. This could lead to inconsistent state.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
+
       batch(() => {
-        for (const v of super.keys()) this.#triggers.dirty(v);
-        super.clear();
-        this.#triggers.dirty($KEYS);
+        this.#triggers.dirtyAll();
```
Exelord, line 95: Using dirtyAll is a huge optimization as it we no longer iterate over `super.keys()`. The keys could contain 10000 elements but not all of might be tracked. Now the dirtyAll iterates only over actually tracked keys.

packages/trigger/src/index.ts
```diff
@@ -90,10 +96,12 @@ export class TriggerCache<T> {
  * dirty(1)
  * // this won't cause an update:
  * dirty(2)
+ * // this will cause an update to all keys:
+ * dirtyAll()
  */
 export function createTriggerCache<T>(
```
Exelord, line 102: Suggestion: We should deprecate this function, it creates needless deoptimization creating a new array and rebinding methods. We should recommend using the class directly.

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
 export class ReactiveSet<T> extends Set<T> {
   #triggers = new TriggerCache<T | typeof $KEYS>();
 
-  constructor(values?: Iterable<T> | null) {
+  constructor(values?: readonly T[] | null) {
```
atk, line 21: The specification of Set accepts an Iterable<T>. Why change this to a const array?

packages/set/src/index.ts
```diff
@@ -18,74 +18,81 @@ const $KEYS = Symbol("track-keys");
...
     super();
```
atk, line 22: Why not simply use `super(values)`?

atk (changes requested):

packages/set/src/index.ts
Exelord: Great catch! I was probably referring to ES6 types that do not support Iterables. Will fix it :)

packages/trigger/src/index.ts
thetarnav: I'm fine with that. It's a primitive for advanced use anyway. But in another PR.

packages/set/src/index.ts
Exelord: Not sure I remember the reason, but I think it has sth to do with super(values) overriding all declared properties. All tests are failing if so.

thetarnav: I think that values are not added to the triggermap this way. But maybe there is something else too. I just remember it wasn't working without the loop add.

atk (approved):

packages/trigger/src/index.ts
```diff
@@ -77,11 +82,12 @@ export class TriggerCache<T> {
  * Trigger signals added to the cache only when tracked under a computation,
  * and get deleted from the cache when they are no longer tracked.
  *
- * @returns a tuple of `[track, dirty]` functions
+ * @returns an array of `[track, dirty, dirtyAll]` functions
```
thetarnav, line 85: 3 elements still make a tuple :)
```suggestion
 * @returns a tuple of `[track, dirty, dirtyAll]` functions
```

packages/set/test/index.test.ts
```diff
@@ -100,33 +100,39 @@ describe("ReactiveSet", () => {
       const dispose = createRoot(dispose => {
         createEffect(() => {
           const run: number[] = [];
-          let i = 0;
           for (const key of fn(set)) {
             run.push(key);
-            if (i === 2) break; // don't iterate over all keys
-            i += 1;
           }
           captured.push(run);
```
thetarnav, line 106: Can you make another test instead of modifying the previous one? One focusing just on what changed. This one is too big already.

packages/set/src/index.ts
thetarnav: Can you mention that in the docs?

thetarnav (commented):
Good changes.

packages/set/src/index.ts
Exelord: Sure, Could you point me a place where I should do that?

Exelord: Not sure I understand what is expected here to document to be honest. There is no break optimization anymore, and the previous was a bug. Should't that be mentioned rather in changelog?

thetarnav (approved):

PR #690, opened 2024-09-12
audio: Set COMPLETE state in ended

### LINKED ISSUE #628: AudioState.COMPLETE is never used in primitive code
### Describe the bug

Hi ! I wrote this code earlier and found out it was never triggering :

```typescript
const [audio, controls] = createAudio(track, () => state.playing, () => state.volume);

createEffect(() => {
  console.log(audio.state); // just for debug :')

  if (audio.state === AudioState.COMPLETE) { // this is never triggered !!
    const trackID = findSuitableTrack();
    console.info(`Playing track ${trackID} next.`);

    setState(prev => ({
      history: [...prev.history, prev.currentTrackID],
      currentTrackID: trackID
    }));
  }
});
```

When I looked through the enum, I thought that `AudioState.COMPLETE` would be perfect for my situation.
Even though, it doesn't work, so I went through the primitive code to see what is happening under the hood.

In the whole file (<https://github.com/solidjs-community/solid-primitives/blob/main/packages/audio/src/index.ts>), `AudioState.COMPLETE` is **NEVER** used, it's just defined in the enum.

In the minimal reproduction link, you can see the state ends up to "paused" instead of "complete".

image

My question is: will there be any implementation concerning that in a near future ?
Thanks you ! 

### Minimal Reproduction Link

https://stackblitz.com/edit/github-p2hm8v?file=src%2FApp.tsx

thetarnav:
fixes #628 

Also improved ssr support by setting the initial duration and volume to zero to match the server state.

In the demo I needed to remove the solid-heroicons package because it was causing hydration errors.
Not sure why exactly, but even though it exports preserved jsx, solid-start was still resolving to `dist/server/index.module.js` on the server, instead of the jsx one.

Added a basic test, but @davedbase would have to double check that I haven't done anything wrong. I don't know much about the audio api. But the primitive looks a bit weird.


atk (approved):

davedbase (approved):
Looks good. I like the improvements and eliminating the dependency on solid-heroicons!

PR #697, opened 2024-09-26
pointer: Replace old primitives with `createPointers`

thetarnav:
It's probably a good time to remove that crap that was the pointer package and replace it with a simple, but actually useful primitive.
The old api was just an experimentation that ended up too complicated and not at at useful. The package is not used much either judging by the downloads.
The new `createPointers` primitive is simple, but it something I would use, it just gives you a list of pointer events of pointers that are currently on the screen. All the other magic like position_to_element, is_inside and is_down can be derived from this list, and I'll be adding examples how to do that to the readme later.


packages/pointer/dev/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { createPointerListeners, pointerHover, createPointerList } from "../src/index.js";
+import { createPointerListeners, pointerHover, createPointerList } from "../src/index-ols.js";
```
atk, line 1: This file does not exist.

packages/pointer/src/index.ts
```diff
@@ -1,386 +1,48 @@
...
-  const addEventListener = (type: Many<keyof HTMLElementEventMap>, fn: Handler) =>
-    createEventListener(target, type, guardCB(fn) as any, { passive });
+import { createEventListener } from "@solid-primitives/event-listener";
 
-  entries(nativeHandlers).forEach(
-    ([name, fn]) => fn && addEventListener(`pointer${name}` as keyof HTMLElementEventMap, fn),
-  );
-  if (onGotCapture) addEventListener("gotpointercapture", onGotCapture);
-  if (onLostCapture) addEventListener("lostpointercapture", onLostCapture);
+function upsert_pointer(pointers: PointerEvent[], e: PointerEvent): PointerEvent[] {
```
atk, line 5: What does "upsert" mean?

packages/pointer/test/index.test.ts
```diff
@@ -1,90 +1,117 @@
...
+  const up_event_mouse = new PointerEvent("pointerup", { pointerId: 1, pointerType: "mouse" });
+  const leave_event_mouse = new PointerEvent("pointerleave", { pointerId: 1, pointerType: "mouse" });
+  const cancel_event_mouse = new PointerEvent("pointercancel", { pointerId: 1, pointerType: "mouse" });
+
+  const move_event_touch = new PointerEvent("pointermove", { pointerId: 2, pointerType: "touch" });
+  const down_event_touch = new PointerEvent("pointerdown", { pointerId: 2, pointerType: "touch" });
+  const up_event_touch = new PointerEvent("pointerup", { pointerId: 2, pointerType: "touch" });
+  const cancel_event_touch = new PointerEvent("pointercancel", { pointerId: 2, pointerType: "touch" });
+
+  const ref = document.createElement("div");
```
atk, line 26: Using the same ref for all tests seem a bit brittle.

packages/pointer/dev/index.tsx
thetarnav: haven’t touched demo yet
well I did
but by mistake
it’s wip :)

packages/pointer/src/index.ts
atk: Maybe just use "set"?

davedbase:
Do we still want to consider this @thetarnav ? Might be good to perform a refactor for Solid2 release. I like where this is going though. Can we close this PR and push to the `next` branch instead?

thetarnav:
Yeah why not.
The current pointers package stuffed with overly complex and not-that-useful primitives, so slimming it a bit to start fresh might be a good idea for 2.0

davedbase:
> Yeah why not. The current pointers package stuffed with overly complex and not-that-useful primitives, so slimming it a bit to start fresh might be a good idea for 2.0

Is it something you want to pursue and have time for?

PR #701, opened 2024-09-29
Rewrite build script to build all packages in one process (Removes CJS)

thetarnav:
Rewrites build script to build all packages in one process
Similarly to how testing works
So no turborepo cashing, but it's much faster overall
5s to build all packages instead of a couple of minutes
So should be much easier for new people to fork and build locally and not have to wait so much
Also website deploys could happen on main now as the netlify build should be a lot faster

~~Proper turbo-like caching could be added later as well and be used for tests too, without sacrificing the time to build all packages.~~
`tsc -b` takes care of caching pretty much

Please run `rm -rf packages/*/dist/* .turbo packages/*/.turbo` to clear old files

This also removes CJS support


atk:
Since the removal of CJS support might be considered a breaking change, shouldn't this be a major update? Minor breaking changes are a violation of semver.

thetarnav:
@atk it's not yet done
I'm experimenting with using `tsc -b` for building now. Then esbuild will be necessary only for the two packages that use jsx - `controlled-props` and `virtual`.

PR #704, opened 2024-10-04
Fix and optimize ReactiveMap and ReactiveWeakMap

Exelord:
This PR is improving and fixing ReactiveMap and ReactiveWeakMap, following up on https://github.com/solidjs-community/solid-primitives/pull/593

All noticeable changes have been commented. @thetarnav please let me know in case you have more questions :)

### DIFF
--- a/packages/map/src/index.ts
+++ b/packages/map/src/index.ts
@@ -115,13 +115,11 @@ export class ReactiveMap<K, V> extends Map<K, V> {
 
   clear(): void {
     if (super.size) {
+      super.clear();
+
       batch(() => {
-        for (const v of super.keys()) {
-          this.#keyTriggers.dirty(v);
-          this.#valueTriggers.dirty(v);
-        }
-        super.clear();
-        this.#keyTriggers.dirty($KEYS);
+        this.#keyTriggers.dirtyAll();
+        this.#valueTriggers.dirtyAll();
       });
     }
   }
--- a/packages/map/src/index.ts
+++ b/packages/map/src/index.ts
@@ -107,10 +107,10 @@ export class ReactiveMap<K, V> extends Map<K, V> {
     if (result) {
       batch(() => {
         this.#keyTriggers.dirty($OBJECT);
+        this.#valueTriggers.dirty($OBJECT);
         this.#keyTriggers.dirty(key);
 
         if (isDefined) {
-          this.#valueTriggers.dirty($OBJECT);
           this.#valueTriggers.dirty(key);
         }
       });
--- a/packages/map/src/index.ts
+++ b/packages/map/src/index.ts
@@ -30,7 +30,12 @@ export class ReactiveMap<K, V> extends Map<K, V> {
 
   constructor(entries?: Iterable<readonly [K, V]> | null) {
     super();
-    if (entries) for (const entry of entries) super.set(...entry);
+
+    if (entries) {
+      batch(() => {
+        for (const entry of entries) super.set(...entry);
+      });
+    }
   }
 
   get size(): number {
@@ -137,9 +142,14 @@ export class ReactiveWeakMap<K extends object, V> extends WeakMap<K, V> {
   #keyTriggers = new TriggerCache<K>(WeakMap);
   #valueTriggers = new TriggerCache<K>(WeakMap);
 
-  constructor(initial?: Iterable<readonly [K, V]> | null) {
+  constructor(entries?: Iterable<readonly [K, V]> | null) {
     super();
-    if (initial) for (const v of initial) super.set(v[0], v[1]);
+
+    if (entries) {
+      batch(() => {
+        for (const entry of entries) super.set(...entry);
+      });
+    }
   }
 
   has(key: K): boolean {
--- a/packages/map/src/index.ts
+++ b/packages/map/src/index.ts
@@ -1,7 +1,7 @@
 import { Accessor, batch } from "solid-js";
 import { TriggerCache } from "@solid-primitives/trigger";
 
-const $KEYS = Symbol("track-keys");
+const $OBJECT = Symbol("track-object");
 
 /**
  * A reactive version of `Map` data structure. All the reads (like `get` or `has`) are signals, and all the writes (`delete` or `set`) will cause updates to appropriate signals.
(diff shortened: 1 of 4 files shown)


packages/map/src/index.ts
```diff
@@ -1,7 +1,7 @@
 import { Accessor, batch } from "solid-js";
 import { TriggerCache } from "@solid-primitives/trigger";
 
-const $KEYS = Symbol("track-keys");
+const $OBJECT = Symbol("track-object");
```
Exelord, line 4: Changing to object as now we need to track object keys and values in both caches

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
  * userPoints.set(user1, { foo: "bar" });
  */
 export class ReactiveMap<K, V> extends Map<K, V> {
-  #keyTriggers = new TriggerCache<K | typeof $KEYS>();
-  #valueTriggers = new TriggerCache<K>();
+  #keyTriggers = new TriggerCache<K | typeof $OBJECT>();
+  #valueTriggers = new TriggerCache<K | typeof $OBJECT>();
```
Exelord, line 25: Values will have an object trigger as well to support and values, entires and forEach

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
   }
+
   get size(): number {
-    this.#keyTriggers.track($KEYS);
+    this.#keyTriggers.track($OBJECT);
     return super.size;
   }
 
   *keys(): MapIterator<K> {
+    this.#keyTriggers.track($OBJECT);
```
Exelord, line 42: Now, keys invalidate whenever a key has been added, changed or removed. No matter if a key was used or not.

This is Removing the same keys optimization as in Set, as we cannot assume a break will happen consistently on each computation.

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
       yield key;
     }
-    this.#keyTriggers.track($KEYS);
   }
+
   *values(): MapIterator<V> {
-    for (const [key, v] of super.entries()) {
-      this.#valueTriggers.track(key);
-      yield v;
+    this.#valueTriggers.track($OBJECT);
```
Exelord, line 50: Values are no longer tracking changes in keys changes. Instead they track only values changes.

This is also Removing the same keys optimization as in Set, as we cannot assume a break will happen consistently on each computation.

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+
+    for (const value of super.values()) {
+      yield value;
     }
-    this.#keyTriggers.track($KEYS);
   }
+
   *entries(): MapIterator<[K, V]> {
+    this.#keyTriggers.track($OBJECT);
+    this.#valueTriggers.track($OBJECT);
```
Exelord, line 59: Entries are now tracking global changes to keys and values, ensuring they will be dirtied whenever anything has changed.

This is also Removing the same keys optimization as in Set, as we cannot assume a break will happen consistently on each computation.

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+
     for (const entry of super.entries()) {
-      this.#valueTriggers.track(entry[0]);
       yield entry;
     }
-    this.#keyTriggers.track($KEYS);
   }
 
-  // writes
+  forEach(callbackfn: (value: V, key: K, map: Map<K, V>) => void, thisArg?: any): void {
```
Exelord, line 66: Similar to entires, forEach is now tracking global changes to keys and values, ensuring they will be dirtied whenever anything has changed.

This is also Removing the same keys optimization as in Set, as we cannot assume a break will happen consistently on each computation.

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+    this.#keyTriggers.track($OBJECT);
+    this.#valueTriggers.track($OBJECT);
+    super.forEach(callbackfn, thisArg);
```
Exelord, line 69: We now support as well `thisArg`, that wasn't supported

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+    const hasChanged = super.get(key) !== value;
+    const result = super.set(key, value);
+
+    if (hasChanged || hadNoKey) {
+      batch(() => {
+        if (hadNoKey) {
+          this.#keyTriggers.dirty($OBJECT);
+          this.#keyTriggers.dirty(key);
+        }
+        if (hasChanged) {
```
Exelord, line 93: Set is now invalidating values only if they changed

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+          this.#valueTriggers.dirty($OBJECT);
```
Exelord, line 94: Global values are dirtied for values, entries and forEach support

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+
+    if (result) {
       batch(() => {
+        this.#keyTriggers.dirty($OBJECT);
+        this.#valueTriggers.dirty($OBJECT);
         this.#keyTriggers.dirty(key);
-        this.#keyTriggers.dirty($KEYS);
-        this.#valueTriggers.dirty(key);
+
+        if (isDefined) {
```
Exelord, line 113: Value trigger is only dirtied if the value actually has changed.

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
   delete(key: K): boolean {
-    const r = super.delete(key);
-    if (r) {
+    const isDefined = super.get(key) !== undefined;
+    const result = super.delete(key);
+
+    if (result) {
       batch(() => {
+        this.#keyTriggers.dirty($OBJECT);
+        this.#valueTriggers.dirty($OBJECT);
```
Exelord, line 110: This is important: we do dirty global values trigger **always**, as values() may depend on order. Even though the actual value of the key may not change, removing it, will have impact on amount of iterations in *values()

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
+      super.clear();
+
       batch(() => {
-        for (const v of super.keys()) {
-          this.#keyTriggers.dirty(v);
-          this.#valueTriggers.dirty(v);
-        }
-        super.clear();
-        this.#keyTriggers.dirty($KEYS);
+        this.#keyTriggers.dirtyAll();
```
Exelord, line 127: clear is now optimized using new dirtyAll()

packages/map/src/index.ts
```diff
@@ -21,100 +21,114 @@ const $KEYS = Symbol("track-keys");
...
       });
     }
-    return r;
+
+    return result;
   }
+
   clear(): void {
     if (super.size) {
+      super.clear();
```
Exelord, line 124: clear is now fixed, calling .clear() at before dirtying

PR #707, opened 2024-10-14
Add `createEvent` and `createSubject`

devagrawal09:
Adds [`solid-events`](https://github.com/devagrawal09/solid-events) as a primitive.

### DIFF
--- a/packages/events/LICENSE
+++ b/packages/events/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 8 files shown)


atk:
The name events easily lends itself to confusion with UI Events. What you implement here is a message bus, not an event system.

thetarnav:
This should be extending the [event-bus package](https://github.com/solidjs-community/solid-primitives/tree/main/packages/event-bus#readme)
I don’t want to have two packages for the same problem.

devagrawal09:
> Why does this depend on rxjs

The rxjs dependency is temporary, I'll work on getting rid of that soon

>This should be extending the [event-bus package](https://github.com/solidjs-community/solid-primitives/tree/main/packages/event-bus#readme)
I don’t want to have two packages for the same problem.

Fair, this is mostly a separate primitive because it's a solidhack submission, if this makes more sense as an extension to event-bus I'm happy to contribute.
I'll spend some time next week rewriting the internals to use event bus instead of rxjs. Would love to hear your ideas around what the api should look like if this were to be an extension to event-bus.

davedbase:

Issue #711, opened 2024-10-27
Running in a worker throws Uncaught ReferenceError: window is not defined


cherealnice:
### Describe the bug

I am trying to run captureStoreUpdates in another thread, and I'm getting Uncaught ReferenceError: window is not defined. This seems to be due to the scheduleIdle fn trying to call window.requestIdleCallback. I'm suggesting a change where we check typeof window and fallback to throttle in the case that it doesn't exist, like what is done for when requestIdleCallback is not defined

### Minimal Reproduction Link

https://stackblitz.com/edit/github-vu8q2i?file=src%2Fworker.ts

PR #712, opened 2024-10-27
schedule - Add check for typeof window, fallback to throttle if undefined

cherealnice:
Add check for typeof window, fallback to throttle if undefined

### DIFF
--- a/packages/scheduled/src/index.ts
+++ b/packages/scheduled/src/index.ts
@@ -112,7 +112,7 @@ export const throttle: ScheduleCallback = (callback, wait) => {
 export const scheduleIdle: ScheduleCallback = isServer
   ? () => Object.assign(() => void 0, { clear: () => void 0 })
   : // requestIdleCallback is not supported in Safari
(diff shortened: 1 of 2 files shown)


cherealnice:
#711

packages/scheduled/src/index.ts
```diff
@@ -112,7 +112,7 @@ export const throttle: ScheduleCallback = (callback, wait) => {
 export const scheduleIdle: ScheduleCallback = isServer
   ? () => Object.assign(() => void 0, { clear: () => void 0 })
   : // requestIdleCallback is not supported in Safari
-    (window.requestIdleCallback as typeof window.requestIdleCallback | undefined)
+    (typeof window !== 'undefined' && window.requestIdleCallback as typeof window.requestIdleCallback | undefined)
```
thetarnav, line 115:
```suggestion
    (typeof requestIdleCallback !== 'undefined')
```
Actually, wouldn't this be better?
since it's on a global scope anyway
if it's present, use it, that's the point
`window` is unnecessary

PR #724, opened 2024-12-09
storage: fix async initialization

atk:
See https://github.com/solidjs-community/solid-primitives/discussions/708


thetarnav:
Can you explain why the setter has to be ran with owner? I don't think that signal setters use the current owner at all. Which makes sense as they should be used generally on user events where there is no reactive context.

PR #728, opened 2025-01-01
Add `MapEntries` control flow component.

AlexErrant:
This PR adds `<MapEntries>`, which is a small iteration on `<Entries>`. A character-level diff may be useful when reviewing this PR, as most of my changes were copy/paste/modify. E.g. https://www.diffchecker.com/zRdYuasY/ (note the "Highlight change", "Character" option on the left side. (That link expires in a month - there is no permanent option.))

I also add a demo for `<Entries>`, `<MapEntries>`, and ReactiveMap working with `<MapEntries>`, which accounts for most of the linecount. Here's a video of the `<MapEntries>` demo showing how it's fine-grained.

https://github.com/user-attachments/assets/e1c1b6c3-9e44-4af7-b8c7-18aa2dc7840b

If this gets accepted I'll also contribute `<SetEntries>` in a followup.


thetarnav:
All good but I this should be added to the `map` package directly.

README.md
```diff
@@ -89,7 +89,7 @@ The goal of Solid Primitives is to wrap client and server side functionality to
...
+|[map](https://github.com/solidjs-community/solid-primitives/tree/main/packages/map#readme)|[![STAGE](https://img.shields.io/endpoint?style=for-the-badge&label=&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-2.json)](https://github.com/solidjs-community/solid-primitives/blob/main/CONTRIBUTING.md#contribution-process)|[ReactiveMap](https://github.com/solidjs-community/solid-primitives/tree/main/packages/map#reactivemap)<br />[ReactiveWeakMap](https://github.com/solidjs-community/solid-primitives/tree/main/packages/map#reactiveweakmap)<br />[MapEntries](https://github.com/solidjs-community/solid-primitives/tree/main/packages/map#mapentries)|[![SIZE](https://img.shields.io/bundlephobia/minzip/@solid-primitives/map?style=for-the-badge&label=)](https://bundlephobia.com/package/@solid-primitives/map)|[![VERSION](https://img.shields.io/npm/v/@solid-primitives/map?style=for-the-badge&label=)](https://www.npmjs.com/package/@solid-primitives/map)|
```
AlexErrant, line 92: Not sure if `<MapEntries>` belongs under the `Reactivity` section, but whatever.

thetarnav: Hmm my bad
I don’t care about the consistency in categories that much
But I thought that MapEntries directly depends on the map package
But it’s only a dev dependency for the demo right?
If usage of ReactiveMap is not required then It shouldn’t be in the map package really, I just wanted them to be together if they were connected.
Also the demo probably shouldn’t use the map package if it’s not required.
I don’t want people to get confused about the relation between these two like I did.

packages/map/dev/reactiveMapEntries.tsx
```diff
@@ -0,0 +1,101 @@
...
+const randomIndex = (list: readonly any[]): number => Math.floor(Math.random() * list.length);
+const getRandomFood = () => foods[randomIndex(foods)]!;
+const randomKey = (map: ReactiveMap<string, string>): string => {
+  const keys = Array.from(map.keys());
+  return keys[randomIndex(keys)]!;
+};
+
+export default function App() {
+  const [store, setStore] = createStore<{ entries: ReactiveMap<string, string> }>({
+    entries: new ReactiveMap([
```
thetarnav, line 45: This could just be a signal
using a store for only one field is kinda an overkill

thetarnav:
Ok
my bad with the initial suggestion, I haven't looked too deeply at the changes and assumed that `MapEntries` directly depends on the `map` package
I think that `MapEntries` should be added to `keyed` package like it was at the beginning (you can just revert I guess)
The demo shows both the usage with `Map` and `ReactiveMap` which is great
But the docs should mention it as well, explain when you'd use one or another, how does that impact rendering
Also some tests are missing as well
It would be useful to ensure that the basic interaction between `Map`, `ReactiveMap` and `MapEntries` is covered by the tests and doesn't change.

thetarnav:
`MapEntries` doesn't have any special interaction with `ReactiveMap`, right?
The demo works the same way for both.
If there is no difference then it might be pointless to even include `ReactiveMap` there.

thetarnav:
this month passed fast

image

AlexErrant:
> I think that MapEntries should be added to keyed package like it was at the beginning

Some food for bikeshedding (I _really_ do not care, just bringing it up cuz why not), as of https://github.com/solidjs-community/solid-primitives/commit/9ebe450269f90a246df2a2676d38ae675ad664b8 , `<Entries>` no longer uses `keyArray`. Nor does `<MapEntries>`, so one could argue that they don't belong in `keyed`. Maybe `Entries`, `MapEntries`, and `SetEntries` should be their own package/primitive? Or part of [List](https://github.com/solidjs-community/solid-primitives/tree/main/packages/list)?

> this month passed fast

Ah, I probably pasted the wrong link; updated the post >_>

thetarnav:
Let's stick with `keyed` right now.
Can be moved to a different package if needed later.

packages/keyed/test/index.test.tsx
AlexErrant, line 1: FYI the tests were based on [the List tests](https://github.com/solidjs-community/solid-primitives/blob/37f0616fc165f5368f817adb11a611b84624fc2e/packages/list/test/index.test.tsx#L31).

packages/keyed/README.md
```diff
@@ -152,6 +153,38 @@ Third argument of the map function is an index signal.
...
+    <div>
+      {key}: {value()}
+    </div>
+  )}
+</MapEntries>;
+```
+
+### Index argument
+
+Third argument of the map function is an index signal.
```
thetarnav, line 176:
```suggestion
Third argument of the map function is an index signal.

`MapEntries` is using [`Map#key()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/keys) so the index and resulting JSX will follow the insertion order.
```

packages/keyed/README.md
```diff
@@ -152,6 +153,38 @@ Third argument of the map function is an index signal.
...
+## `<MapEntries>`
+
+Creates a list of elements by mapping Map entries. Similar to Solid's `<For>` and `<Index>`, but here, render function takes three arguments, and both value and index arguments are signals.
+
+### How to use it
+
+```tsx
+import { MapEntries } from "@solid-primitives/keyed";
+
+<MapEntries of={map()} fallback={<div>No items</div>}>
```
thetarnav, line 165:
```suggestion
import { MapEntries } from "@solid-primitives/keyed";

const [map, setMap] = createSignal(new Map());

<MapEntries of={map()} fallback={<div>No items</div>}>
```

packages/keyed/dev/entries.tsx
```diff
@@ -0,0 +1,106 @@
+// changes to this file might be applicable to similar files - grep 95DB7339-BB2A-4F06-A34A-25DDF8BF7AF7
```
thetarnav, line 1: ?

packages/keyed/dev/entries.tsx
```diff
@@ -0,0 +1,106 @@
...
+const randomIndex = (list: readonly any[]): number => Math.floor(Math.random() * list.length);
+const getRandomFood = () => foods[randomIndex(foods)]!;
+const randomKey = (record: Record<string, string>): string => {
+  const keys = Object.keys(record);
+  return keys[randomIndex(keys)]!;
+};
+
+export default function App() {
+  const [store, setStore] = createStore<{ entries: Record<string, string> }>({
+    entries: {
```
thetarnav, line 44: This could just be a signal `const [entries, setEntries] = createSignal({...})`

packages/keyed/dev/mapEntries.tsx
```diff
@@ -0,0 +1,110 @@
+// changes to this file might be applicable to similar files - grep 95DB7339-BB2A-4F06-A34A-25DDF8BF7AF7
```
thetarnav, line 1: ?

packages/keyed/dev/entries.tsx
AlexErrant: Do you think the comment is unnecessary? Or is the comment confusing? I'm confused by your confusion, lol.

packages/keyed/dev/entries.tsx
AlexErrant: I feel like this should remain a `store` because it's an object which Solid natively tracks, and making it a signal would kinda make the "Clone" button pointless (or more pointless than it already is, hah).

thetarnav: youre right lets leave it

packages/keyed/dev/entries.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const addRandom = () => {
+    setStore("entries", Math.random().toString(), getRandomFood());
+  };
+  const removeRandom = () =>
+    setStore(
+      "entries",
+      produce(p => {
+        delete p[randomKey(p)];
+      }),
+    );
```
thetarnav, line 62: Woudn't this be the same? it's consistent with changing.
```suggestion
    setStore("entries", p => ({ [randomKey(p)]: undefined }));
```

packages/keyed/dev/entries.tsx
thetarnav: Never seen that before
interesting way to say that multiple files are similar
I guess it's fine

packages/keyed/dev/entries.tsx
AlexErrant: oh duh!

PR #738, opened 2025-01-15
storage: solve hydration issue

atk:
In order to simplify avoiding hydration issues, makePersisted now accepts `isHydrated` from the lifecycle package as an option to delay the initialization from storage until the parent is hydrated; since this way both server and client will have the same default on initial rendering, this will make them render the same, thus no longer causing hydration mismatch errors.

Another solution would be to sync the storage with the server and use makePersisted on the server, too.

This addresses https://github.com/solidjs-community/solid-primitives/discussions/737

### DIFF
--- a/.changeset/few-ties-boil.md
+++ b/.changeset/few-ties-boil.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/storage": minor
+---
+
+simplify workaround for hydration mismatches based on storage initialization
--- a/packages/storage/README.md
+++ b/packages/storage/README.md
@@ -51,7 +51,7 @@ type PersistedOptions<Type, StorageOptions> = {
 - values persisted in asynchronous storage APIs will not overwrite already changed signals or stores
 - setting a persisted signal to undefined or null will remove the item from the storage
 - to use `makePersisted` with other state management APIs, you need some adapter that will project your API to either the output of `createSignal` or `createStore`
-- if you experience hydration mismatch issues, add `isHydrated` from the [lifecycles package](../lifecycle/) to your options to delay the initialization until the parent component is hydrated
+- if you experience hydration mismatch issues, set `deferInit` to true to delay the initialization from storage until the parent component is hydrated - this way, client and server will use the same initial data and avoid hydration conflicts
 
 ### Using `makePersisted` with resources
 
@@ -68,7 +68,7 @@ result is discarded not to overwrite more current data.
 
 ### Using `makePersisted` with Suspense
 
-In case you are using an asynchronous storage and want the initialisation mesh into Suspense instead of mixing it with Show, we provide the output of the initialisation as third part of the returned tuple:
+In case you are using an asynchronous storage and want the initialization mesh into Suspense instead of mixing it with Show, we provide the output of the initialization as third part of the returned tuple:
 
 ```ts
 const [state, setState, init] = makePersisted(createStore({}), {
--- a/packages/storage/README.md
+++ b/packages/storage/README.md
@@ -40,6 +40,8 @@ type PersistedOptions<Type, StorageOptions> = {
   deserialize?: (value: string) => Type(value),
   // sync API (see below)
   sync?: PersistenceSyncAPI
+  // isHydrated from @solid-primitives/lifecycle
+  isHydrated?: () => boolean
 };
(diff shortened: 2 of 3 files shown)


thetarnav:
Why not use isHydrated (or it's implementation directly since it's just some "is hydrating" checks and onMount) in the primitive by default?
The primitives should be usable with ssr without needing to enable that support.
```ts
if (!isServer && sharedConfig.context) {/*delay init*/}
else {/*init immediately*/}
```
This is what all the primitives what read something from the dom like `createWindowSize` do. Even though `window.innerWidth` can be read immediately, it is delayed when hydrating.

atk:
I thought we should leave it as an option and I did not want to increase the package count.

thetarnav:
Is there a way to ensure consistency automatically?
I don't mean sync.
Just a way for the server to "tell the client" to use a fallback or not?
Like this seems like something that the primitive should handle on it's own.
Maybe by using resources?
Which also have a `onHydrated` callback option that could be used to handle hydration.

thetarnav:
> Is there a way to ensure consistency automatically?

If this is out of scope of this pr, and you just want to introduce a temporary fix, then I don't mind.
But the api could be less convoluted imo.
For example as a option flag:
```ts
makePersisted(init, {useInitDuringHydration: true})
```
As opposed to telling users to pull another dependency.
It's not like `isHydrated` can have multiple implementations to allow for an interface.

atk:
There is no way for us to know on either server or client what data the other side will have unless we either sync them manually or deliberately let them be out of sync.

> It's not like isHydrated can have multiple implementations to allow for an interface

No, but one could also choose this to block initialization until any moment of your choice.

atk:
OK, I switched to your suggested solution.

thetarnav:
looks good
although there are some “isHydrated” leftovers in the comments and readme
also if you

> No, but one could also choose this to block initialization until any moment of your choice.

If you think this valuable, `delayInit` could accept a function matching onMount type (`{delayInit: onMount}`), which would give the flexibility while keeping the api simple.
But yeah, no idea if that’s necessary.

jcalfee:
The new `delayInit` name is great.  I was thinking defer, but "delay" helps make this unique.  Doesn't isHydrated happen before onMount though?  If it isHyraded v. onMount makes any real difference, esp. re-rendering, I suggest to choose well if one always works or go back to the original PR with a function passed in but with the name changed to `delayInit`.

And fyi, good idea in your example @thetarnav, but I don't think onMount returns a value (I'm not on the latest version though).  We might want to keep isHydrated in the docs at least..  I'm making some assumptions here though I don't know all the details about the difference in onMount and isHydrated.
```js
function onMount(fn) {
  createEffect(() => untrack(fn));
}
```

thetarnav:
@jcalfee 

> Doesn't isHydrated happen before onMount though?

No, isHydrated is pretty much onMount as a signal with a hydration check:
https://github.com/solidjs-community/solid-primitives/blob/main/packages/lifecycle/src/index.ts#L35

> good idea in your example @thetarnav, but I don't think onMount returns a value

why would it return a value?

currently the logic is:
```ts
const initialize = () => {...}
if (options.deferInit) {
    onMount(initialize)
} else {
    initialize()
}
```

and my idea would look something like this
```ts
const initialize = () => {...}
if (options.deferInit) {
    options.deferInit(initialize)
} else {
    initialize()
}
```

jcalfee:
I was referring to this:

> If you think this valuable, delayInit could accept a function matching onMount type (`{delayInit: onMount}`), 

If onMount did return a value that would be useful.

atk:
`options.deferInit(initialize)` is somewhat of a foot gun IMO. If you don't read the instructions carefully, you'll easily break things. I think a boolean i option s fine for now. We can iterate if necessary.

aguilera51284:
any update about this?

Tyg-g:
Hey, guys, I came up with a semi-temporary solution. Maybe it will help sb while this issue gets resolved _"officially"_.

I posted it here: https://github.com/solidjs-community/solid-primitives/discussions/737#discussioncomment-16333843

davedbase:
This PR is now old. I'll close it but @atk hopefully we too the learning and adapted it to the 2.0 effort.

PR #754, opened 2025-02-14
fix: selection contenteditable node detection

atk:
If you triple click on an li inside a list inside a contenteditable div (except for the last one), you get a selectionEnd of NaN. This is because for reasons unknown, in this case we receive the parent of the text node instead of the text node in the range.


thetarnav:
Why not change the logic only in the NaN case?

atk:
This fix addresses the only cause for the NaN case that I could find. First checking for NaN and only then checking if the parent matches will only increase the code size.

PR #760, opened 2025-02-16
Add `Match` component and package

thetarnav:
Control-flow component for matching discriminated union (tagged union) members.

```tsx
type MyUnion = {
  kind: 'foo',
  foo:  'foo-value',
} | {
  kind: 'bar',
  bar:  'bar-value',
}

const [value, setValue] = createSignal<MyUnion>({kind: 'foo', foo: 'foo-value'})
 
<Match on={value()} case={{
  foo: v => <>{v().foo}</>,
  bar: v => <>{v().bar}</>,
}} />
```


chiefcll:
Any progress on releasing this? I'd like to add it to the lightning demo app

PR #762, opened 2025-02-20
add createAsync primitive

birkskyum:
This is the createAsync primitive from the `@solidjs/router`

https://github.com/solidjs/solid-router/blob/main/src/data/createAsync.ts


thetarnav:
Although I don't mind, after solid 2.0 this will just get deprecated?

birkskyum:
Yes, it'll be obsoleted.

ryansolid:
Hold on this until we decide this is worth the temporary disruption.

Issue #757, opened 2025-02-15
All packages "Stage 0" on https://primitives.solidjs.community


thehappycheese:
### Describe the bug

The listing at https://primitives.solidjs.community/ currently shows all packages as Stage 0
The main [readme.md](https://github.com/solidjs-community/solid-primitives/blob/main/README.md#primitives) on this repo the correct stages appear to be listed.

(thank you to all authors/maintainers solid-primitives is fantastic! 💖) 

e.g. `storage` is meant to be stage 3;

![Image](https://github.com/user-attachments/assets/f8d61a57-e6b7-457d-8b48-bbc3e827d1a2)

### Minimal Reproduction Link

https://primitives.solidjs.community/

PR #768, opened 2025-03-02
fix: stage badge for specific primitive

jaysencpp:
The stage badge is not displaying correctly.

Relevant issue: #757

This was already fixed for the badges shown on the front page (#759), however it's not fixed for a specific primitive's page.

To reproduce error:

* Open any primitive specific page (e.g. https://primitives.solidjs.community/package/active-element)
* See that the Stage is not correct.

### DIFF
--- a/site/src/routes/package/[name]/(package).tsx
+++ b/site/src/routes/package/[name]/(package).tsx
@@ -40,7 +40,7 @@ const Page: Component = () => {
       return cachedData()?.primitives ?? dataResource()?.primitives;
     },
     get stage() {
-      return cachedData()?.stage ?? dataResource()?.stage;
+      return cachedData()?.primitive?.stage ?? dataResource()?.primitive?.stage;
     },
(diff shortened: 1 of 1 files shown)


atk:
The fix is already on main, we just need to roll it out.

jaysencpp:
> The fix is already on main, we just need to roll it out.

Oh, that's great. Though I'm not quite sure what you mean by that since what I'm seeing on the main branch is wrong: https://github.com/solidjs-community/solid-primitives/blob/main/site/src/routes/package/%5Bname%5D/(package).tsx#L43

aquaductape:
Sorry @jaysencpp, I didn't realize you made the correct fix, I went ahead and pushed a fix https://github.com/solidjs-community/solid-primitives/pull/793 and the live site shows the correct stage badge now.

Thank you so much for taking the time to fix this, had I seen your PR first, I would have merged it

PR #781, opened 2025-03-26
Fix create deep signal

juanrgm:
The setter of `createDeepSignal` is not working if you pass a non-function value.

```ts
const [data, { mutate }] = createResource(() => ({ counter: 1 }), {
  storage: createDeepSignal,
});
// mutate(() => ({ counter: 2 })); // works
mutate({ counter: 2 }); // not working
createEffect(() => {
  console.log(data()?.counter); // 1
});
```

### DIFF
--- a/packages/resource/src/index.ts
+++ b/packages/resource/src/index.ts
@@ -298,7 +298,8 @@ export function createDeepSignal<T>(v?: T): Signal<T> {
   return [
     () => store[0],
     (update: T) => (
-      setStore(0, reconcile(typeof update === "function" ? update(unwrap(store[0])) : v)), store[0]
+      setStore(0, reconcile(typeof update === "function" ? update(unwrap(store[0])) : update)),
+      store[0]
     ),
   ] as Signal<T>;
 }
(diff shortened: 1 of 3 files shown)


atk (approved):

packages/resource/src/index.ts
```diff
@@ -298,7 +298,8 @@ export function createDeepSignal<T>(v?: T): Signal<T> {
   return [
     () => store[0],
     (update: T) => (
-      setStore(0, reconcile(typeof update === "function" ? update(unwrap(store[0])) : v)), store[0]
+      setStore(0, reconcile(typeof update === "function" ? update(unwrap(store[0])) : update)),
+      store[0]
```
juanrgm, line 302: Is this line necessary? @atk

atk: The setter function is supposed to get the original object, not a Proxy. If it gets mutated, we need reconcile in order to facilitate the updates safely. Using the setter directly could work, but also introduce errors.

thetarnav (approved):

PR #784, opened 2025-04-02
Make ReactiveSet.clear() and ReactiveMap.clear() more efficient by only notifying listeners of existing members

webstrand:
Listeners of non-existing members cannot observe different values before or after `.clear()` `.get(nonexsting) -> undefined` or `.has(nonexisting) -> false` so there's no use in notifying them.

### DIFF
--- a/packages/set/src/index.ts
+++ b/packages/set/src/index.ts
@@ -88,13 +88,14 @@ export class ReactiveSet<T> extends Set<T> {
   }
 
   clear(): void {
-    if (super.size) {
+    if (!super.size) return;
+    batch(() => {
+      this.#triggers.dirty($KEYS);
+      for (const member of super.values()) {
+        this.#triggers.dirty(member);
+      }
       super.clear();
-
-      batch(() => {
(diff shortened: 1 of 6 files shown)


packages/set/src/index.ts
```diff
@@ -88,13 +88,14 @@ export class ReactiveSet<T> extends Set<T> {
...
 
   clear(): void {
-    if (super.size) {
+    if (!super.size) return;
+    batch(() => {
+      this.#triggers.dirty($KEYS);
+      for (const member of super.values()) {
+        this.#triggers.dirty(member);
+      }
       super.clear();
```
webstrand, line 97: I did some benchmarking and did not see a performance hit to moving `super.clear()` into the batch operation. But if you like, I can save a snapshot prior to `.clear()` and iterate over that in batch.

atk:
This is still missing a changeset. Run `npx changeset` in the project root.

webstrand:
Ah, I had no idea that message was for me and not the maintainers. Fixed.

PR #786, opened 2025-04-20
fix: <Repeat> reactivity

### LINKED ISSUE #785: <Repeat/> does not update when its times.prop changes
### Describe the bug

```tsx
function Counter() {
  const [count, setCount] = createSignal(2);
  const increment = () => setCount((count) => count + 1);
  const decrement = () => setCount((count) => count - 1);

  return (
    <>
      <button onClick={increment}>increment</button>
      <button onClick={decrement}>decrement</button>
      <Repeat times={count()}>{() => "hallo"}</Repeat>
    </>
  );
}
```

steps to reproduce bug:
- initial render: `hallohallo`
- press decrement: stays `hallohallo`
- press increment: stays `hallohallo`
- only when times becomes 0 it will render empty

[I was able to narrow the bug to the usage of `createMemo` in `<Repeat/>`](https://playground.solidjs.com/anonymous/cfa666cf-17cc-4aca-a091-a19f949bd2a5)

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/6c6feac1-bd0d-41f1-b6ec-61bca51182e3

atk:
fixes #785

### DIFF
--- a/.changeset/thick-moose-grow.md
+++ b/.changeset/thick-moose-grow.md
@@ -1,0 +1,5 @@
+---
+"@solid-primitives/range": patch
+---
+
+fix: reactivity in <Repeat>
--- a/packages/range/src/repeat.ts
+++ b/packages/range/src/repeat.ts
@@ -54,7 +54,7 @@ export function repeat<T>(
       const diff = prevLen - len;
       if (diff > 0) {
         for (let i = prevLen - 1; i >= len; i--) disposers[i]!();
-        items.splice(len, diff);
+        items = items.slice(0, len);
         disposers.splice(len, diff);
         return items;
       }


thetarnav:
can i has test case? :)

webstrand:
You'll also need to clone the array here:

https://github.com/solidjs-community/solid-primitives/blob/7fbc758b495dd89c0f029c73c5d798749b7cc5b0/packages/range/src/repeat.ts#L50

currently it reuses the array when growing, but slices it when shrinking.

I hadn't realized anyone else was working on a fix for this, I did not mean to duplicate work.

atk:
I just tested it - this approach does not work, since we return the initial instance and overwriting it internally will not change the external reference. This needs more thought.

PR #787, opened 2025-04-20
fix: repeat must not mutate its result object in place (#785)

### LINKED ISSUE #785: <Repeat/> does not update when its times.prop changes

webstrand:
Fixes #785

This substantially rewrites `repeat` to be more straight-forward with fewer closures, unlike #786 which is more conservative.

### DIFF
--- a/packages/range/src/repeat.ts
+++ b/packages/range/src/repeat.ts
@@ -1,4 +1,4 @@
-import { Accessor, JSX, createMemo, createRoot, onCleanup, untrack } from "solid-js";
+import { Accessor, JSX, createMemo, createRoot, onCleanup } from "solid-js";
 import { toFunction } from "./common.js";
 
 /**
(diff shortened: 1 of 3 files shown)


packages/range/src/repeat.ts
```diff
@@ -23,60 +23,68 @@ export function repeat<T>(
...
+      next[index] = createRoot(dispose => {
+        disposers[index] = dispose;
+        return mapFn(index);
+      });
+    }
 
-  const memoLen = createMemo(() => Math.floor(Math.max(times(), 0)));
-  return () => {
-    const len = memoLen();
-    return untrack(() => {
```
webstrand, line 75: I removed this `untrack` call+closure, as far as I can tell it's redundant: `createRoot` already does both Owner and Listener.

webstrand:
This is actually kind of broken still, if the array returned from `repeat()()` is mutated, internal state can be corrupted

thetarnav:
That’s fine
The same probably happens with mapArray and others
When mutating the result one should copy
Nothing broken with that

webstrand:
Is it appropriate to narrow the return type to `readonly T[]` then to signal to users not to mutate the return value?

I've narrowed the return type in the most recent batch of commits, and I've fixed a bug in my implementation and added another test.

I'm not sure if you'd rather have a single commit in the PR or multiple?

thetarnav:
readonly arrays are not allowed by JSX types, so I wouldn't do that.

PR #788, opened 2025-04-25
Updates tauri-storage for Tauri 2 stable

bpevs:
The api for plugin-store has changed since the last commit.
This pr updates that, and allows for passing store options obj.

### DIFF
--- a/packages/storage/src/tauri.ts
+++ b/packages/storage/src/tauri.ts
@@ -14,7 +14,7 @@ import { AsyncStorage } from "@solid-primitives/storage";
  */
 export function tauriStorage(
   name = "solid-storage.dat",
-  options: StoreOptions,
+  options: StoreOptions = {},
 ) {
   const api: AsyncStorage = {
     _store: null,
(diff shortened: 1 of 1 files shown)


atk:
There might still be projects using v1, so I would prefer to split the module.

bpevs:
oh this is meant to be compatible with v1? I didn't realize the apis were exactly the same; the example app uses 2.0-beta.

atk:
No, it's just that support for both is on my to-do list.

bpevs:
Okie. Mmmm If you want me to format/split this in a particular way, I don't mind making a few changes to help. But I'm not sure on your preferences on how to split, and with this small amount of code, I feel like the communication might be more work than just making the actual edits.

So feel free to directly edit/close/do whatever with this pr, if it's easier. I was just making the pr since I had already made these updates locally to support my app.

atk:
I had not yet decided whether to call the sub-modules tauri-v2/-v1 or tauri/-legacy or transparently support both.

vincehi:
Oh no, why has it been closed?

PR #791, opened 2025-05-29
Fix for `<List>` "unused" signal

maciek50322:
I found that the "signal created when used" is giving wrong values when it's read only later after changes (eg. only in event callback function and none in jsx or other signals or component creation - work correctly if it's used there).

### DIFF
--- a/packages/list/README.md
+++ b/packages/list/README.md
@@ -36,8 +36,7 @@ function Component() {
       {(value, index) => {
         return (
           <div>
-            {" "}
-            {index()}: {value()}{" "}
+            {index()}: {value()}
           </div>
         );
       }}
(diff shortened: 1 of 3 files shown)


maciek50322:
Done.

>  `mapArray` and `keyArray` are using `mapFn.length` to check if the index signal is used or not.

But for that I argue that you can use arguments that are over what `mapFn.length` gives...

```js
function mapFn(...args) {
    const value = args[0];
    const index = args[1];
    console.log(`value = ${value}, index = ${index}`);
}

console.log(mapFn.length); // logs: 0
mapFn(1, 2);               // logs: value = 1, index = 2
```

Probably uncommon but still

maciek50322:
Ok so I found different way to fix this issue - when creating signal get fresh item index & value instead of one that it was created with. After this change new test passes

PR #796, opened 2025-06-04
fix: (fs) rm also removes signal

### LINKED ISSUE #794: SyncFileSystem does not throw when accessing a removed file that has been accessed in an effect/memo/resource
### Describe the bug

Given the following testing setup:

```tsx
const fs = createSyncFileSystem(makeVirtualFileSystem());
fs.writeFile("index.html", "hallo");
fs.writeFile("index2.html", "hallo");

createEffect(() => fs.readFile("index.html"));

test("index.html");
test("index2.html");

function test(path: string){
  try {
    fs.rm(path)
    const result = fs.readFile(path);
    console.error(`FAILURE ${path}: Expected throw but received "${result}"`);
  } catch (error) {
    console.log(`SUCCESS ${path}`);
  }
}
```

Currently
- the test for `index.html` will fail, because it received `"hallo"` from `fs.readFile`
- the test for `index2.html` will succeed, because `fs.readFile` threw

Reading the file in an effect seems to cause the file to not be cleaned up when being removed.

### Minimal Reproduction Link

https://playground.solidjs.com/anonymous/a4c5374f-9432-4ec1-bf37-e65888e54a8f

atk:
fixes #794

The rm method deleted the signals for directories, but not for the files themselves, so even deleted files were still accessible without throwing the expected error.


.changeset/clean-baboons-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@solid-primitives/filesystem": minor
```
thetarnav, line 2:
```suggestion
"@solid-primitives/filesystem": patch
```
it's just a bug fix right?

thetarnav (approved):

PR #803, opened 2025-07-01
`virtual` Improvements (Dynamic Item Sizes, Fix Broken Reactivity)

jaydevelopsstuff:
This adds support for dynamic item sizes in the `virtual` package.

This is accomplished by adding `| (row: T[number], index: number) => number)` to the `rowHeight` parameter, and type-checking for either case within `createVirtualList`.
```typescript
type VirtualListConfig<T extends readonly any[]> = {
    // ...
    rowHeight: MaybeAccessor<number | ((row: T[number], index: number) => number)>;
};
```
I also added `scrollToItem` to `VirtualListReturn` as a utility function for scrolling to a specified item index.

I'm implementing this because I need it for use in another project, so if I end up needing more features I'll implement them (but I'm also happy to work on other stuff if it's requested). This is an early stage PR and my first crack back at SolidJS in almost half a year after working on other projects, so I'm very rusty and probably didn't implement all of this idiomatically—please correct me if necessary.

*(Also contains a small patch to add `type` to an import in `site/src/primitives/DocumentHydrationHelper.tsx`, since the dev server would not run without that change)*

### DIFF
--- a/packages/virtual/dev/index.tsx
+++ b/packages/virtual/dev/index.tsx
@@ -4,19 +4,25 @@ import { VirtualList } from "../src/index.jsx";
 
 const intl = new Intl.NumberFormat();
 
-const items = new Array(100_000).fill(0).map((_, i) => i);
-
+const items = new Array(100_000).fill(0).map((_, i) => [i, Math.random() * 72 + 24]);
 const clampRange = (min: number, max: number, v: number) => (v < min ? min : v > max ? max : v);
 
 const App: Component = () => {
+  const [dynamicSize, setDynamicSize] = createSignal(true);
   const [listLength, setListLength] = createSignal(100_000);
   const [overscanCount, setOverscanCount] = createSignal(5);
   const [rootHeight, setRootHeight] = createSignal(240);
   const [rowHeight, setRowHeight] = createSignal(24);
+  let scrollToItem!: (itemIndex: number) => void;
 
   return (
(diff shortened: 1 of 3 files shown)


jaydevelopsstuff:
I tried to provide a way to access `scrollToItem` from the `VirtualList` component, but after doing a bit of searching I couldn't seem to find any documentation on idiomatic ways to do this. So I implemented this wildly hacky method which I'm hoping someone can help me replace with something more practical:
```typescript
type VirtualListProps<T extends readonly any[], U extends JSX.Element> = {
    // ...
    setScrollToItem: (scrollToItem: (itemIndex: number) => void) => void;
};

// Inside `VirtualList`
let scrollContainer!: HTMLDivElement; // Attached Ref
props.setScrollToItem((itemIndex: number) => scrollToItem(itemIndex, scrollContainer));
```
In Use:
```tsx
let scrollToItem!: (itemIndex: number) => void;
// ...
<VirtualList
    // ...
    setScrollToItem={fn => (scrollToItem = fn)}
    // ...
>
{/* ... */}
</VirtualList>
```

jaydevelopsstuff:
Alright, I fixed the issue with reactivity being broken by not unwrapping accessors at the top level of `createVirtualList`.

atk (changes requested):

davedbase:
@jaydevelopsstuff and @atk can we push this effort to completion or consider adapting it to our Solid Primitives 2.0 effort? Would love to see these improvement land.

PR #815, opened 2025-10-14
feat(pagination): new segmentation primitive, jump pages, fix set page to non-existent page, fix demo

atk:


thetarnav (approved):
I like it

PR #831, opened 2026-02-21
Add SSE primitive and transform utilities

davedbase:
This PR adds server-sent events to the Solid Primitives project. I developed this for a project and went a bit further with it by adding optional Worker and built-in transform utilities. The goal was to make the package composable and flexible for many uses. I rather like the idea of the transforms and might add that into the websocket package as well.


davedbase:
> it's interesting to include the "transforms" here
> that aren't related strictly to sse, they are just normal functions
> but are small enough that it's probably fine

I mean I can not include them or move them somewhere more central. Technically I believe websocket could leverage them.

Would it make sense to put them in utils?

PR #842, opened 2026-04-15
resize-observer: add box model option to createElementSize

james-tindal:
Pass the box model option to ResizeObserver.prototype.observe.

```ts
const size = createElementSize(element, 'border-box')
```


james-tindal:
Actually this doesn't even do anything. 🤔

PR #845, opened 2026-04-20
upgrade: intersection-observer package upgrade for Solid 2.0

davedbase:
- Update peer deps to solid-js 2.0.0-beta.7
- Replace onMount/Setter with createStore + pending state (NOT_SET sentinel)
- Curried add(callback)(el) form for AddViewportObserverEntry
- Remove deprecated makeIntersectionObserver and JSX directive
- WeakMap for element→callback tracking, full test suite for new API


packages/intersection-observer/src/index.ts
```diff
@@ -99,62 +91,109 @@ export function makeIntersectionObserver(
...
-): void {
-  if (isServer) return;
+  options?: MaybeAccessor<IntersectionObserverInit>,
+): readonly [
+  entries: readonly IntersectionObserverEntry[],
+  isVisible: (el: Element) => boolean,
+] {
+  if (isServer) {
+    const isVisible = (_el: Element): boolean => {
+      throw new NotReadyError("IntersectionObserver not available on server");
```
atk: Is that really the way to go? This way, we require error boundaries or extra guards around isVisible for isomorphic components, which is less than ideal. I would prefer having isVisible return false in all cases on the server, since that is a correct answer (nothing is visible until it reaches the client and gets rendered).

packages/intersection-observer/src/index.ts
```diff
@@ -99,62 +91,109 @@ export function makeIntersectionObserver(
...
+    });
+  }
+
+  createEffect(
+    () => elements(),
+    (list: Element[], prev: Element[] = []) => {
+      handleDiffArray(list, prev, el => observe(el, io), el => io.unobserve(el));
+      trackedEls = list;
+    },
+    [] as Element[],
```
atk: since you already use a default parameter, this is superfluous, isn't it?

PR #848, opened 2026-04-21
upgrade: fullscreen package upgrade for Solid 2.0

davedbase:
Bumps the peer dependency to `solid-js@^2.0.0-beta` and updates the package to match the 2.0 API.

**Breaking**
- `use:createFullscreen` directive removed — use `ref={fullscreen(opts)}` instead
- Requires `solid-js@^2.0.0-beta` and `@solidjs/web@^2.0.0-beta`

**Changes**
- New `fullscreen()` ref factory export as the Solid 2.0 replacement for the directive
- `isServer` imported from `@solidjs/web`
- `createEffect` updated to the split compute/effect signature

## Release Notes

* **Breaking Changes**
  * Requires Solid.js 2.0 (beta.14+); not compatible with Solid 1.x.
  * Fullscreen API updated; usage pattern changed to `ref` directive.

* **New Features**
  * Enhanced fullscreen primitives with improved state tracking and control methods.

* **Documentation**
  * README updated with complete API reference and usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/848?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/fullscreen/src/index.ts
```diff
@@ -1,72 +1,133 @@
...
+ *
+ * @example
+ * ```ts
+ * const [enter, exit] = makeFullscreen(ref);
+ * button.addEventListener("click", enter);
+ * ```
+ */
+export const makeFullscreen = (
+  element: HTMLElement,
+  options?: FullscreenOptions,
```
atk, line 22: This API no longer makes sense, since we no longer have directives. It would make more sense to return two setters, one for the element and one for the behavior (options | true = enter full screen, undefined | false = exit full screen):

```ts
export function makeFullscreen() {
  const [ref, setRef] = createSignal<HTMLElement>();
  const [options, fullscreen] = createSignal<FullscreenOptions | boolean>(false);
  createEffect(() => [ref, options], ([ref, options]) => {
    if (ref && options !== false) {
      ref.requestFullscreen(options === true ? undefined : options);
    } else {
      document.exitFullscreen();
    }
  });
}
```

atk (changes requested):

PR #850, opened 2026-04-21
upgrade: clipboard package upgrade for Solid 2.0

davedbase:
Migrates the clipboard package to Solid.js 2.0

### Breaking changes

- **`createClipboard` now returns an async accessor** — `createResource` (removed in 2.0) is replaced with an async `createMemo`. The returned `clipboard` accessor starts synchronously with an empty array; after `refetch()` it reads from the clipboard asynchronously. The previous value stays visible during a pending read (`isPending(() => clipboard())` returns `true`).

- **`copyToClipboard` directive → ref function** — Solid 2.0 no longer supports the `use:` directive namespace extension pattern. `copyToClipboard` is now a factory that returns a `ref` callback:
  ```tsx
  // before
  <input use:copyToClipboard />

  // after
  <input ref={copyToClipboard()} />
  <button ref={copyToClipboard({ value: "copy me" })}>Copy</button>
  ```
  The `declare module "solid-js"` JSX directive augmentation has been removed accordingly.

- **`isServer` import** — moved from `solid-js/web` → `@solidjs/web`.

- **Peer dependency** bumped to `solid-js@^2.0.0-beta`.

### Internal

- Extracted `readClipboardItems()` async helper (replaces inline `on`+`createResource` pattern).
- Fixed a long-standing typo in the element tag check: `"texfield"` → ``"textarea"`.
- `copyToClipboard` now uses `@ts-expect-error` instead of `@ts-ignore`.


packages/utils/src/index.ts
```diff
@@ -13,7 +13,9 @@ import {
   DEV,
   equalFn,
 } from "solid-js";
-import { isServer } from "solid-js/web";
+// isServer moved from solid-js/web (1.x) to @solidjs/web (2.x).
+// typeof window is a universal fallback compatible with both versions.
+const isServer = typeof window === "undefined";
```
atk, line 24: maybe `globalThis.window !== globalThis` instead?

packages/clipboard/test/index.test.ts
```diff
@@ -1,27 +1,43 @@
 import "./setup";
-import { createComputed, createRoot } from "solid-js";
-import { createClipboard } from "../src/index.js";
+import { createRoot, flush, resolve } from "solid-js";
+import { createClipboard, type ClipboardResourceItem } from "../src/index.js";
 import { describe, expect, it } from "vitest";
 
+// flush/resolve are Solid 2.0 APIs; they don't exist in 1.x.
```
atk, line 6: Not sure if that is the right way. We should aim for a full 2.0 supported version.

packages/clipboard/README.md
```diff
@@ -8,55 +8,41 @@
...
+- [`writeClipboard`](#writeclipboard) - One-shot async write to the clipboard.
+- [`createClipboard`](#createclipboard) - Reactive clipboard backed by an async memo. Reads on demand, writes reactively from a signal.
+- [`copyToClipboard`](#copytoclipboard) - `ref` directive factory that copies an element's value to clipboard on click.
 
 ## Installation
 
 ```bash
 npm install @solid-primitives/clipboard
 # or
-yarn add @solid-primitives/clipboard
```
atk, line 23: Why not support yarn anymore?

atk (approved):

packages/clipboard/README.md
davedbase: lol not sure why that was removed tbh

PR #851, opened 2026-04-21
upgrade: storage package upgrade for Solid 2.0

atk:
- update to Solid 2.0
- simplified setter wrapping
- simpler types using function overloads
- TODO: figure out optimistic/actions - can we even reasonably support those and if yes, how?

## Summary by CodeRabbit

* **New Features**
  * Added package detail pages displaying documentation, installation info, and package metadata
  * Added interactive playground pages for testing primitives in the browser

* **Chores**
  * Updated storage package to support Solid 2.0 beta
  * Updated Tauri dependencies to stable versions

* **Refactor**
  * Simplified `makePersisted` API with improved type inference and setter behavior


packages/storage/src/persisted.ts
```diff
@@ -241,7 +249,7 @@ export const wsSync = (ws: WebSocket, warnOnError: boolean = isDev): Persistence
         key,
         newValue,
         timeStamp: +new Date(),
-        ...(isServer ? {} : { url: location.href }),
+        ...(globalThis.window ? { url: location.href } : {}),
```
davedbase, line 260: I still am not sure why we wouldn't just use isServer?

packages/storage/src/persisted.ts
```diff
@@ -167,28 +175,28 @@ export function makePersisted<
...
-          options.sync?.[1](name, value);
-          storage.setItem(name, value, storageOptions);
-          unchanged = false;
-        },
+    signal[0], 
+    (value: any) => untrack(() => {
+      const output = signal[1](value);
+      persist();
+      unchanged = false;
+      return output instanceof Promise ? output.then((result) => (persist(), result)) : output;
```
davedbase: persist() is called synchronously to write the optimistic value, then again in .then() on line 197. There is no catch so if the action rejects, Solid rolls back the in-memory signal, but storage is never re-synced. It permanently holds the pre-rejection optimistic value. On the next page load the signal hydrates with stale data?

packages/storage/src/persisted.ts
```diff
@@ -241,7 +249,7 @@ export const wsSync = (ws: WebSocket, warnOnError: boolean = isDev): Persistence
         key,
         newValue,
         timeStamp: +new Date(),
```
davedbase, line 259: Isn't Date.now() more the idiomatic form because it avoids creating an unnecessary date object?

packages/storage/src/persisted.ts
```diff
@@ -58,70 +55,81 @@ export type PersistenceSyncAPI = [
...
+  const actionFn = options.action && options.action(signal);
+  if (actionFn) {
+    (signal as any)[1] = action(actionFn) as unknown as S[1];
+  }
   if (!storage) {
-    return [signal[0], signal[1], null] as PersistedState<S>;
+    return [signal[0], signal[1], null] as unknown as PersistedState<S>;
   }
   const storageOptions = (options as unknown as { storageOptions: O }).storageOptions;
   const serialize: (data: T) => string = options.serialize || JSON.stringify.bind(JSON);
```
davedbase, line 144: I was confused by the bind, neither method uses this...so isn't this just allocating a pointless closure? Maybe I missed something? lol

packages/storage/src/persisted.ts
```diff
@@ -167,28 +175,28 @@ export function makePersisted<
     });
   }
 
+  const getter = typeof signal[0] === "function" ? signal[0] as () => T : () => snapshot(signal[0] as T);
```
davedbase: You're going a few of these evaluations on line 140, 166 and 179. Maybe just have `const isSignal = ...`?

packages/storage/src/persisted.ts
```diff
@@ -1,8 +1,5 @@
-import type { Accessor, Setter, Signal } from "solid-js";
-import { createUniqueId, untrack } from "solid-js";
-import { isServer, isDev } from "solid-js/web";
-import type { SetStoreFunction, Store } from "solid-js/store";
-import { reconcile } from "solid-js/store";
+import type { Signal, StoreSetter, Store } from "solid-js";
+import { action, createUniqueId, latest, untrack, reconcile, snapshot, DEV } from "solid-js";
```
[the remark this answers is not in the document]
atk: On the contrary; reconcile and snapshot are exported from `"solid-js"` starting with version 2.

packages/storage/src/persisted.ts
```diff
@@ -135,16 +143,16 @@ export function makePersisted<
...
           try {
             const value = deserialize(data);
-            (signal[1] as any)(reconcile(value));
+            (signal[1] as any)(reconcile(value, () => undefined));
           } catch (e) {
             // eslint-disable-next-line no-console
-            if (isDev) console.warn(e);
+            if (DEV) console.warn(e);
           }
         };
```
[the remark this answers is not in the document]
atk: reconcile works very different in Solid-2.0 - we will have to recreate the previous functionality if we don't want to over-react when setting stores.

site/src/routes/package/$name/index.tsx~port to tanstack start v1
```diff
@@ -0,0 +1,87 @@
...
+import { PackageInstallation } from "./-components/package-installation.js";
+import { createPrimitiveNameTooltips } from "./-components/primitive-name-tooltips.js";
+
+export const Route = createFileRoute("/package/$name/")({
+  component: PackagePage,
+  loader: async ({ params }) => fetchPackageData(params.name),
+  head: ({ params }) => ({
+    meta: [{ title: `${kebabCaseToCapitalized(params.name)} — Solid Primitives` }],
+  }),
+});
```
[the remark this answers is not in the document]
atk: This is something that should be done on a global level for all packages and is out of scope for this PR.

packages/storage/src/persisted.ts
atk: changed to isServer

packages/storage/src/persisted.ts
atk: This is an interesting discussion. Should we persist what the user intended (even if the server did not accept it for whatever reason)? Though I just reached the opinion that handling server errors should be handled outside of persist, so I will add a catch branch that persists again and propagates the error.

packages/storage/src/persisted.ts
atk: Good point.

packages/storage/src/persisted.ts
atk: In some implementations of JSON, the main object is an instance that needs to be in scope for stringify and parse to work; it will otherwise error out if called out of scope. IIRC, that was the case in an slightly older version of Safari and Opera.

packages/storage/src/persisted.ts
atk: Reduced it to two, but created a named variable nevertheless.

PR #853, opened 2026-04-30
Added createRcMemo primitive

clinuxrulz:
`createRcMemo` is similar to `createLazyMemo` except its lifetime is bounded by the existence of listeners. (the listeners are reference counted.)

- Initially when there are no listeners, the internal memo is not created.
- When it goes from 0 listeners, to some listeners the internal memo is constructed.
- When it goes from some listeners, to zero listeners the internal memo is deconstructed.
- The internal memo can be reconstructed again after it has been deconstructed.
- For the context being passed along, it will carry the context of the owner of which `createRcMemo` was called inside.

## Summary by CodeRabbit

* **New Features**
  * Introduced `createRcMemo` - a reference-counted memo primitive that automatically manages computational resources by executing only when active listeners exist, and cleaning up when all listeners are removed.

* **Documentation**
  * Added comprehensive `createRcMemo` documentation with usage examples and behavior details.

### DIFF
--- a/packages/memo/src/index.ts
+++ b/packages/memo/src/index.ts
@@ -428,10 +428,15 @@ export function createLazyMemo<T>(
  * Reference counted `createMemo`. The memo calculation will only run when there is at least one listener.
  *
  * Once the number of listeners drops to zero, the internal memo will be disposed after a microtask.
- * If a new listener is added later, the memo will be re-constructed.
+ * If a new listener is added later, the internal memo will be re-constructed.
  *
  * Unlike `createLazyMemo`, the internal memo's lifetime is managed by the number of listeners,
  * and it will stop updating when there are no listeners.
+ *
+ * It can be created outside of a reactive root, as it manages its own internal owner.
+ * If it is created inside a reactive root, it will carry the context of that root.
+ *
+ * **Note:** Because the internal memo is disposed when there are no listeners, the `prev` value in the calculation function will become stale if the internal memo is reconstructed.
  *
  * @param calc pure reactive calculation returning some value
  * @param value initial value for the calculation
@@ -482,14 +487,14 @@ export function createRcMemo<T>(
   const owner = getOwner();
   //
   return () => {
-    if (getListener() == null) {
+    if (getListener() === null) {
       if (existing === undefined) {
         return calc(undefined);
       } else {
         return existing.memo();
       }
     } else {
-      if (existing == undefined) {
+      if (existing === undefined) {
         runWithOwner(owner, () => {
           existing = createRoot((dispose) => {
             return {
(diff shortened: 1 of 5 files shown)


packages/memo/src/index.ts
```diff
@@ -423,6 +424,109 @@ export function createLazyMemo<T>(
   };
 }
 
+/**
+ * Reference counted `createMemo`. The memo calculation will only run when there is at least one listener.
+ *
+ * Once the number of listeners drops to zero, the internal memo will be disposed after a microtask.
```
atk, line 430: Please add that the `prev` argument to the memo will be based on the previous value after being without listeners and being re-initialized, even if intermediate computations would have led to a different value. Maybe consider providing an additional stale argument (or similar mechanism) to the handler.

packages/memo/src/index.ts
```diff
@@ -423,6 +424,109 @@ export function createLazyMemo<T>(
...
+
+export function createRcMemo<T>(
+  calc: (prev: T | undefined) => T,
+  value?: T,
+  options?: EffectOptions,
+): Accessor<T> {
+  if (isServer) {
+    let calculated = false;
+    return () => {
+      if (!calculated) {
```
atk, line 475: Document the server behaviour.

packages/memo/src/index.ts
```diff
@@ -423,6 +424,109 @@ export function createLazyMemo<T>(
...
+            };
+          });
+        });
+      } else {
+        existing.refCount++;
+      }
+      let existing2 = existing!;
+      onCleanup(() => {
+        existing2.refCount--;
+        if (existing2.refCount == 0) {
```
atk: Please use type-safe comparisons unless you consciously want to match other types, e.g. in line 485.

atk (changes requested):
There is no documentation in the README.md, and this also needs a changeset file (please run `npx changeset` and add a minor version). Lastly, add createRcMemo to package.json in the primitive.list array.

clinuxrulz:
@atk I believe I have address your requested changes.
Except for `npx changeset`, that command did not make any file changes on my machine, so I am a little confused.

Maybe you can try the `npx changeset` command on your end?

Regarding `prev` to clear things up. When an internal memo is disposed later on, and reconstructed again later later on, the `prev` value will not reset to the `initial` value provided to `createRcMemo`, it will instead carry on the last-seen `prev` value. If it seems unclear in the documentation still, I can try and clear it up.

I've used that `createRcMemo` pattern a lot in production code at work. I believe it's a handy one to have available.

clinuxrulz:
@atk changeset added. Apologies for before, I did not understand how the `npx changeset` worked.

atk (approved):

davedbase:
@clinuxrulz I'm going to merge this in but our 2.0 implementation wont have the changes. I'll try to port it with a new PR later this evening.

PR #854, opened 2026-05-03
upgrade: keyboard package upgrade for Solid 2.0

davedbase:
- Bumped peer deps to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`
- Replaced `isServer` import from `solid-js/web` → `@solidjs/web`
- Removed deprecated tuple API from `useKeyDownList` (the `[keys, { event }]` destructuring form)
- Fixed `useKeyDownSequence` — `createMemo` second arg is now options in Solid 2.0, so `prev` starts as `undefined`; handle with `prev ?? []`
- Rewrote `createShortcut` to use fully self-contained imperative state instead of reading reactive signals from event listeners (signal reads outside reactive computations return the pre-batch committed value in Solid 2.0, so shortcut detection and `preventDefault` must be synchronous/imperative)
- Moved `createKeyHold`'s `preventDefault` side effect out of `createMemo` into a dedicated `keydown` listener (memos must not have side effects in Solid 2.0)
- Added `INTERNAL_OPTIONS` (`{ ownedWrite: true }`) to signals written inside singleton roots to satisfy the owned-scope write restriction
- Added `test/server.test.ts` covering all six exported primitives under SSR conditions
- Updated browser tests to call `flush()` after dispatching keyboard events (Solid 2.0 batches signal writes; `flush()` commits them before assertions)


packages/keyboard/src/index.ts
```diff
@@ -222,10 +200,12 @@ export const useKeyDownSequence = /*#__PURE__*/ createSingletonRoot<Accessor<str
...
   const keys = useKeyDownList();
 
-  return createMemo(prev => {
+  // createMemo's second arg is options (not initialValue). The prev
+  // parameter starts as undefined; handle it with a fallback.
+  return createMemo((prev: string[][] | undefined) => {
     if (keys().length === 0) return [];
-    return [...prev, keys()];
-  }, []);
+    return [...(prev ?? []), keys()];
```
atk, line 207: I would prefer a default argument for prev as a cleaner solution.

atk (approved):

PR #856, opened 2026-05-03
upgrade: tween package upgrade for Solid 2.0

davedbase:
- isServer` import moved from `solid-js/web` → `@solidjs/web`
- Replaced `createEffect(on(target, handler, {defer: true}))` with the Solid 2.0 two-argument `createEffect(() => target(), handler, {defer: true})` — `on()` was removed in v2
- RAF cancellation now returned from the apply function instead of `onCleanup` (Solid 2.0 pattern)
- `current()` inside the apply callback wrapped in `untrack()` to suppress strict-mode warnings
- Test suite updated: `flush()` added after signal writes and RAF flushes to account for Solid 2.0's microtask batching
- Added `test/server.test.ts` (SSR safety verification)
- Peer deps bumped to `solid-js@^2.0.0-beta.10` + `@solidjs/web@^2.0.0-beta.10


packages/tween/src/index.ts
```diff
@@ -53,17 +53,15 @@ export default function createTween(
...
-        delta = target() - startValue;
-        cancelId = requestAnimationFrame(tick);
-        onCleanup(() => cancelAnimationFrame(cancelId));
-      },
-      { defer: true },
-    ),
+    () => target(),
+    newTarget => {
+      start = performance.now();
+      startValue = untrack(current);
```
atk: do we need untrack here? I thought effect callbacks were untracked by default?

atk (approved):

packages/tween/src/index.ts
davedbase: Good call. Fixing.

PR #857, opened 2026-05-03
upgrade: scroll package upgrade for Solid 2.0

davedbase:
- Migrate `@solid-primitives/scroll` to Solid.js v2.0 (beta.10). Updates `isServer` import to `@solidjs/web`, replaces `onMount`/`sharedConfig.context` with their Solid 2.0 equivalents, and rewrites the internal scroll-trigger mechanism to avoid the new `createSignal(fn)` derived-signal semantics. Tests updated to use `flush()` for batched-update assertions.
- Adds `createPreventScroll` to `@solid-primitives/scroll` — prevents scrolling outside a given element by intercepting `wheel` and `touchmove` events and optionally hiding the `<body>` scrollbar
- Splits the existing `src/index.ts` into `src/scrollPosition.ts` and `src/preventScroll.ts`; `index.ts` is now a clean barrel re-export
- Migrates the package to Solid 2.0 beta.12 (`@solidjs/web`, `flush()` in tests, `ownedWrite`, split `createEffect`)

## Attribution

`createPreventScroll` is adapted from [solid-prevent-scroll](https://github.com/corvudev/corvu/tree/main/packages/solid-prevent-scroll) by **Jasmin Noetzli (GiyoMoon)**, part of the [corvu](https://corvu.dev) project, which is itself inspired by [react-remove-scroll](https://github.com/theKashey/react-remove-scroll) by **Anton Korzunov**. Both are credited in `LICENSE` and `package.json`.

## `createPreventScroll` API

```ts
createPreventScroll({
  element?: MaybeAccessor<HTMLElement | null>,
  enabled?: MaybeAccessor<boolean>,               // default true
  hideScrollbar?: MaybeAccessor<boolean>,          // default true
  preventScrollbarShift?: MaybeAccessor<boolean>,  // default true
  preventScrollbarShiftMode?: MaybeAccessor<"padding" | "margin">, // default "padding"
  restoreScrollPosition?: MaybeAccessor<boolean>,  // default true
  allowPinchZoom?: MaybeAccessor<boolean>,         // default false
});
```

Multiple active instances stack; only the topmost installs event listeners. Body styles are shared and only restored once all instances clean up.


packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+
+import { createEffect, createSignal } from "solid-js";
+import { isServer } from "@solidjs/web";
+import { contains, INTERNAL_OPTIONS, access, type MaybeAccessor } from "@solid-primitives/utils";
+
+type Axis = "x" | "y";
+
+export type CreatePreventScrollProps = {
+  /** Element that is allowed to scroll. Wheel/touch events inside it are passed through. *Default = `null`* */
+  element?: MaybeAccessor<HTMLElement | null>;
```
jer3m01: Could we change this to use `undefined` instead.

packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+
+function getScrollAtLocation(
+  location: HTMLElement,
+  axis: Axis,
+  stopAt?: HTMLElement,
+): [number, number] {
+  const directionFactor =
+    axis === "x" && window.getComputedStyle(location).direction === "rtl" ? -1 : 1;
+
+  let currentElement: HTMLElement | null = location;
```
jer3m01: `undefined`

packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+  } while (currentElement && !wrapperReached);
+
+  return [availableScroll, availableScrollTop];
+}
+
+function wouldScroll(
+  target: HTMLElement,
+  axis: Axis,
+  delta: number,
+  wrapper: HTMLElement | null,
```
jer3m01: `undefined`

packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+ * Adapted from [solid-prevent-scroll](https://github.com/corvudev/corvu/tree/main/packages/solid-prevent-scroll)
+ * by Jasmin Noetzli (GiyoMoon), part of the corvu project.
+ */
+export const createPreventScroll = (props: CreatePreventScrollProps = {}): void => {
+  if (isServer) return;
+
+  const id = String(_nextId++);
+
+  let currentTouchStart: [number, number] = [0, 0];
+  let currentTouchStartAxis: Axis | null = null;
```
jer3m01: `undefined`

packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+  let currentTouchStartDelta: number | null = null;
```
jer3m01: `undefined`

packages/scroll/src/preventScroll.ts
```diff
@@ -0,0 +1,323 @@
...
+      };
+    },
+  );
+
+  // 3. Install wheel/touch event listeners (only when this is the topmost instance).
+  createEffect(
+    () => ({
+      active: isActive(id),
+      enabled: access(props.enabled) ?? true,
+      wrapper: access(props.element) ?? null,
```
jer3m01: `undefined`

packages/scroll/src/scrollPosition.ts
```diff
@@ -0,0 +1,114 @@
+import { createEventListener } from "@solid-primitives/event-listener";
+import { createHydratableSingletonRoot } from "@solid-primitives/rootless";
+import { createDerivedStaticStore } from "@solid-primitives/static-store";
+import { type Accessor, createSignal, onSettled, sharedConfig } from "solid-js";
+import { isServer } from "@solidjs/web";
+
+export function getScrollParent(node: Element | null): Element {
```
jer3m01: `undefined`

packages/scroll/README.md
```diff
@@ -119,6 +120,42 @@ createEffect(() => {
...
+### How to use it
+
+```ts
+import { createPreventScroll } from "@solid-primitives/scroll";
+
+// Prevent all page scroll (no element specified)
+createPreventScroll();
+
+// Prevent scroll outside a specific element
+createPreventScroll({ element: () => myElement });
```
jer3m01, line 138: We should showcase using a signal ref.

`createPreventScroll({ element: refAccessor });`

jer3m01 (commented):
Looks good, few things I'd change by preference. Also make sure to use signal refs.

packages/scroll/src/preventScroll.ts
davedbase: MaybeAccessor<HTMLElement | undefined>?

packages/scroll/README.md
davedbase: It's on line 144 in the README

packages/scroll/src/scrollPosition.ts
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const FALLBACK_SCROLL_POSITION = { x: 0, y: 0 } as const satisfies Position;
+
+/**
+ * Get an `{ x: number, y: number }` object of element/window scroll position.
+ */
+export function getScrollPosition(target: Element | Window | undefined): Position {
+  if (isServer || !target) {
+    return { ...FALLBACK_SCROLL_POSITION };
```
atk: Why putting the fallback in a new object every time? It is already read-only according to TS. Why not just return it as is?

atk (approved):

packages/scroll/src/scrollPosition.ts
davedbase: I think the logic was to copy the object in case it's modified, it doesn't effect the default. But that should be up to the user. Fixed.

PR #858, opened 2026-05-03
upgrade: resize-observer package upgrade for Solid 2.0

davedbase:
Migrates `@solid-primitives/resize-observer` to Solid.js v2.0 (beta.10). Updates peer/dev deps, converts both `createEffect` usages to the Solid 2.0 split compute/apply pattern, moves `isServer` to `@solidjs/web`, and fixes the hydration guard (`sharedConfig.hydrating`). Tests updated to use `flush()` for explicit effect scheduling, `onSettled` in place of `onMount`, and the new store setter function API.


packages/resize-observer/src/index.ts
```diff
@@ -88,11 +88,12 @@ export function createResizeObserver<T extends Element>(
...
 
-  createEffect((prev: T[]) => {
-    const refs = filterNonNullable(asArray(access(targets)));
-    handleDiffArray(refs, prev, observe, unobserve);
-    return refs;
-  }, []);
+  createEffect(
+    () => filterNonNullable(asArray(access(targets))),
+    (refs: T[], prev?: T[]) => {
+      handleDiffArray(refs, prev ?? [], observe, unobserve);
```
atk: This no longer returns refs so you don't get `prev`. Maybe there is a missing test case or prev was never really used??

packages/resize-observer/src/index.ts
```diff
@@ -177,21 +178,25 @@ export function createElementSize(
...
-      if (el) {
-        setSize(getElementSize(el));
-        ro.observe(el);
-        onCleanup(() => ro.unobserve(el));
-      }
-    });
+    createEffect(
+      () => (target as Accessor<Element | false | undefined | null>)(),
+      (el: Element | false | null | undefined) => {
+        if (el) {
```
atk, line 193: Maybe use an early return instead to simplify the structure:

```ts
if (!el) return setSize(ELEMENT_SIZE_FALLBACK);

setSize(getElementSize(el));
ro.observe(el);
return () => ro.unobserve(el);
```

packages/resize-observer/src/index.ts
davedbase: Fascinating. I added a test to verify if it works and it seems to checkout. I'll explore this a bit more tomorrow...you should be right though.

atk: Then we should probably remove prev.

packages/resize-observer/test/index.test.ts
```diff
@@ -112,17 +118,93 @@ describe("createResizeObserver", () => {
...
+    _targets = new Set<Element>();
+    _observe_calls = [];
+    const { dispose, setRefs } = createRoot(dispose => {
+      const [refs, setRefs] = createSignal([div1, div2]);
+      createResizeObserver(refs, () => {});
+      return { dispose, setRefs };
+    });
+    flush();
+    expect(_observe_calls.length).toBe(2);
+    expect(_observe_calls.filter(t => t === div1).length).toBe(1);
```
atk, line 197: better use `expect(_observe.calls.some(t => t === div1)).toBeTrue()`

atk (approved):

PR #862, opened 2026-05-03
upgrade: refs package upgrade for Solid 2.0

davedbase:
Updated peer/dev dependencies to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`
- `isServer` import moved from `solid-js/web` → `@solidjs/web`
- `createComputed` (removed in Solid 2.0) replaced with split `createEffect(compute, apply)` in both `Refs` and `Ref` components; `untrack()` calls removed as the apply phase is implicitly untracked
- Removed deprecated type re-exports (`ResolvedChildren`, `ResolvedJSXElement`) that were marked for deletion in the next major
- Tests: added `flush()` after signal writes, `{ ownedWrite: true }` on test signal, `onMount` → `onSettled` in mergeRefs test, `renderToString` import updated to `@solidjs/web`
- Added a package-local `vitest.config.ts` setting `moduleName: "@solidjs/web"` so JSX in test files resolves to the correct Solid 2.0 runtime (the shared config doesn't set this per-package)


packages/refs/test/mergeRefs.test.tsx
```diff
@@ -1,23 +1,31 @@
...
-        dispose();
-      });
-    }));
+  test("ignores undefined refs", () => {
+    const el = document.createElement("button") as HTMLButtonElement;
+    let called = false;
+    const merged = mergeRefs<HTMLButtonElement>(undefined, e => {
+      called = true;
+    });
+    merged(el);
```
atk, line 28: Shouldn't we check that the callback is *not* called for without an actual element?

packages/refs/vitest.config.ts
atk, line 1: Instead of a local vitest config for every new primitive, wouldn't it be better to have a vitest-solid2.config.ts in the root that can be used by multiple primitives and then overwrite the previous one?

davedbase: Yeah, I meant to do that but then forgot. Maybe you can do it on the next branch? Re-aligning them later isn't a problem

PR #863, opened 2026-05-03
upgrade: mutation-observer package upgrade for Solid 2.0

davedbase:
- Upgrades `@solid-primitives/mutation-observer` to Solid.js v2.0-beta.10.
- Replaces removed `onMount` with `onSettled` so observation begins after the component fully settles (async-aware)
- Switches `isServer` import to `@solidjs/web`, removing the manual `typeof window` check
- Guards `start()` against missing DOM globals in Node.js environments
- Adds `test/server.test.ts` to cover SSR safety
- Bumps peer dependencies to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10


packages/mutation-observer/src/index.ts
```diff
@@ -34,27 +35,26 @@ export type E = JSX.Element;
...
  *   { attributes: true, subtree: true },
  *   records => console.log(records)
  * );
- * 
+ *
  * // Usage as a directive
  * const [mutationObserver] = createMutationObserver([], e => {...})
-
-<div use:mutationObserver={{ childList: true }}>...</div>
+ * <div use:mutationObserver={{ childList: true }}>...</div>
```
atk: There are no more directives in Solid 2.0. So the use in the future is probably more like

```ts
const [mutationObserver] = createMutationObserver([], e => {...});

<div ref={(el) => mutationObserver(el, () => ({ childList: true }))}>...</div>
```

Either that or we move the options into the create call so the mutationObserver can directly be used as ref function.

atk (changes requested):
We need to adapt the JSDoc comment to Solid-2.0 - maybe even consider changing the API so the options are moved into the creation instead of the directive.

packages/mutation-observer/src/index.ts
davedbase: Ah I just forgot to replace the directive! Done now.

atk (approved):

PR #871, opened 2026-05-04
upgrade: pagination package upgrade for Solid 2.0

davedbase:
Migrate `@solid-primitives/pagination` to Solid.js v2.0 beta.10. Replaces `createResource` (removed) with a manual fetch + cancellation effect, replaces `createComputed` (removed) with a derived memo for page clamping, moves `isServer` to `@solidjs/web`, and removes `batch()`.

## Release Notes

* **Breaking Changes**
  * Updated to require Solid.js v2.0-beta.10 with `@solidjs/web` peer dependency
  * `createInfiniteScroll` no longer exposes `pages.loading` and `pages.error` properties
  * `isServer` must now be imported from `@solidjs/web`
  * `createInfiniteScroll` return type: `loader` is now a ref function instead of a directive

* **Documentation**
  * Updated README with peer dependency requirements and revised API usage examples

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/871?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/pagination/src/index.ts
```diff
@@ -137,28 +135,25 @@ export const createPagination = (
   options?: MaybeAccessor<PaginationOptions>,
 ): [props: Accessor<PaginationProps>, page: Accessor<number>, setPage: Setter<number>] => {
   const opts = createMemo(() => Object.assign({}, PAGINATION_DEFAULTS, access(options)));
-  const [page, _setPage] = createSignal(opts().initialPage || 1);
+  // ownedWrite allows setPage to be called from event handlers and reactive scopes
+  const [rawPage, setRawPage] = createSignal(opts().initialPage || 1, { ownedWrite: true });
```
atk: I don't like this change. When wrapping a setter, I either use the underscore notation or call it something like updatePage, so that page and the wrapped setPage are still the pair they are supposed to be. On the other hand, I wanted to create a wrapSetter primitive to wrap the setter of any signal/store - that could be used here, too.

packages/pagination/src/index.ts
```diff
@@ -396,28 +389,34 @@ export function createInfiniteScroll<T>(fetcher: (page: number) => Promise<T[]>)
...
-
-  createComputed(() => {
-    const content = contents.latest;
-    if (!content) return;
-    batch(() => {
-      if (content.length === 0) setEnd(true);
-      setPages(p => [...p, ...content]);
-    });
-  });
+    createEffect(
```
atk, line 396: Why not use createMemo?

packages/pagination/src/index.ts
atk: Update: if you rebase, there's now `wrapSetter` in utiils.

atk (changes requested):

PR #876, opened 2026-05-05
upgrade: virtual package upgrade for Solid 2.0

davedbase:
- Bumps peer/dev dependencies to `solid-js@2.0.0-beta.10` and `@solidjs/web@2.0.0-beta.10`
- `createVirtualList` now returns `[Accessor<VirtualState>, onScroll]` — the first element is reactive and must be called to read state
- `VirtualList` children signature updated to `(item: Accessor<T>, index: Accessor<number>)` matching Solid 2.0's `<For>` behavior
- Internal scroll offset signal uses `{ ownedWrite: true }` to allow writes from the scroll event handler
- Adds a per-package `vitest.config.ts` using a custom Babel plugin with `babel-preset-solid@2.0.0-experimental.16` to fix a JSX transform incompatibility (`vite-plugin-solid@2.11.12` generates single-argument `_$effect()` calls which are incompatible with `@solidjs/web`'s two-argument `effect()` requirement)
- Overrides `jsxImportSource: "@solidjs/web"` in `tsconfig.json` so TypeScript resolves JSX types from the correct package


packages/virtual/package.json
```diff
@@ -48,17 +48,21 @@
...
   "peerDependencies": {
-    "solid-js": "^1.6.12"
+    "@solidjs/web": "^2.0.0-beta.10",
+    "solid-js": "^2.0.0-beta.10"
   },
   "devDependencies": {
-    "solid-js": "^1.9.7"
+    "@babel/core": "^7.27.0",
+    "@solidjs/web": "2.0.0-beta.10",
+    "babel-preset-solid": "2.0.0-experimental.16",
```
brenelz: is the experimental the proper version?

davedbase: Nope, good eye should be beta 10 as well

PR #877, opened 2026-05-06
upgrade: broadcast-channel package upgrade for Solid 2.0

davedbase:
Updates peer dependencies to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`. Moves `isServer` import to `@solidjs/web`, updates tests to the split `createEffect(compute, apply)` pattern, and removes use of the now-removed `on` helper from README examples.

### DIFF
--- a/packages/broadcast-channel/test/index.test.ts
+++ b/packages/broadcast-channel/test/index.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect, beforeAll, afterAll } from "vitest";
-import { createEffect, createRoot } from "solid-js";
+import { createEffect, createRoot, flush } from "solid-js";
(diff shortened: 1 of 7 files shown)


packages/broadcast-channel/test/index.test.ts
```diff
@@ -223,20 +223,25 @@ describe("createBroadcastChannel", () => {
     const dispose = createRoot(dispose => {
       const { message } = createBroadcastChannel<MsgType>(channelName);
 
-      createEffect(() => {
-        captured.push(message());
-      });
+      createEffect(
+        () => message(),
```
atk: just use `createEffect(message, (msg) => captured.push(msg))`.

atk (approved):

PR #878, opened 2026-05-06
upgrade: pointer package upgrade for Solid 2.0

davedbase:
Updates peer dependencies to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`. Moves `isServer` to `@solidjs/web`, fixes `JSX` import (now from `@solidjs/web` where `JSX.Element` lives in Solid 2.0), adds `flush()` calls in the reactive-target test to account for deferred `createEffect` scheduling, and adds a new `server.test.ts` with SSR coverage for all primitives.

### DIFF
--- a/packages/pointer/package.json
+++ b/packages/pointer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@solid-primitives/pointer",
-  "version": "0.3.5",
+  "version": "0.4.0",
   "description": "A collection of primitives, giving you a nicer API to handle pointer events in a reactive context.",
   "author": "Damian Tarnawski @thetarnav <gthetarnav@gmail.com>",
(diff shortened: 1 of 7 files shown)


packages/pointer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@solid-primitives/pointer",
-  "version": "0.3.5",
+  "version": "0.4.0",
```
atk: Why manually change the version? Shouldn't we leave that to changeset?

atk (approved):

packages/pointer/package.json
davedbase: lol that was probably the AI hallucinating :p

PR #879, opened 2026-05-06
upgrade: autofocus package upgrade for Solid 2.0

davedbase:
onMount` replaced by `onSettled` from `solid-js` in the `autofocus` directive
- `createAutofocus` updated to split `createEffect(compute, apply)` form — the apply phase now also clears any pending timeout on re-renders, avoiding stale focus calls
- `JSX` type now imported from `@solidjs/web` (was `solid-js`)
- Tests rewritten to call the directive directly (no JSX compilation) and use explicit `flush()` calls for Solid 2.0's deferred effect scheduling
- Added `test/server.test.ts` for SSR coverage
- Peer dependencies updated to `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10


packages/autofocus/src/index.ts
```diff
@@ -1,59 +1,67 @@
...
- * <button autofocus ref={autofocus}>Autofocused</button>;
+ * // Disable autofocus
+ * <button ref={autofocus(false)} autofocus>Not autofocused</button>
  * ```
  */
-export const autofocus = (element: HTMLElement, autofocus?: Accessor<boolean>) => {
-  if (autofocus?.() === false) {
-    return;
-  }
+export const autofocus = (enabled?: boolean) => {
```
atk: Why would we even support the "enabled===false" use case? You can already disable it by removing the autofocus attribute on the node. I believe it was originally added because directives always had a single accessor argument whether you used it or not, so I would simply remove it and document the decision.

atk (changes requested):

packages/autofocus/src/index.ts
davedbase: Good point. Ok this is now adjusted and documented.

PR #885, opened 2026-05-08
new: orientation primitive for Solid 2.0

davedbase:
- Adds a new `orientation` package (Stage 0 / Sensors) targeting Solid.js 2.0 beta.10
- `makeOrientation(onChange)` — non-reactive, attaches a listener for `screen.orientation` change events and returns a cleanup function
- `createOrientation()` — reactive, returns `angle` and `type` signals initialized to the current screen orientation and updated on each change; SSR-safe (returns static defaults server-side)
- Falls back to the deprecated `window.orientationchange` event on browsers without `screen.orientation`

### DIFF
--- a/packages/orientation/README.md
+++ b/packages/orientation/README.md
@@ -53,10 +53,13 @@ import { createEffect } from "solid-js";
 
 const { angle, type } = createOrientation();
 
-createEffect(() => {
-  console.log(angle()); // 0 | 90 | 180 | 270
-  console.log(type());  // "portrait-primary" | "landscape-primary" | ...
-});
+createEffect(
+  () => ({ angle: angle(), type: type() }),
+  ({ angle, type }) => {
+    console.log(angle); // 0 | 90 | 180 | 270
+    console.log(type);  // "portrait-primary" | "landscape-primary" | ...
+  }
+);
 ```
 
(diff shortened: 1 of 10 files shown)


packages/orientation/src/index.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @returns object with `angle` and `type` signal accessors
+ *
+ * @example
+ * ```ts
+ * const { angle, type } = createOrientation();
+ *
+ * createEffect(() => {
+ *   console.log(angle(), type());
+ * });
```
brenelz: this example is out of date with solid 2.0

packages/orientation/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+On the server, returns static defaults: `angle: 0`, `type: "portrait-primary"`.
+
+```ts
+import { createOrientation } from "@solid-primitives/orientation";
+import { createEffect } from "solid-js";
+
+const { angle, type } = createOrientation();
+
+createEffect(() => {
```
brenelz: also outdated createEffect example

davedbase: Good catch, thanks :)

PR #886, opened 2026-05-08
new: vibrate primitive for Solid 2.0

davedbase:
New `packages/vibrate` package wrapping the [Vibration API](https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API) with six exports following the repo's `make*`/`create*` conventions

- All primitives are SSR-safe (no-ops on server), peer-dependent on `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`
- 31 tests passing (28 browser + 3 SSR), ESLint and tsc clean
- createPulse` uses a ~1000 ms multi-cycle chunk strategy: builds a pattern array covering ~1 second and re-queues it with `setInterval` at the exact chunk duration, so patterns join seamlessly without gaps
- Both reactive primitives use the `createEffect(compute, apply)` split form and a plain boolean mirror (`isVibrating` / `isPulsing`) to avoid `STRICT_READ_UNTRACKED` warnings in Solid 2.0's dev mode
- `dutyCycle` option (default `0.5`) controls the on/off ratio within each pulse cycle — higher values produce longer buzzes, lower values produce shorter taps


packages/vibrate/src/index.ts
```diff
@@ -0,0 +1,302 @@
...
+  supported: boolean;
+} {
+  const supported = isVibrationSupported();
+
+  if (!supported) {
+    return { vibrating: () => false, start: noop, stop: noop, supported };
+  }
+
+  const { interval } = options;
+  const [vibrating, setVibrating] = createSignal(false, INTERNAL_OPTIONS);
```
brenelz, line 101: what is `INTERNAL_OPTIONS`?

brenelz (approved):

packages/vibrate/src/index.ts
davedbase: It injects the owned write property from utils

PR #889, opened 2026-05-10
upgrade: lifecycle package upgrade for Solid 2.0

davedbase:
Upgrades `@solid-primitives/lifecycle` to Solid 2.0 beta.10. All API signatures (`createIsMounted`, `isHydrated`, `onElementConnect`) are unchanged.

**Breaking changes for downstream consumers:**
- Peer dependencies now require `solid-js@^2.0.0-beta.10` and `@solidjs/web@^2.0.0-beta.10`
- Internally: `onMount` → `onSettled`, `getListener` → `getObserver`, `sharedConfig.context` → `sharedConfig.hydrating`, `isServer` from `@solidjs/web

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated to require Solid.js v2.0 (beta.14) and `@solidjs/web` v2.0 (beta.14) as peer dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/889?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/lifecycle/src/index.ts
```diff
@@ -39,7 +39,7 @@ export function createIsMounted(): Accessor<boolean> {
  * @see https://github.com/solidjs-community/solid-primitives/tree/main/packages/lifecycle#isHydrated
  */
 export const isHydrated = (): boolean =>
-  !isServer && (!sharedConfig.context || (!!getListener() && createIsMounted()()));
+  !isServer && (!sharedConfig.hydrating || (!!getObserver() && createIsMounted()()));
```
atk, line 42: Wouldn't that create mutliple isMounted signals in the worst case, since we're async now? Maybe flush() the signal so it is synchronous.

packages/lifecycle/test/index.test.ts
```diff
@@ -1,20 +1,19 @@
 import { describe, test, expect } from "vitest";
-import { createEffect, createRoot } from "solid-js";
+import { createRoot, flush } from "solid-js";
 import { createIsMounted, isHydrated } from "../src/index.js";
 
 describe("createIsMounted", () => {
   test("createIsMounted", () => {
```
atk, line 6: We should test that the isMounted signal is not needlessly created multiple times.

atk (changes requested):

packages/lifecycle/src/index.ts
davedbase: flush doesn't work in `onSettled` or `createTrackedEffect`. I added a test to verify if this is a concern at all. It demonstrates the property: a second `onSettled` registered after `createIsMounted` can already read `isMounted()` as `true`, proving that `setIsMounted(true)` is applied synchronously within the flush which is exactly why `flush()` inside `onSettled` is maybe not needed.

This one is tricky, you may want to study the tests I added more closely and run it directly.

packages/lifecycle/test/index.test.ts
davedbase: Added a few

atk (approved):

PR #890, opened 2026-05-11
upgrade: memo and destructure packages upgrade for Solid 2.0

davedbase:
**@solid-primitives/memo`:**
- `createDebouncedMemo` / `createThrottledMemo` sections: changed from "deprecated" to **"Removed in v2"**, linking to `@solid-primitives/schedule` with a note that it isn't yet updated for Solid 2.0
- Added a new `createAsyncMemo` section: **"Removed in v2"**, explaining that `createResource` no longer exists and pointing to the Solid 2.0 alternatives
- `createLazyMemo` "Not ownerless" section: replaced with **"Auto-disposal and recomputation"** — the v2 native `createMemo({ lazy: true })` behavior, including the note that `prev` resets to `undefined` after a full unsubscribe/resubscribe cycle
- "No `equals`" section: tightened the prose to match the v2 pull-based rationale

This completes the deprecation of create of previously primitives but I'm questioning if they should probably be brought back. Looking for suggestion and confirmation that this is indeed the correct path foward.

**`@solid-primitives/destructure`:**
- Added `pnpm` install line (was missing)
- Added a Solid 2.0 requirement callout
- Documented the two new options (`name`, `equals`) that were previously inherited implicitly from `MemoOptions


packages/destructure/test/destructure.test.ts
```diff
@@ -40,249 +41,217 @@ describe("destructure", () => {
     dispose();
   });
 
-  test("spread array", () =>
-    createRoot(dispose => {
-      const [numbers, setNumbers] = createSignal([1, 2, 3] as [number, number, number]);
-      const [first, second, last] = destructure(numbers);
+  test("spread array", () => {
+    const [numbers, setNumbers] = createSignal([1, 2, 3] as [number, number, number]);
```
atk, line 45: Use a const assertion to make this a tuple with less work.

packages/destructure/src/index.ts
```diff
@@ -112,7 +114,7 @@ export function destructure<T extends ReactiveSource, O extends DestructureOptio
     const calc = getter(key);
     if (config.deep && isReactiveObject(value))
       result[key] = destructure(calc, { ...config, memo });
-    else result[key] = memo ? createMemo(calc, undefined, options) : calc;
+    else result[key] = memo ? createMemo(calc, options as any) : calc;
```
atk: once MemoOptions is exposed again, we should definitely use the correct type.

atk: Update the current version exposes it again.

atk (changes requested):
Please use MemoOptions

atk (approved):

PR #899, opened 2026-05-15
new: controlled-signal package for Solid 2.0

davedbase:
Summary

- Adds a new `@solid-primitives/controlled-signal` package porting Kobalte's `createControllableSignal` to Solid 2.0 Beta 12
- Fixes pre-existing Solid 2.0 compatibility gaps in `@solid-primitives/utils`
- Credits original Kobalte contributors in `package.json`, `LICENSE`, and `README`

### New package: `controlled-signal`

Implements the controlled/uncontrolled state pattern used by headless UI components — when an external `value` prop is provided, the primitive defers to it and fires `onChange`; otherwise it manages its own internal state from `defaultValue`.

**Exports:**
- `createControllableSignal<T>(props)` — base primitive
- `createControllableBooleanSignal(props)` — coerces `undefined` to `false`
- `createControllableArraySignal<T>(props)` — coerces `undefined` to `[]`
- `createControllableSetSignal<T>(props)` — coerces `undefined` to `new Set()`

**Solid 2.0 adaptations:**
- `{ ownedWrite: true }` on the internal signal (required when setter is callable from reactive owner scopes)
- `createMemo` for control-mode detection (replaces `createComputed`, removed in 2.0)
- Raw signal read inside `setValue` to avoid deferred-memo staleness
- `accessWith` from `@solid-primitives/utils` for the value-or-updater function pattern

### DIFF
--- a/packages/controlled-signal/src/index.ts
+++ b/packages/controlled-signal/src/index.ts
@@ -1,0 +1,91 @@
+/*
+ * Adapted from @kobaltedev/kobalte
+ * https://github.com/kobaltedev/kobalte/tree/main/packages/core/src/primitives/create-controllable-signal
+ * Copyright (c) 2022 Kobalte Contributors — MIT License
+ */
+
+import { type Accessor, createMemo, createSignal, untrack } from "solid-js";
+import { accessWith } from "@solid-primitives/utils";
+
+export interface CreateControllableSignalProps<T> {
+  /** The value to be used in controlled mode. */
+  value?: Accessor<T | undefined>;
+
+  /** The initial value to be used in uncontrolled mode. */
+  defaultValue?: Accessor<T | undefined>;
+
+  /** Called when the value changes. */
+  onChange?: (value: T) => void;
+}
+
+/**
+ * Creates a reactive signal that can operate in either controlled or uncontrolled mode.
+ *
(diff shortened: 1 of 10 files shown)


packages/controlled-signal/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Variant of {@link createControllableSignal} for Set values.
+ * Falls back to `new Set()` when the value is `undefined`.
+ */
+export function createControllableSetSignal<T>(props: CreateControllableSignalProps<Set<T>>) {
+  const [_value, setValue] = createControllableSignal(props);
+  const value: Accessor<Set<T>> = () => _value() ?? new Set();
+  return [value, setValue] as const;
+}
```
jer3m01, line 91: imo wouldn't include these, I never really use them and can just write a generic controllable signal with a bool/other as value.

jer3m01 (changes requested):
Could remove some duplicate functions

packages/controlled-signal/src/index.ts
davedbase: Other people might have a need for them. You can focus on just suing `createControlledSignal` for your purposes. No harm in leaving the rest in there.

jer3m01 (approved):

PR #900, opened 2026-05-15
new: interaction package for Solid 2.0

davedbase:
Adds a new package for detecting and hiding interactions outside a given element — a common requirement for accessible dismissible UI (dropdowns, dialogs, popovers, tooltips).

### Primitives

- **`createInteractOutside`** — registers document-level `pointerdown` and `focusin` listeners and fires `onPointerDownOutside`, `onFocusOutside`, and `onInteractOutside` custom events when the user interacts outside a referenced element. Each event wraps the original DOM event and can be cancelled with `preventDefault()`.

### Solid 2.0 adaptations

- Split `createEffect(compute, apply)` throughout — single-arg form is not supported in beta.12.
- Cleanup is returned from `apply` rather than using `onCleanup` (which has no owner context in `apply`).
- `ref` is tracked inside the `compute` function so `createInteractOutside` re-activates when an element mounts into a `<Show>` after initial render.
- Custom `vitest.config.ts` with `moduleName: "@solidjs/web"` to avoid the missing `solid-js/web` specifier.
- Tests written as plain `.ts` (no JSX) using `document.createElement` + `createRoot` to avoid the `use is not a function` error from `ref=` attributes in `@solidjs/web`.

### Credits

Based on [`createInteractOutside`](https://github.com/kobaltedev/kobalte/blob/main/packages/core/src/primitives/create-interact-outside/create-interact-outside.ts) from Kobalte, which itself draws from Radix UI, Zag (Chakra UI), and React Spectrum (Adobe).


packages/interaction/src/index.ts
```diff
@@ -0,0 +1,432 @@
...
+  };
+
+  observerStack.push(observerWrapper);
+
+  return () => {
+    observer.disconnect();
+
+    for (const node of hiddenNodes) {
+      const count = refCountMap.get(node);
+      if (count == null) return;
```
davedbase: @jer3m01 should this be a return statement or continue? If it's a return the remaining nodes wont be cleaned up?

jer3m01: Yeah `continue` seems safer.

packages/interaction/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ *     { onInteractOutside: () => setOpen(false) },
+ *     () => ref,
+ *   );
+ *
+ *   return <Show when={open()}><div ref={ref}>Popover content</div></Show>;
+ * }
+ * ```
+ */
+export function createInteractOutside<T extends Element>(
+  props: CreateInteractOutsideProps,
```
jer3m01: Tiny nitpick, maybe call this options? Not sure if there's a convention in solid primitives.

packages/interaction/README.md
```diff
@@ -0,0 +1,176 @@
...
+## Usage
+
+### Dismiss a popover on outside interaction
+
+```tsx
+import { createSignal, Show } from "solid-js";
+import { createInteractOutside } from "@solid-primitives/interaction";
+
+function Popover() {
+  let ref: HTMLDivElement | undefined;
```
jer3m01: Signal ref instead

jer3m01 (approved):
Looks good!

packages/interaction/src/index.ts
davedbase: As long as it's consistent across the board, I'm fine with being picky :)

packages/interaction/src/index.ts
```diff
@@ -0,0 +1,255 @@
...
+  isContextMenu: boolean;
+};
+
+export type PointerDownOutsideEvent = CustomEvent<EventDetails<PointerEvent>>;
+export type FocusOutsideEvent = CustomEvent<EventDetails<FocusEvent>>;
+export type InteractOutsideEvent = PointerDownOutsideEvent | FocusOutsideEvent;
+
+export interface CreateInteractOutsideOptions {
+  /** Whether the interact outside events should be listened or not. */
+  isDisabled?: MaybeAccessor<boolean | undefined>;
```
jer3m01: rename to `disabled`?

jer3m01 (approved):
LGTM!

PR #907, opened 2026-05-21
upgrade: context package upgrade for Solid 2.0

davedbase:
Migrates `@solid-primitives/context` to Solid.js 2.0 (beta.13) and adds three quality-of-life additions to the API.

---

## Solid 2.0 migration (breaking)

- **`Context.Provider` removed** — In Solid 2.0, the `Context` object itself is the provider component. `createContextProvider` and `MultiProvider` updated; `<MyCtx.Provider value={...}>` becomes `<MyCtx value={...}>`.
- **`ContextProviderComponent` import fixed** — Previously imported from an internal `node_modules` path; now a proper export from `solid-js`.
- **`JSX.Element` → `Element`** — Public API types now use `Element` from `solid-js` (renderer-neutral), matching the Solid 2.0 type model.
- **`MultiProvider` cleanup** — Removed the `.Provider` fallback branch; `Context` is always a function in Solid 2.0.
- **Peer deps**: `solid-js@^2.0.0-beta.13` and `@solidjs/web@^2.0.0-beta.13` required.
- **Version bumped to `1.0.0`** (major).

**Noteworthy Solid 2.0 behaviour:** context providers now wrap children in a lazy memo (`children(() => props.children)`). Tests that previously used bare `createRoot(() => { <Provider/>; })` never forced child evaluation. The `MultiProvider` browser test was updated to use `render()` so the DOM renderer drives evaluation of the lazy memo chain.

---

### `createStrictContextProvider`

Like `createContextProvider` without defaults, but with the contract made explicit in types: the hook returns `T` (never `T | undefined`), and Solid throws `ContextNotFoundError` at runtime when called outside a provider. Accepts an optional `{ name }` for DevTools labeling.

```tsx
const [AuthProvider, useAuth] = createStrictContextProvider(
  () => {
    const [user, setUser] = createSignal<User | null>(null);
    return { user, setUser };
  },
  { name: "Auth" },
);

// No `!` needed — type is T, throws if no provider
const { user } = useAuth();
```

### `createLayeredContext`

Each provider extends the parent context value rather than replacing it. The factory receives `(props, parent)` where `parent` is the nearest enclosing provider's value (falling back to `defaults` at the root). You control the merge strategy.

```tsx
const [ThemeProvider, useTheme] = createLayeredContext(
  (props: { primary?: string }, parent) => ({
    ...parent,
    primary: props.primary ?? parent.primary,
  }),
  { primary: "blue", secondary: "gray" },
);

<ThemeProvider primary="red">            // { primary: "red",   secondary: "gray" }
  <ThemeProvider primary="green">        // { primary: "green", secondary: "gray" }
    <App />
  </ThemeProvider>
</ThemeProvider>
```

Useful for themes, permission layers, i18n patches, and any context where child providers should inherit what they don't explicitly override.

### `ContextProviderOptions` + `name` on `createContextProvider`

All three primitives accept `{ name?: string }` for the Solid DevTools Symbol label. On `createContextProvider` it's the third argument (with-defaults overload only):

```ts
createContextProvider(factory, defaultTheme, { name: "Theme" })

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added `createStrictContextProvider` for strict context creation with error handling outside providers
  * Added `createLayeredContext` for nested context inheritance
  * Added optional debug naming for context providers to improve developer tooling and error traces

* **Breaking Changes**
  * Requires Solid.js v2.0 beta
  * Context objects now used directly as provider components instead of `.Provider`
  * Updated type system from `JSX.Element` to `Element`

* **Documentation**
  * Updated installation and usage guides for new primitives
  * Updated examples for MultiProvider configuration

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/907?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: packages/context/src/index.ts [hunk 1]
```diff
@@ -36,21 +46,67 @@ export type ContextProvider<T extends ContextProviderProps> = (
...
 export function createContextProvider<T, P extends ContextProviderProps>(
   factoryFn: (props: P) => T,
 ): [provider: ContextProvider<P>, useContext: () => T | undefined];
 export function createContextProvider<T, P extends ContextProviderProps>(
   factoryFn: (props: P) => T,
   defaults?: T,
+  options?: ContextProviderOptions,
 ): [provider: ContextProvider<P>, useContext: () => T | undefined] {
-  const ctx = createContext(defaults);
+  const ctx = createContext<T | undefined>(defaults, options);
```
atk: Why do we introduce `undefined` here? It is no longer necessary.

packages/context/src/index.ts
```diff
@@ -36,21 +46,67 @@ export type ContextProvider<T extends ContextProviderProps> = (
...
+ * const [AuthProvider, useAuth] = createStrictContextProvider(() => {
+ *   const [user, setUser] = createSignal<User | null>(null);
+ *   return { user, setUser };
+ * }, { name: "Auth" });
+ *
+ * // throws if used outside <AuthProvider>
+ * const { user } = useAuth();
+ * ```
+ */
+export function createStrictContextProvider<T, P extends ContextProviderProps>(
```
atk: Solid's contexts are now strict by default

atk (changes requested):

packages/context/src/index.ts
davedbase: Good call.

PR #909, opened 2026-05-21
new: async package for Solid 2.0

atk:
This should replace the resource package, featuring the following primitives:

- `fromStream`/`fromJSONStream`: (small helper to aggregate web streams); maybe in the future, we can also support fromJSONpStream for progressive JSON or similar formats
- `makeAbortable`/`createAbortable`: the same as in the resource package, but chainable for actions
- `makeRetrying`
- `makeCache` (-> will go into the fetch package)

## Summary by CodeRabbit

* **New Features**
  * Introduced `@solid-primitives/async` package featuring stream utilities for incremental data processing, abort/signal management with timeout support, automatic retry mechanisms with configurable delays, and reactive value aggregation capabilities.
  * Added comprehensive documentation and interactive Storybook examples demonstrating practical usage patterns.

* **Documentation**
  * Updated resource package documentation with migration guidance for Solid 2.0.


packages/async/src/index.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export type AbortableReturn = [
+  signal: () => AbortSignal,
+  abort: (reason?: string) => void,
+  filterAbortError: (err: any) => void,
+]
+
+export type AbortableOptions = {
+  noAutoAbort?: boolean;
```
davedbase: I'd call this `autoAbort`, you don't need no in front of it, it's verbose.

packages/async/src/index.ts
```diff
@@ -0,0 +1,123 @@
...
+    const reader = stream?.getReader();
+    if (!reader) {
+      console.warn('No ReadableStream found!')
+      return;
+    }
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) return;
+      if (value) {
+        parts += typeof value === 'string' ? value : String.fromCharCode(value);
```
davedbase: Why not use `TextDecoder`?

const decoder = new TextDecoder();
parts += decoder.decode(value, { stream: true });

packages/async/src/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+  if (options.chainTo) {
+    chained.set(options.chainTo, [...(chained.get(options.chainTo) || []), () => abort("chain abort")]);
+  }
+  function signal() {
+    if (!options.noAutoAbort && controller?.signal.aborted === false) {
+      abort("retry");
+    }
+    controller = new AbortController();
+    if (!controller) { throw new Error('AbortController is not supported!'); }
```
davedbase: Shouldn't you perform the check before initializing the class?

packages/async/src/index.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Options are optional and include:
+ * - `timeout`: time in Milliseconds after which the fetcher aborts automatically
+ * - `noAutoAbort`: can be set to true to make a new source not automatically abort a previous request
+ * - `chainTo`: listen to another abort signal to abort this signal
+ */
+export function makeAbortable(
+  options: AbortableOptions = {},
+): AbortableReturn {
+  let controller: AbortController;
+  let timeout: NodeJS.Timeout | number | undefined;
```
davedbase: Using `NodeJS.Timeout` leaks a Node type into a browser-targeting package. Should be `ReturnType<typeof setTimeout>`.

packages/async/test/server.test.ts
```diff
@@ -0,0 +1,9 @@
+import { describe, test, expect } from "vitest";
+import { createPrimitiveTemplate } from "../src/index.js";
+
+describe("createPrimitiveTemplate", () => {
```
davedbase: Change createPrimitiveTemplate name :)

packages/async/src/index.ts
```diff
@@ -0,0 +1,266 @@
...
+ * - `delay` - number of Milliseconds to wait before retrying; default is 5s
+ * - `retries` - number of times a request should be repeated before giving up throwing the last error; default is 3 times
+ */
+export function makeRetrying<T>(
+  fetcher: (v?: T) => PromiseLike<T> | AsyncIterable<T> | T,
+  options: RetryOptions = {},
+): () => AsyncGenerator<T, void> {
+  const delay = options.delay ?? 5000;
+  let retries = options.retries || 3;
+
```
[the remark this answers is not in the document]
atk: Repetition in this case is better for compression. Even though 0 retires make no sense and the option is not reactive, I use the nullish coalescing operator now.

packages/async/src/index.ts
```diff
@@ -0,0 +1,266 @@
...
+        }
+        if (delay) await new Promise<void>(resolve => setTimeout(resolve, delay));
+      }
+    }
+  };}
+
+
+function toArray(item: any) {
+  return Array.isArray(item) ? item : item ? [item] : [];
+}
```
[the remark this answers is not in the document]
atk: This is 100% intentional.

packages/async/src/index.ts
```diff
@@ -0,0 +1,268 @@
...
+    }, []).join("");
+
+/**
+ * **aggregates web stream chunks into a memo supporting partial JSON**
+ * ```ts
+ * // from Response:
+ * const streamed = createMemo(fromStream(() => fetch(url())));
+ *
+ * // from another ReadableStream:
+ * const streamed = createMemo(fromStream(() => getStream()));
```
brenelz: these jsdocs should say "fromJSONStream"?

packages/async/src/index.ts
```diff
@@ -0,0 +1,268 @@
...
+ * ```ts
+ * const data = createMemo(makeRetrying(() => fetch(url()), { retries: 5 }));
+ * ```
+ * Receives the fetcher and an optional options object and returns a wrapped fetcher that retries on error after a delay multiple times.
+ *
+ * The optional options object contains the following optional properties:
+ * - `delay` - number of Milliseconds to wait before retrying; default is 5s
+ * - `retries` - number of times a request should be repeated before giving up throwing the last error; default is 3 times
+ */
+export function makeRetrying<T>(
```
brenelz, line 193: this is super cool

brenelz (approved):

packages/async/src/index.ts
atk: Thanks!

packages/async/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "type": "module",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "browser": {},
+  "exports": {
+    "import": {
+      "@solid-primitives/source": "./src/index.ts",
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
```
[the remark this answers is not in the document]
atk: This is meant to allow typescript to directly access code changes even before a build script is started.

PR #911, opened 2026-05-22
upgrade: set package upgrade for Solid 2.0

davedbase:
Upgrades `@solid-primitives/set` to Solid.js v2.0 (beta.13) and ships five new exports.

### Migration

- Removed `batch()` wrappers from `ReactiveSet.add()`, `delete()`, and `clear()` — Solid 2.0 auto-batches all signal writes, so the sequential `dirty()` calls are already grouped.
- Updated test suite to use the split `createEffect(compute, apply)` form with `flush()` calls between mutations and assertions, replacing the removed `createComputed`.
- Added `test/server.test.ts` verifying both classes behave as plain non-reactive data structures on the server.
- Peer dependencies updated to `solid-js@^2.0.0-beta.13` and `@solidjs/web@^2.0.0-beta.13`.

### New exports

| Export | Description |
|---|---|
| `union(a, b)` | Elements in `a`, `b`, or both |
| `intersection(a, b)` | Elements present in both `a` and `b` |
| `difference(a, b)` | Elements in `a` not in `b` |
| `symmetricDifference(a, b)` | Elements in either set but not both |
| `readonlySet(set)` | Type-level cast from `ReactiveSet` to `ReadonlySet` |

All four algebra functions return `Accessor<ReadonlySet<T>>` backed by `createMemo`. `intersection` and `difference` use per-key tracking on `b` (`b.has(v)` inside the memo) rather than subscribing to `b`'s `$KEYS` signal — meaning adding an element to `b` that has no overlap with `a` will not invalidate the memo.

### Breaking changes

Requires Solid.js v2.0. No changes to the `ReactiveSet` or `ReactiveWeakSet` public API.

## Summary by CodeRabbit

* **New Features**
  * Added set algebra operations: `union`, `intersection`, `difference`, and `symmetricDifference` for reactive set computations.
  * Added `readonlySet` helper for type-safe read-only views of reactive sets.

* **Documentation**
  * Revised README with comprehensive exports table and improved examples for all set operations.

* **Chores**
  * Updated to support Solid.js v2.0-beta.13.

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/911?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Monkeylordz:
What do you think of adding the set comparison class methods that call the underlying methods? Ex.
```ts
union(other: ReadonlySet<T>): Set<T> {
	this.#triggers.track($KEYS);
	if (other instanceof ReactiveSet) {
		other.values();
	}
	return super.union(others);
}
```

packages/set/src/index.ts
```diff
@@ -154,3 +148,68 @@ export function createSet<T>(initial?: T[]): SignalSet<T> {
...
   return new ReactiveWeakSet(initial);
 }
+
+/**
+ * Reactive union — elements that appear in `a`, `b`, or both.
+ * Re-derives when either input changes.
+ */
+export function union<T>(a: ReadonlySet<T>, b: ReadonlySet<T>): Accessor<ReadonlySet<T>> {
+  return createMemo(() => {
+    const result = new Set<T>(a);
```
atk: Should be similarly fast: `new Set([...a, ...b])`.

atk (approved):

davedbase:
> What do you think of adding the set comparison class methods that call the underlying methods? Ex.
> 
> ```ts
> union(other: ReadonlySet<T>): Set<T> {
> 	this.#triggers.track($KEYS);
> 	if (other instanceof ReactiveSet) {
> 		other.values();
> 	}
> 	return super.union(others);
> }
> ```

I think we'll stick with the current pattern for now. Can reconsider a fundamental change like that later.

PR #914, opened 2026-05-22
upgrade: deep package upgrade for Solid 2.0

davedbase:
Dependencies**
- Peer deps bumped to `solid-js@^2.0.0-beta.13` and `@solidjs/web@^2.0.0-beta.13`
- Store APIs (`createStore`, `reconcile`, `snapshot`) moved from `solid-js/store` → `solid-js`
- `isServer` moved from `solid-js/web` → `@solidjs/web`

**`trackStore`** — rewritten to fix two correctness bugs present in the old implementation:
- Now correctly subscribes to **getter** properties by accessing values through the store proxy (previously getters were skipped, so signals read inside them were never tracked)
- Correctly rejects **plain object wrappers** around stores (`{ myStore: store }` is not itself a store, so it's a no-op — `pojo: false` behaviour is preserved)

**`trackDeep`** — `traverse` parameter typed as `unknown` instead of `Store<T>`, which is the correct type for a function that does runtime type-checking and recurses into untyped children

**`store-updates`**
- `unwrap` → `snapshot`; `createRoot` → `runWithOwner(null, ...)` for detached lazy memos
- DEV warning now only fires for non-store arguments client-side, not on every server-side call
- Warning message corrected from stale name `createStoreDelta` → `captureStoreUpdates`
- All unnecessary type assertions removed (`as Static`, `signal!`, `as any` on already-`any` nodes)
- Inline comments added throughout explaining the algorithm, invariants, and non-obvious implementation choices

**Tests**
- All effects converted to split compute/apply form
- `batch()` removed; `flush()` added after mutations
- `unwrap` → `snapshot`; store setters converted to draft form
- `reconcile` calls in tests replaced with direct draft mutations (beta.13 requires a key function argument, which these tests didn't need)

**README** — all examples updated to split `createEffect`, draft setters, and `snapshot

## Release Notes

* **Breaking Changes**
  * Migrated `@solid-primitives/deep` to Solid.js v2.0 (beta.13). Updated peer dependency requirements and API patterns including import sources, store utilities, and effect usage.

* **Documentation**
  * Updated README examples to reflect new Solid v2 patterns and best practices for tracking and capturing store updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/914?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


atk:
Solid now comes with `deep(store)`, and our utils also provide `wrapSetter`, so this package isn't even needed anymore and should be sunsetted.

brenelz (approved):
LGTM! One question is I know solid core ships with a `deep` helper. I am guessing this is different enough from that?

brenelz:
Some conflicts

davedbase:
> LGTM! One question is I know solid core ships with a `deep` helper. I am guessing this is different enough from that?

There are subtle differences in the implementation. I documented it in the README but we can always re-address it in the future.

PR #915, opened 2026-05-22
upgrade: signal-builders package upgrade for Solid 2.0

davedbase:
Upgrades `@solid-primitives/signal-builders` to `solid-js@^2.0.0-beta.13`.

### Changes

**Solid 2.0 compatibility**
- Removed `on()` helper usage in `capitalize` (API removed in 2.0) — replaced with a plain `createMemo`
- Updated tests to create signals outside `createRoot` and call `flush()` after writes, required by Solid 2.0's owned-scope write restriction and async batching model

**Bug fixes**
- `get` and `merge` were missing `createMemo` wrappers — they declared `Accessor<T>` return types but returned plain non-reactive values; this fixes them to actually be reactive

**Package**
- Peer dependency bumped to `solid-js@^2.0.0-beta.13`

**Docs & metadata**
- README rewritten: cleaner usage section, fixed typos (`intiger`, `an float`, `recursing`), collapsed the "Object/Array" category into "Object", clarified `remove` vs `filterOut` distinction, noted shallow-merge behavior on `merge` and setter-function support on `update`, removed the outdated "proof of concept" section
- Source comments cleaned up: removed comments that restated function names; kept only non-obvious behavioral notes
- `package.json`: fixed placeholder author, updated `primitive.list`, added keywords

### Breaking changes
- `solid-js@^2.0.0-beta.13` is now required
- `get()` and `merge()` now return reactive `Accessor<T>` values — code that treated their results as plain values (working around the previous bug) will break

## Summary by CodeRabbit

* **Documentation**
  * Updated package readme with revised usage guidance, installation instructions, and builder reference documentation with expanded coverage examples.

* **Refactor**
  * Migrated to Solid.js v2.0-beta.13; `get` and `merge` now return reactive accessor values (breaking change from previous non-reactive behavior).

* **Tests**
  * Updated test suite for Solid.js v2.0-beta.13 compatibility.

* **Chores**
  * Updated peer dependencies and package metadata.

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/915?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/signal-builders/src/array.ts
```diff
@@ -11,9 +11,6 @@ import * as _ from "@solid-primitives/utils/immutable";
 import { type Accessor, createMemo } from "solid-js";
 import type { MappingFn, Predicate, FlattenArray } from "@solid-primitives/utils/immutable";
 
-/**
- * signal-builder `Array.prototype.push()`
```
atk, line 15: First, instead of removing JSdoc, wouldn't it be better to improve it?

```ts
/**
 * push - pushes an arbitrary number of items onto an immutable list
 * ```ts
 * const list = createSignal([1, 2, 3]);
 * const extendedList = push(list, 4, () => 5);
 * ```
 * List and items can be values or accessors.
 */
``` 

Same goes for all the other functions.

packages/signal-builders/package.json
```diff
@@ -13,14 +13,26 @@
...
+      "push",
+      "filter",
+      "sort",
+      "map",
+      "get",
+      "merge",
+      "update",
+      "add",
+      "clamp",
+      "template"
```
atk, line 25: The full list is

string
float
int
join
lowercase
uppercase
capitalize
substring
template
add
substract
multiply
divide
power
round
ceil
floor
clamp
push
drop
dropRight
filter
filterOut
sort
map
slice
splice
remove
removeItems
concat
flatten
filterInstance
filterOutInstance
omit
pick
get
merge
update

packages/signal-builders/test/index.test.ts
atk, line 1: This is not even testing half of the exported primitives.

atk (changes requested):
The missing tests are not that important, but the JSdoc comments are bad DX and we should add all primitives to the list.

packages/signal-builders/package.json
davedbase: Not putting the full list because it junks up the list on the main README. Only placed the key ones

packages/signal-builders/test/index.test.ts
davedbase: Had AI produce some tests

PR #919, opened 2026-05-23
upgrade: mediastream migration and adaptation upgrade for Solid 2.0

davedbase:
Replaces `@solid-primitives/stream` with a fully Solid 2.0–compatible package. Switches from `createResource` to reactive signals + split `createEffect` for async stream acquisition. `isServer` now imported from `@solidjs/web`. Return types simplified — `mutate`/`refetch` removed; use source accessors for reactivity. Loading/error states handled via `<Loading>` and `<Errored>` from `@solidjs/web`.

**Design changes (Step 1)**
- `createResource` is gone in Solid 2.0 → replaced with `createSignal(INTERNAL_OPTIONS)` + split `createEffect` for async stream acquisition
- An `active` flag in the effect's apply phase prevents a superseded `getUserMedia` call from overwriting a newer stream (race condition fix)
- `createAmplitudeFromStream` `createEffect` converted to split compute/apply form
- The rAF `loop()` starts synchronously inside `createRoot`, so `setAmplitude` needs `INTERNAL_OPTIONS` to avoid `SIGNAL_WRITE_IN_OWNED_SCOPE`

**API changes**
- `createStream` / `createScreen`: return is now `[Accessor<MediaStream | undefined>, { stop, mute }]` — `mutate` and `refetch` removed (reactivity is source-driven)
- `createAmplitudeStream`: second element simplified to `{ stream, stop }`
- `ResourceActions` type export removed

## Summary by CodeRabbit

* **New Features**
  * Introduced `@solid-primitives/mediastream` package for Solid.js v2 with media stream utilities
  * Added primitives for creating and managing audio/video input streams from devices
  * Added screen capture functionality
  * Added microphone amplitude detection and audio analysis
  * Added media permission request handling

* **Documentation**
  * Added comprehensive README with usage examples
  * Added CHANGELOG documenting API changes and migration guidance

[Review Change Stack](https://app.coderabbit.ai/change-stack/solidjs-community/solid-primitives/pull/919?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/mediastream/src/index.ts
```diff
@@ -0,0 +1,304 @@
...
+          deviceId: { exact: device.deviceId },
+        },
+      }
+    : device;
+};
+
+const stopStream = (stream: MediaStream | undefined) =>
+  stream?.getTracks().forEach(track => track.stop());
+
+const muteStream = (stream: MediaStream | undefined, muted?: boolean) =>
```
atk: maybe use `muted = true` as default argument and then `track.enabled = !muted` in line 35. I think that is more obvious than the inverse logic in this version (I know it was me who wrote this in the first version, but now I know better).

atk (approved):

packages/mediastream/src/index.ts
davedbase: No problem, adjusted

PR #921, opened 2026-05-25
feat: added functional Storybook implementation

davedbase:
Summary

- Sets up **Storybook 10** (`storybook-solidjs-vite@10.1.0`) at the repo root as a development and documentation tool for Solid 2.0-migrated primitives
- Stories live colocated with each package at `packages/[name]/stories/*.stories.tsx`, discovered via a single root glob
- Uses the `storybook-solidjs-vite/next` typed API (`definePreview` / `meta.story`) for type-safe stories
- Resolves workspace packages via the `@solid-primitives/source` custom condition (same approach as the `site/` playground), so stories always run against TypeScript source
- Includes two compatibility shims in `viteFinal` for Solid 2.0 beta: a `solid-js/web → @solidjs/web` alias and a transform that patches `addEventListener → addEvent` generated by an older `babel-preset-solid` nested inside `storybook-solidjs-vite`'s bundled `vite-plugin-solid`

### First package: `@solid-primitives/audio`

Four stories covering the full API surface:
- **`createAudio` — reactive player**: track switching via a reactive source signal, `<Loading>` integration for `duration`
- **`createAudio` — reactive source swap**: demonstrates automatic reconnect when the source accessor changes
- **`makeAudio` — non-reactive**: raw `HTMLAudioElement` with manual lifecycle
- **`makeAudioPlayer` — controls**: imperative `play/pause/seek/setVolume` without signals

Audio samples live in `packages/audio/stories/assets/` and are served by Storybook at `/audio/*`.

### Scaffolding

- `template/stories/index.stories.tsx` stub added so every new package generated by `new-package` ships with the stories pattern
- `storybook-static/` added to `.gitignore`

### Usage

```bash
pnpm storybook        # dev server at localhost:6006
pnpm build-storybook  # static build

### CODE REVIEW

File: packages/audio/stories/_helpers.tsx [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+    <span>{Math.round(props.value() * 100)}%</span>
+  </label>
+);
+
+export const SeekSlider = (props: {
+  current: () => number;
+  max: number;
+  onSeek: (t: number) => void;
+}) => (
+  <input
```
atk: Maybe add a label to make this more accessible.

packages/autofocus/stories/_helpers.tsx
atk, line 1: This shouldn't require an extra file, really.

.storybook/ui/primitives.tsx
```diff
@@ -0,0 +1,193 @@
+import type { JSX } from "solid-js";
+import { For } from "solid-js";
+import { colors, font, radii } from "./tokens.js";
+
+type ContainerOptions = { width?: number; minWidth?: number; gap?: string };
+
+export const makeContainer = (options: ContainerOptions | number = {}): Record<string, string> => {
```
atk: Since we're reusing some very simple container settings again and again, why not export them from here:

```ts
export const container360 = makeContainer(360);
```

packages/autofocus/stories/_helpers.tsx
davedbase: Agreed, originally had leaned into using them but then centralized them all so need to go back and clean up the redunancies.

PR #923, opened 2026-05-31
feat(context) [next] - Improve `createContextProvider` types, Add `createOptionalContextProvider`

Monkeylordz:
In Solid 2.0, `useContext` now throws if the value is undefined, which made the existing `createContextProvider` types out-of-date. Is also introduced a gap which we can solve - namely that some contexts may actually want a missing or undefined value. This PR makes the following additions to address these issues:

- Changed `createContextProvider`'s types so that the `useContext` function always returns a non-undefined value
- Added `createOptionalContextProvider` which allows undefined/missing values

## Summary by CodeRabbit

* **New Features**
  * Introduced `createOptionalContextProvider` to handle missing or undefined context values with optional fallback support.
  * Enhanced `createContextProvider` with stronger type safety guarantees for required context scenarios.

* **Documentation**
  * Expanded context API documentation with comprehensive usage examples and detailed behavior specifications for both providers.


Monkeylordz:
@coderabbit review

atk (approved):

PR #926, opened 2026-06-03
new: form and a11y package for Solid 2.0

davedbase:
Summary

- Introduces `createForm<C>` and `toFormData` for Solid 2.0
- Per-field signals (`value`, `error`, `touched`, `pending`), form-level signals (`dirty`, `valid`, `submitting`, `submitted`), and DOM binding via the 2.0 two-phase ref directive pattern
- Sync and async validators are first-class; validators are plain `(value) => string | null | Promise<string | null>` functions — no adapter needed for Zod, Valibot, Arktype, etc.
- `validateOn: "change" | "blur" | "submit"` controls error display timing, per field or form-wide
- Cross-field rules via `form.validate(fn)`; server-side errors via `form.setError(name, msg)`
- SSR-safe: static accessors on the server, no reactive primitives created

### Implementation notes (fixes applied during review)

**Bug fix — async validators fired 3× per keystroke**  
The original architecture called each validator from three code paths per value change: once from the `_rawError` memo, once from the effect's sync-check, and once from the effect's collection loop. For async validators making network requests, this meant 3 API calls per keystroke with only the last result used.

Fixed by pre-classifying validators at setup time (one probe call each with `fc.initial`): sync validators go into a lazy `syncMemo` that is the only reactive subscriber to `value()` and never touches async validators; async validators go into `asyncFns`. The initial async Promises from classification are saved and reused on the first effect run if the value hasn't changed, avoiding a second API call. When async settles and `asyncError()` changes, `_rawError` recomputes by reading signals only — no validator is called. Result: async validators fire **once** per value change.

**`toFormData` — `false` now omitted**  
Previously `false` was coerced to the string `"false"`, which doesn't match HTML's native behavior (unchecked checkboxes are absent from form payloads). `false` is now omitted alongside `null` and `undefined`.

**`reset(newValues?)` overload**  
Accepts an optional partial values object. Named fields adopt the new values as their dirty baseline, making edit-form patterns possible without fighting the initial-value comparison. A version counter forces `dirty` to recompute even when the field value and new baseline are identical (no-op signal write wouldn't invalidate the memo otherwise).

**`setError` / `field.setError`**  
Injects an external (server-side) error into the reactive graph. Appears in `field.error()`, `form.errors()`, and gates `form.valid()`. Cleared automatically when the user edits the field (`setValue` calls the wrapped setter which clears it).

**Documentation fixes**  
- README type definition for `bind` was missing `| HTMLTextAreaElement`  
- `validate()` note incorrectly said "before `form.valid()` is first read" — the version counter handles late registration  
- `errors()` table row now documents the exclusion of cross-field `validate()` errors

## Summary by CodeRabbit

* **New Features**
  * Added a reactive form library with field-level validation, error gating, cross-field rules, async validation, DOM binding, and form submission handling.
  * Added accessibility primitives for form controls including ARIA label/description/error chaining and validation state propagation.
  * Added live region announcements for screen reader messaging.
  * Added utilities for reduced motion preferences detection.

* **Documentation**
  * Added comprehensive README and Storybook stories demonstrating form validation patterns, cross-field rules, async validators, and ARIA form field integration.

* **Tests**
  * Added full test coverage for form and accessibility functionality including SSR compatibility.


davedbase:
@coderabbitai can you please re-review this package

brenelz (approved):

PR #929, opened 2026-06-03
Prevent reconnect loop in ReconnectingWebSocket

### LINKED ISSUE #927: Calling ReconnectingWebSocket.reconnect() triggers a reconnect loop
### Describe the bug

If you call `ReconnectingWebSocket.reconnect()` (like what `makeHeartbeatWS` does), it triggers a reconnect loop. In detail:

- The `reconnect` method just calls `getWS`
- `getWS` first closes the existing WebSocket, which triggers the `close` listener, scheduling another call to `getWS` via `setTimeout`
- New WebSocket is created
- Some time later the timeout fires and calls `getWS` again
- Repeat infinitely

### Minimal Reproduction Link

https://stackblitz.com/edit/github-ktfjsmho?file=src%2FApp.tsx

daveisadork:
Closes #927 

The fix here is to change `getWS` to remove the `close` event listener that schedules `getWS`, preventing a recursive reconnect scenario.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket reconnection stability through refined connection event listener management.


davedbase:
Released as version `@solid-primitives/websocket@1.4.0`. Thanks for the patch :)

PR #934, opened 2026-06-08
upgrade: controlled-props package upgrade for Solid 2.0

davedbase:
Migrates `@solid-primitives/controlled-props` to Solid.js 2.0 beta.14, fixes several pre-existing bugs uncovered during the migration, adds new utilities, and adds Storybook stories.

---

### Breaking Changes

- **Peer deps bumped** — requires `solid-js@^2.0.0-beta.14` and `@solidjs/web@^2.0.0-beta.14`
- **`JSX` type** moved from `solid-js` to `@solidjs/web`
- **Batched writes** — `createControlledProp` setters are now batched; call `flush()` from `solid-js` before reading updated values in tests
- **`ownedWrite: true`** on the internal signal — required so setters can be called from within reactive scopes (component bodies, effects) without the `SIGNAL_WRITE_IN_OWNED_SCOPE` error

---

### Bug Fixes

- **`min`/`max`/`step` silently ignored** — `NumberProp` and `RangeProp` were constructed without forwarding options; fixed by extracting them before the return
- **`SelectProp` stale after `setValue`** — the selected index was computed once with `untrack` at render time, so calling `setValue` after render had no visible effect; replaced with a reactive `createMemo` that tracks `props.value()`
- **`createSignal` type error** — initial value now cast to `Exclude<T, Function>` to satisfy Solid 2.0's new function-form overload for derived signals

---

### New Features

- **`RangeProp`** — new exported component that renders a range slider (`<input type="range">`) with a live value readout; also accessible as a standalone component
- **`step` option** — `TestPropOptions` now exposes `step` and forwards it to both `NumberProp` and `RangeProp

## Release Notes

* **New Features**
  * Added `RangeProp` component for range slider controls
  * Added `step` option to `createControlledProp` for discrete increments

* **Documentation**
  * Updated README with structured examples, control type inference table, and component documentation
  * Added exported field components section

* **Tests**
  * Extended test coverage for range slider and numeric option handling

* **Chores**
  * Updated dependency versions for compatibility
  * Added Storybook stories for feature demonstrations


brenelz (approved):
LGTM!

PR #936, opened 2026-06-09
upgrade: list package upgrade for Solid 2.0

davedbase:
Adds four Storybook stories for `@solid-primitives/list` under `Control Flow/List`: index tracking on reorder, in-place value updates, fallback on empty, and direct `listArray` usage
- Fixes a TypeScript overload error in `src/index.ts` (`T` → `Exclude<T, Function>` cast on `createSignal`)
- Improves JSDoc throughout `src/index.ts` — expanded `listArray` and `List` docs, per-field `ListItem` comments, and plain-English labels on all six reconciliation passes
- Rewrites `README.md` with a `<For>` / `<Index>` / `<List>` comparison table, per-primitive prose and code examples, a "when to use" guide, props table, and reactivity guarantees
- Removes the `dev/` folder; adds the `docs & demos` badge

## Summary by CodeRabbit

* **New Features**
  * Package promoted to stage-3 stable release

* **Documentation**
  * Expanded README with detailed examples, clearer behavior descriptions, and explicit reactivity guarantees (clarifying when signals update during reorders vs value changes)

* **Tests**
  * Added Storybook interactive stories demonstrating List reordering, value updates, fallback rendering, and array mapping functionality


packages/list/test/index.test.tsx
```diff
@@ -114,6 +122,8 @@ describe("List", () => {
 
     const nextArray = [1, 3, 2, 4];
     set(nextArray);
+    flush();
+    flush(); // apply batched index setter writes
```
maciek50322, line 126: I don't think double `flush()` is necessary in tests.
Checked tests without second `flush()` and they pass.

maciek50322 (approved):
Looks good. 
First I thought of just copying new `<For>` and just adding some lines.
Can't build this project probably because I'm on windows or missing something

davedbase:
> Looks good. First I thought of just copying new `<For>` and just adding some lines. Can't build this project probably because I'm on windows or missing something

I just went with a raw porting, if you think it could align more closely with Solid 2.0 that way then we should probably consider doing just that?

maciek50322:
So this `<List>` is based on new `<For>` and has opt-in `recycle` option. The original implementation uses many internal / hidden variables / functions, so I had to use available interface and so expected the performance to be slightly off.

Because it is based on new `<For>`, it also has new `keyed` option that wasn't present in previous `<List>` version. It's not as well tested, but again, because it's based on `<For>`, these cases should be covered.

## Notable changes (compared to `<For>`)

- The `recycle` option can create new array that is filled with unused items indexes in step 1)
  from
  ```ts
  ...
  if (j !== undefined && j !== -1) {
    // fill temp array
  } else {
    this._nodes[i]!.dispose();
  }
  ...
  ```
  to
  ```ts
  ...
  if (j !== undefined && j !== -1) {
    // fill temp array
  } else if (unusedIndexes) {
    unusedIndexes.push(i);
  } else {
    this._nodes[i]!.dispose();
  }
  ...
  ```
- Then there's block of code wrapped in `if`
  ```ts
  if (unusedIndexes) {
    // recycling
  }
  ```
  Where all recycling magic happens, so unused `recycle` option shouldn't affect regular calculation times.
- Types for mapping function arguments (deciding which ones are reactive) are dependent on specific `keyed` and `recycle` values. Described also in readme.
- Skip common prefix and suffix now check first if it's unkeyed. If it is, then there's no need to do equality check
- Also for skip common prefix and suffix: now `compare` function doesn't fit, because recycling introduces case where `this._rows` has value and `this._key` doesn't (specifically case of `<List recycle keyed={true}>`) - the `compare` function would just always return true. So now instead there's check for key function and only when they are equal it is true. The unkeyed case is already handled by previous check.
  ```ts
  for (
    start = 0, end = Math.min(this._len, newLen);
    start < end &&
    (this._unkeyed || // Unkeyed check
      this._items[start] === newItems[start] ||
      (this._rows &&
        this._key !== undefined && // Custom key check
        this._key(this._items[start]) === this._key(newItems[start]))); // no need for compare function
    start++
  )
  ```

## Benchmark results
- The first benchmark would be comparison of implementations (with internal elements vs exposed interface), turns out it's about 5% worse, this is for simple `<For>` and `<List>` (it should act the same).
```
   ✓ create 1000 elements, settle time (2) 1314ms
     name       hz      min      max     mean      p75      p99     p995     p999      rme  samples
   · For   53.4822  16.0815  24.9077  18.6978  22.7646  24.9077  24.9077  24.9077   ±6.48%       27   fastest
   · List  50.9842  15.3637  37.9780  19.6139  22.6410  37.9780  37.9780  37.9780  ±10.52%       26
   
  For - test/index.bench.tsx > create 1000 elements, settle time
    1.05x faster than Recycle List
```

next up 
Keyed For = `<For>`
Keyed List = `<List>`
Unkeyed For = `<For keyed={false}>`
Recycle Keyed List = `<List recycle>`

- Recycling here is the slowest, because besides reordering being slower than setting value (Keyed For vs Unkeyed For), the recycling has index and value signals, which it sets even when one of then did not change - so it does slightly more than just Keyed List. This probably could be optimized. Base keyed `<List>` and `<For>` trading blows here, 3% faster than internal implementation, meaning that it's probably unnecessary to compare them.
```
   ✓ rotate 1000 elements by 10 (4) 2459ms
     name                    hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · Keyed For           670.40  1.3988  3.2136  1.4916  1.4845  2.0679  2.3564  3.2136  ±1.14%      336
   · Keyed List          686.05  1.3863  2.3376  1.4576  1.4503  1.9905  2.1369  2.3376  ±0.81%      344
   · Unkeyed For         894.49  1.0604  1.7042  1.1180  1.1099  1.5909  1.6720  1.7042  ±0.74%      448   fastest
   · Recycle Keyed List  633.37  1.4739  2.6518  1.5789  1.5706  2.5195  2.5556  2.6518  ±1.13%      317   slowest
 
  Unkeyed For - test/index.bench.tsx > rotate 1000 elements by 10
    1.30x faster than Keyed List
    1.33x faster than Keyed For
    1.41x faster than Recycle Keyed List
```
- Now instead of reordering all elements, some are new, some are expired. The Recycled Keyed List is now faster than simple Keyed - turns out reusing elements is faster than destroying and recreating them. For this case Unkeyed still wins.
```
   ✓ roll window of 1000 elements by 10 (4) 2464ms
     name                    hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · Keyed For           496.88  1.7925  5.3934  2.0126  1.9674  4.8854  5.0055  5.3934  ±2.91%      249   slowest
   · Keyed List          501.27  1.8119  5.1038  1.9949  1.9431  4.7821  4.9938  5.1038  ±2.81%      251
   · Unkeyed For         803.89  1.1708  2.0686  1.2439  1.2522  1.7447  1.9058  2.0686  ±0.79%      402   fastest
   · Recycle Keyed List  604.21  1.5541  2.5259  1.6550  1.6388  2.2848  2.3676  2.5259  ±1.04%      303
   
  Unkeyed For - test/index.bench.tsx > roll window of 1000 elements by 10
    1.33x faster than Recycle Keyed List
    1.60x faster than Keyed List
    1.62x faster than Keyed For
 ```
 - Replacing all elements in array - again recreating (just Keyed) is very slow, the recycle does very good as middle ground between keyed and unkeyed.
 ```
   ✓ replace 1000 elements (4) 3384ms
     name                     hz      min      max     mean      p75      p99     p995     p999     rme  samples
   · Keyed For           15.4173  61.6648  77.4411  64.8621  64.3228  77.4411  77.4411  77.4411  ±5.07%       10   slowest
   · Keyed List          15.6481  55.8738  77.7890  63.9055  63.5850  77.7890  77.7890  77.7890  ±6.26%       10
   · Unkeyed For          804.28   1.1789   2.7646   1.2434   1.2266   2.4983   2.5201   2.7646  ±1.25%      403   fastest
   · Recycle Keyed List   683.28   1.3808   2.9671   1.4635   1.4369   2.8339   2.9190   2.9671  ±1.45%      342
   
  Unkeyed For - test/index.bench.tsx > replace 1000 elements
    1.18x faster than Recycle Keyed List
    51.40x faster than Keyed List
    52.17x faster than Keyed For
 ```
 - Replacing and reordering elements on the whole array - again recreating is slow, recycling slightly better, but reordering hurts more than just updating value. Still recycling is a middle ground, but just updating value (unkeyed) is much faster... than reordering.
 ```
   ✓ replace every 2nd element and roll window by 2 of 100 elements (4) 2451ms
     name                      hz     min      max    mean     p75     p99     p995     p999     rme  samples
   · Keyed For             253.01  3.3784  15.4095  3.9524  3.5530  9.8487  15.4095  15.4095  ±7.18%      127   slowest
   · Keyed List            256.97  3.4943  10.2682  3.8916  3.5981  8.6832  10.2682  10.2682  ±5.54%      129
   · Unkeyed For         8,505.62  0.1129   1.8864  0.1176  0.1159  0.1322   0.1386   1.7353  ±1.49%     4253   fastest
   · Recycle Keyed List  1,235.08  0.7787   1.8690  0.8097  0.7982  1.7064   1.7521   1.8690  ±1.07%      618

  Unkeyed For - test/index.bench.tsx > replace every 2nd element and roll window by 2 of 100 elements
    6.89x faster than Recycle Keyed List
    33.10x faster than Keyed List
    33.62x faster than Keyed For
```
- This one is sort of a cherry pick (or maybe a real world scenario?) - similar to previous one, but now elements are heavy - computations on changing value are slow, so recreating is also slow. In this case recycling shines - when reordering is faster than updating value (recycling prioritizes reordering) and changing value is faster than recreating, it easily beats Keyed and (finally) Unkeyed.
```
   ✓ replace every 2nd element and roll window by 2 of 100 heavy elements (4) 2510ms
     name                    hz     min      max    mean     p75      p99     p995     p999     rme  samples
   · Keyed For           127.03  6.8415  15.7202  7.8719  7.1096  15.7202  15.7202  15.7202  ±8.17%       65   slowest
   · Keyed List          128.43  6.7449  15.4422  7.7863  6.9673  15.4422  15.4422  15.4422  ±8.19%       65
   · Unkeyed For         327.39  2.8503   6.1040  3.0544  3.0690   5.6192   6.1040   6.1040  ±1.69%      164
   · Recycle Keyed List  417.31  2.1104  13.9674  2.3963  2.3280   3.8945   3.9654  13.9674  ±4.82%      209   fastest
   
  Recycle Keyed List - test/index.bench.tsx > replace every 2nd element and roll window by 2 of 100 heavy elements
    1.27x faster than Unkeyed For
    3.25x faster than Keyed List
    3.29x faster than Keyed For
 ```
 

## Summed up

It's working. There might be a room for small optimizations (not calling setters for the same values / indexes) and more testing (cases of base `<For>` or custom function key with recycling), but its shape should be final.

# ❗ The recycling is now opt-in ❗
To have the same behavior as `<List>` from previous version, use `<List recycle>`

brenelz (approved):

PR #938, opened 2026-06-09
upgrade: spring package upgrade for Solid 2.0

davedbase:
Migrates the spring package to Solid 2.0 (`beta.14`) and expands the API with several additions.

**Migration changes:**
- Updates peer deps to `solid-js@^2.0.0-beta.14` and `@solidjs/web@^2.0.0-beta.14`
- Adopts `INTERNAL_OPTIONS` / `ownedWrite` pattern for internal signals (animated value and `isAnimating`)
- Replaces `createComputed` (removed in beta.10) with the split `createEffect` form in `createDerivedSpring`
- Uses `isServer` from `@solidjs/web` instead of `solid-js/web`

**API additions:**
- **`makeSpring`** — non-reactive base primitive returning `[value, set, extras, cleanup]`, consistent with the repo's `make*`/`create*` split convention. `createSpring` now wraps `makeSpring` and calls `onCleanup(cleanup)`
- **`isAnimating`** — reactive `Accessor<boolean>` in the `extras` tuple; `true` while the RAF loop is running, always `false` on the server
- **Reactive options** — the `options` argument now accepts `Accessor<SpringOptions>`, read on every frame so physics parameters (e.g. for `prefers-reduced-motion`) update without restarting the spring
- **Promise accumulator fix** — replaced a single `resolve` reference with a `resolvers: (() => void)[]` array so multiple pending `set()` promises all resolve together on settlement rather than orphaning earlier ones

**Docs & stories:**
- README rewritten with full API reference, a physics options table, setter options docs, and examples for all three exports
- Storybook stories added covering the physics playground (reactive options), 2D object interpolation, derived signal tracking, and setter mode comparison
- Dev demo removed (superseded by Storybook)

## Release Notes

* **Breaking Changes**
  * Requires Solid.js v2.0.0-beta.14 and `@solidjs/web` v2.0.0-beta.14
  * `isServer` import path changed to `@solidjs/web`

* **New Features**
  * Added `isAnimating` reactive property to track animation state
  * Spring options now accept reactive accessors for dynamic physics parameters
  * New non-reactive spring primitive for advanced use cases

* **Documentation**
  * Comprehensive README rewrite with detailed API documentation and usage examples
  * Added Storybook stories demonstrating physics playground, object animation, derived tracking, and setter modes

* **Chores**
  * Updated dependencies to Solid v2 beta.14


packages/spring/src/index.ts
```diff
@@ -45,176 +53,316 @@ export type SpringTarget =
...
   const [springValue, setSpringValue] = createSpring(target(), options);
 
-  createEffect(() => setSpringValue(target() as WidenSpringTarget<T>));
+  if (!isServer) {
+    createEffect(
+      () => target(),
+      (value: T) => {
+        setSpringValue(value as WidenSpringTarget<T>);
+      },
+      { defer: true },
```
brenelz, line 363: what does defer true do again?

brenelz (approved):

packages/spring/src/index.ts
davedbase: `defer: true` skips the entire initial execution — both compute and apply — until the tracked value actually changes.

PR #940, opened 2026-06-10
upgrade: mutable package upgrade for Solid 2.0

davedbase:
New files:**
- `stories/tsconfig.json` — extends `.storybook/tsconfig.json` three levels up (matching all other packages)
- `stories/mutable.stories.tsx` — 4 stories, all under `"Reactivity/Mutable"` with `tags: ["autodocs"]` on the file

**Stories created:**
1. `createMutable — edit nested fields` — profile object with nested user sub-object; four buttons mutate individual fields; shows deep-path reactivity
2. `createMutable — array push & splice` — mutable array of tags; push/pop/clear via native array methods; `For` rendering with a fallback for empty state
3. `createMutable — getter / setter` — fullName computed get/set; TextField drafts a new full name and Apply splits it back to firstName/lastName
4. `modifyMutable — grouped update` — cycles through preset addresses using `modifyMutable`; an update counter confirms the call happened; shows multiple fields updating as one named operation

**Removed:**
- `dev/index.tsx` and the `dev/` directory
- `"dev"` script from `package.json`
- Demo section (deployed link + source code link) from README

**Updated README:**
- `[**Docs (Storybook)**]` link added as the first item in the link list, pointing to `reactivity-mutable--docs

## Summary by CodeRabbit

* **New Features**
  * Updated to support Solid.js v2.0-beta.14 with improved auto-batching for signal writes.

* **Breaking Changes**
  * Major version bump (v2.0.0) requires Solid.js v2.0-beta.14+ peer dependency.
  * `modifyMutable` API simplified: modifier functions no longer need to return a value.
  * `produce` and `reconcile` helpers from `solid-js/store` are no longer available; use direct mutation instead.

* **Documentation**
  * Added Storybook stories demonstrating nested mutations, array operations, and batched updates.
  * Updated README with examples reflecting simplified mutation patterns.


.changeset/mutable-solid2-migration.md
```diff
@@ -0,0 +1,46 @@
...
+modifyMutable(state, reconcile({ firstName: "Jake" }));
+modifyMutable(state, produce(s => { s.firstName = "Jake"; }));
+
+// After (v2) — modifier receives the proxy; mutate it directly
+modifyMutable(state, s => {
+  s.firstName = "Jake";
+});
+```
+
+`produce` and the `reconcile`/`produce` helpers from `solid-js/store` no longer exist in Solid 2.0. Use a plain mutation function instead.
```
brenelz: reconcile still exists i think? `produce` doesn't

packages/mutable/src/index.ts
```diff
@@ -279,37 +288,34 @@ export type MutableOptions = { name?: string };
...
   /*
     TODO: improve server noop
     https://github.com/solidjs/solid/issues/1733
   */
   if (isServer) return state;
 
-  // TODO: remove this later
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-  const unwrappedStore = solid_store.unwrap(state || {});
+  const unwrappedStore = state;
```
brenelz, line 298: i know `unwrap` was renamed to `snapshot` do we need that here or is `state` good enough

brenelz (commented):
Looks good. Just a couple quick comments. Nothing major

.changeset/mutable-solid2-migration.md
davedbase: Good catch!

packages/mutable/src/index.ts
davedbase: I could be so wrong but, the `state` is already a mutable proxy, `wrap()` handles it via the `$PROXY` check in the `get` trap (returns `receiver`), so no double-wrapping occurs. No `snapshot` call needed here methinks

PR #969, opened 2026-07-05
enhancement: reimplement createInfiniteScroll

davedbase:
I've gone ahead and reimplemented `createInfiniteScroll` mostly to account for better error handling as referenced in issue https://github.com/solidjs-community/solid-primitives/pull/797. I also noticed we could do a better job of supporting Solid2 capabilities. This PR is meant to explore deeper use and exploration for infinite load pagination with the new API and capabilities.

## Summary by CodeRabbit

* **New Features**
  * Infinite scroll now loads each page independently, with per-page loading, error, and retry states.
  * Added a reset option to restart infinite scrolling from the beginning.
  * Pagination automatically adjusts when the total page count changes.

* **Bug Fixes**
  * Failed page loads no longer stop infinite scrolling permanently; you can retry and continue.
  * Reducing the number of pages now correctly clears pages that are no longer needed.

* **Documentation**
  * Updated pagination examples and API docs to match the new behavior.


packages/pagination/src/index.ts
```diff
@@ -362,49 +364,168 @@ declare module "solid-js" {
...
-    setPages: Setter<T[]>;
+    pageCount: Accessor<number>;
+    setPageCount: Setter<number>;
     end: Accessor<boolean>;
-    fetching: Accessor<boolean>;
-    error: Accessor<unknown>;
+    reset: () => void;
   },
 ] {
+  const initialPageCount = isServer ? 0 : 1;
```
atk: Wouldn't it be preferable to allow for rendering the initial page(s) for SEO reasons and speed during SSR? That would be a nice configuration option.

atk (approved):

packages/pagination/src/index.ts
davedbase: Good idea! Added!

Issue #770, opened 2025-03-10
useUrl


huseeiin:
### Describe The Problem To Be Solved

on the server `getRequestEvent()` has `Request.url` but its not a convenient `URL` object, and on the client we have `location` which is not exactly the same as `URL`

### Suggest A Solution

```ts
import { getRequestEvent, isServer } from "solid-js/web";

function useUrl() {
  return new URL(isServer ? getRequestEvent()!.request.url : location.href);
}
```
should this be added? just a suggestion!

atk:
How about making it reactive in the client (on history change event)?

huseeiin:
> How about making it reactive in the client (on history change event)?

awesome idea!

atk:
But in a composable way, so keep useUrl and add createUrl.

huseeiin:
> But in a composable way, so keep useUrl and add createUrl.

> But in a composable way, so keep useUrl and add createUrl.

```ts
function useUrl() {
  const [currentUrl, setCurrentUrl] = createSignal(
    new URL(isServer ? getRequestEvent()!.request.url : location.href)
  );

  createEffect(() => {
    const handleLocationChange = () => {
      setCurrentUrl(new URL(location.href));
    };

    addEventListener("popstate", handleLocationChange);
    addEventListener("hashchange", handleLocationChange);

    return () => {
      removeEventListener("popstate", handleLocationChange);
      removeEventListener("hashchange", handleLocationChange);
    };
  });

  return currentUrl;
}
```
updated code with help from claude. you can add anything else you want.

atk:
Claude still treats Solid as if it was React. Instead of returning the unsubscriber, you need to wrap it in onCleanup.

huseeiin:
> Claude still treats Solid as if it was React. Instead of returning the unsubscriber, you need to wrap it in onCleanup.

```ts
createEffect(() => {
  const handleLocationChange = () => {
    setCurrentUrl(new URL(location.href));
  };

  addEventListener("popstate", handleLocationChange);
  addEventListener("hashchange", handleLocationChange);

  onCleanup(() => {
    removeEventListener("popstate", handleLocationChange);
    removeEventListener("hashchange", handleLocationChange);
  });
});
```
is this correct?

atk:
Since there are no reactive properties inside the effect, it is actually the same, but onMount is shorter, so go for it.

huseeiin:
[...]

Issue #427, opened 2023-05-07
useLocationHash


jer3m01:
### Describe The Problem To Be Solved

reactively use location hash

### Suggest A Solution

```ts
const locationHash = useLocationHash();

createEffect(() => {
  console.log(locationHash()); // => '#' + string
});
```

https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event

thetarnav:
I was working on a [`url` package](https://github.com/solidjs-community/solid-primitives/pull/77) long time ago, but dropped it because I didn't know what I was doing and solid-router provides similar functionality. But maybe it could be revived with a smaller, more focus scope.
This it `createLocationState` which was kinda doing that: https://github.com/solidjs-community/solid-primitives/blob/url/packages/url/src/location.ts#L113

davedbase:
Double reported here as well https://github.com/solidjs-community/solid-primitives/issues/770

davedbase:
Ok I took @thetarnav work in the previous PR and attempting finishing it off. It's now a proposal for review and suggestions: https://github.com/solidjs-community/solid-primitives/pull/978

PR #978, opened 2026-07-07
new: url primitive for Solid 2.0

davedbase:
Attempts to provide a solution for https://github.com/solidjs-community/solid-primitives/issues/427 and https://github.com/solidjs-community/solid-primitives/issues/770

Adds `@solid-primitives/url`, adapted from the design proposed in the unmerged [PR #77](https://github.com/solidjs-community/solid-primitives/pull/77), rewritten from scratch for Solid 2.0:
  - `createLocationState` / `useSharedLocationState` / `updateLocation` / `setLocationFallback` — reactive `window.location`, with `push`/`replace`/`navigate` setters
  - `createURL` / `ReactiveURL` / `createURLRecord` — a granularly reactive `URL`-like object and store
  - `createSearchParams` / `ReactiveSearchParams` / `createLocationSearchParams` / `useSharedLocationSearchParams` / `getSearchParamsRecord` — a granularly reactive `URLSearchParams`-like object, standalone or bound to the browser's location
- Rebuilt on current Solid-2.0-native primitives instead of the original PR's bespoke utilities: `@solid-primitives/static-store` for the shallow reactive records, `@solid-primitives/trigger`'s `TriggerCache` for `ReactiveSearchParams`' granular per-key tracking, a `createStore(fn, seed)` projection for the dynamic-keyed `SearchParamsRecord`, and a guarded imperative two-way sync between `.search` and `.searchParams` (replacing the removed `createComputed`/`on`)
- Fixed two Solid-2.0-specific correctness bugs surfaced while testing the two-way sync live in Storybook (not caught by the unit suite, since it doesn't render through a host framework's reactive graph): a memo-tainting leak from an unwrapped `this.search` read during `ReactiveSearchParams`'s lazy construction, and a stale post-write read caused by Solid 2's write-batching once `.search` is bound reactively elsewhere
- Adds Storybook stories (`Browser APIs/URL`) and a full test suite (33 tests: 28 client + 5 SSR)
- README documents the API and calls out Solid Router as the better choice for apps that need routing anyway, positioning this package as the lightweight alternative
- Registers the package in the root README's package table and adds a changeset (minor, initial release)

## Summary by CodeRabbit

* **New Features**
  * Added a new URL primitives package with reactive handling for location state, URL values, and search parameters.
  * Introduced shared and SSR-friendly helpers for working with browser location and query strings.
  * Added Storybook examples showcasing URL, location, and search-parameter reactivity.

* **Documentation**
  * Updated the main catalogue and README to include the new URL primitives.
  * Added package-specific documentation and license information.

* **Tests**
  * Added coverage for browser and server-side behavior, including reactivity and URL synchronization.


packages/url/src/common.ts
```diff
@@ -0,0 +1,2 @@
+/** A value to write, or a function deriving it from the previous value. */
+export type SetterValue<Prev, Next = Prev> = Next | ((prev: Prev) => Next);
```
atk: This is not exactly the correct type, since if Next is a function, we always need to use the function format, so the correct type would be:

```ts
export type SetterValue<Prev, Next = Prev> = 
  Next extends Function ? ((prev: Prev) => Next) : Next | ((prev: Prev) => Next);
```

packages/url/src/location.ts
```diff
@@ -0,0 +1,196 @@
...
+export function createLocationState(fallback?: LocationFallbackInit): [
+  state: LocationState,
+  setters: {
+    push: LocationSetter;
+    replace: LocationSetter;
+    navigate: LocationSetter;
+  },
+] {
+  if (isServer) {
+    const state = fallback ? getLocationStateFromFallback(fallback) : locationFallback;
```
atk: Why can't we create a fallback ourselves from the Request Event object that we should have in the server?

```ts
const url = isServer ? new URL(getRequestEvent().request.url) : location;
```

packages/url/test/location.test.ts
```diff
@@ -0,0 +1,105 @@
+import { createRoot } from "solid-js";
+import { describe, test, expect, afterEach } from "vitest";
+import { createLocationState, updateLocation } from "../src/index.js";
+
+const origin = location.origin;
+
+describe("createLocationState", () => {
+  afterEach(() => {
+    history.replaceState(null, "", origin + "/");
```
atk: Instead of adding "/" to origin after every test, you could have added it to the variable. It might only save the fraction of a millisecond, but saving energy and time never hurts.

packages/url/test/searchParams.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    expect(params).toBeInstanceOf(URLSearchParams);
+    expect(params.get("foo")).toBe("1");
+    expect(params.getAll("foo")).toEqual(["1", "2"]);
+    expect(params.toString()).toBe("foo=1&foo=2&bar=baz");
+  });
+
+  test("is granularly reactive per key", () =>
+    createRoot(dispose => {
+      const params = new ReactiveSearchParams("foo=1&bar=2");
+      let fooUpdates = 0;
```
atk: There's a pattern that I like for tests like this:

```ts
// instead of 
let fooUpdates = 0;
createEffect(() => params.get("foo"), () => { fooUpdates++; }, { defer: true })
expect(fooUpdates).toBe(1);

// I prefer using stubs
let fooUpdates = vi.fn();
createEffect(() => params.get("foo"), fooUpdates, { defer: true })
expect(fooUpdates).toHaveBeenCalledTimes(1);
// and then you can also do something like
expect(fooUpdates).toHaveBeenCalledWith("newParam");
```

packages/url/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
```
atk: Maybe update the year

atk (approved):

packages/url/src/common.ts
davedbase: When `Next` is itself a function type, `SetterValue` now only accepts the updater form — removing the ambiguity the reviewer flagged. No call sites needed changes since none currently pass a function as `Next`.

packages/url/src/location.ts
davedbase: That makes sense!

packages/url/test/searchParams.test.ts
davedbase: Got it.

PR #994, opened 2026-07-19
Fix/storage sync url

### LINKED ISSUE #703: [storage] The `messageSync` is not syncing between multiple browser tabs on different paths

Currently, the `url` of `postMessage` argument is manually set to `location.href` in the code:
https://github.com/solidjs-community/solid-primitives/blob/dceb3800fac83c9c1d99b8054a0dd44a280c6646/packages/storage/src/persisted.ts#L214-L217

This means that synchronization is only available for tabs that are on the same path. If the tabs are on different paths, synchronization will not occur. However, synchronization will resume if the tabs navigate back to the same path and a message is posted.

### Suggest A Solution

- Allow customizing the `postMessage`'s `url` argument for the `messageSync`.
- Remove this argument in the code so synchronization is available for all paths by default.

atk:
fixes #703


zobweyt:
Now you can only change the url itself, right? If I want the synchronization to work on tabs with different paths, then I can't do that? At least I don't see this case in the added tests

PR #1013, opened 2026-08-17
new: drag & drop primitive for Solid 2.0

davedbase:
A composable, tree-shakeable drag-and-drop primitives, built as two independent systems: **Pointer-events** (`makeDraggable`, `createDraggable`, `makeDroppable`, `createDroppable`, `createSortable`, `createDragContext`) for moving UI elements within the app (cards, kanban boards, sortable lists), with pluggable collision detection (`closestCenter`, `closestCorners`, `rectIntersection`, `pointerWithin`); and **Native HTML5 DnD** (`makeNativeDroppable`, `createNativeDroppable`) for OS file drops and other native drag payloads. `@solid-primitives/upload`'s `createDropzone` now composes this instead of hand-rolling native drag events itself.

Keyboard-accessible — `createDraggable` responds to `Space`/`Enter` (pick up/drop), arrow keys (nudge by `keyboardStep`, default 25px), and `Escape` (cancel), reusing the exact same collision pipeline as a pointer drag. `ref` sets `tabindex`/`role`/`aria-roledescription` automatically unless already present. Scroll-safe — the reported `transform` is corrected for page scroll mid-drag, so the dragged element doesn't visually drift from the pointer. `autoScroll` — optional window auto-scroll near viewport edges during a drag. `arrayMove` — pure reorder helper for `onDragEnd` + `createSortable`. Scales to large lists — `isDragging`/`isOver` are backed by `createProjection` rather than a per-instance `active()?.id === id` memo, so a drag start/end or hover change in a list only notifies the specific items whose state actually changed, not every item. Dev-mode warning when `createDroppable`/`createSortable` is used without a `createDragContext` ancestor. JSR-publishable (`deno.jsonc` added, verified with `deno publish --dry-run --check`).

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive drag-and-drop toolkit supporting pointer, keyboard, sortable-list, and native HTML5 file interactions.
  * Added configurable collision detection, drag contexts, acceptance filters, auto-scrolling, accessibility support, and array reordering.
  * Added documentation and interactive examples covering common drag-and-drop workflows.

* **Improvements**
  * Updated upload drop zones to use the new native drag-and-drop capabilities while preserving existing file-drop functionality.
  * Improved drag-state handling and supported drop callbacks.

* **Documentation**
  * Added package usage guidance, API references, design documentation, and licensing information.


packages/drag-drop/src/collision.ts
```diff
@@ -0,0 +1,61 @@
...
+/** Returns the droppable whose nearest corner is closest to the pointer. Uses squared distance — no sqrt needed. */
+export const closestCorners: CollisionDetector = (_draggable, droppables, pointer) => {
+  let best: string | number | null = null;
+  let bestDistSq = Infinity;
+  const px = pointer.x, py = pointer.y;
+  for (const d of droppables) {
+    const { left, right, top, bottom } = d.rect;
+    let dx: number, dy: number, distSq: number;
+    // top-left
+    dx = px - left;  dy = py - top;    distSq = dx * dx + dy * dy; if (distSq < bestDistSq) { bestDistSq = distSq; best = d.id; }
```
atk: instead of re-calculating `px - left/right`, `py - top/bottom` and their respective squares twice, better use

```ts
const dxl = px - left, dxr = px - right, dyt = py - top, dyb = py - bottom;
const dxls = dxl * dxl, dxrs = dxr * dxr, dyts = dyt * dyt, dybs = dyb * dyb;
distSq = dxl + dyt; if (distSq < bestDistSq) { bestDistSq = distSq; best = d.id; }
distSq = dxr + dyt; if (distSq < bestDistSq) { bestDistSq = distSq; best = d.id; }
distSq = dxl + dyb; if (distSq < bestDistSq) { bestDistSq = distSq; best = d.id; }
distSq = dxr + dyb; if (distSq < bestDistSq) { bestDistSq = distSq; best = d.id; }
```

to eke out a tiny bit more performance.

packages/drag-drop/src/collision.ts
```diff
@@ -0,0 +1,61 @@
+import type { CollisionDetector } from "./types.ts";
+
+/** Returns the droppable whose center is nearest to the pointer. Uses squared distance — no sqrt needed. */
+export const closestCenter: CollisionDetector = (_draggable, droppables, pointer) => {
+  let best: string | number | null = null;
+  let bestDistSq = Infinity;
+  const px = pointer.x, py = pointer.y;
+  for (const d of droppables) {
+    const dx = px - (d.rect.left + d.rect.width * 0.5);
```
atk: If you want to optimize this for a sizable performance gain, you can use `>> 1` instead of `* 0.5`. Floating point multiplications needs 48-120 ticks on modern architecture, whereas bit shift needs only 1-2 ticks.

packages/drag-drop/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Solid Primitives Working Group
```
atk: maybe adjust the year?

packages/drag-drop/README.md
```diff
@@ -0,0 +1,288 @@
...
+  <img width="100%" src="https://assets.solidjs.com/banner?type=Primitives&background=tiles&project=Drag%20%26%20Drop" alt="Solid Primitives Drag & Drop">
+</p>
+
+# @solid-primitives/drag-drop
+
+[![size](https://img.shields.io/bundlephobia/minzip/@solid-primitives/drag-drop?style=for-the-badge&label=size)](https://bundlephobia.com/package/@solid-primitives/drag-drop)
+[![version](https://img.shields.io/npm/v/@solid-primitives/drag-drop?style=for-the-badge)](https://www.npmjs.com/package/@solid-primitives/drag-drop)
+[![stage](https://img.shields.io/endpoint?style=for-the-badge&url=https%3A%2F%2Fraw.githubusercontent.com%2Fsolidjs-community%2Fsolid-primitives%2Fmain%2Fassets%2Fbadges%2Fstage-0.json)](https://github.com/solidjs-community/solid-primitives#contribution-process)
+
+Composable, tree-shakeable drag-and-drop primitives.
```
atk, line 11: While this documentation is accessible to experienced developers, amateurs have no idea how drag-and-drop works in the DOM, so maybe at least link to an explanation. I'm guilty of this style of documentation myself, so I vow to improve this.

atk (approved):
This is a really great addition to our collection. I had a few minor comments, but approve of this code regardless of changes.

packages/drag-drop/src/collision.ts
davedbase: Good call!

packages/drag-drop/src/collision.ts
davedbase: Claude says lol:

`d.rect.width` from `getBoundingClientRect()` is very commonly fractional (sub-pixel rendering, high-DPI scaling, CSS transforms, percentage layouts are all normal sources of non-integer widths). `>>` in JS isn't a generic "divide by 2" — it coerces its operand to a 32-bit signed integer first (`ToInt32`), *then* shifts. So `width >> 1` would silently truncate the fractional part of `width` before halving it, e.g. `123.9 * 0.5 = 61.95` (correct center offset) vs `123.9 >> 1 = 61` (loses ~1px). For a distance-comparison algorithm like `closestCenter`, that's exactly the kind of error that can flip which droppable is chosen when two are close.

packages/drag-drop/README.md
davedbase: Yep, I agree we want to provide the best guidance we can.

